PR #49, opened 2018-12-05
Fix integration tests.

phaazon:
The problem was in the way the Qt dispatcher was handled in
SynchronizeFromLastBlock. Nasty.


phaazon:
Still segfaulting. I tried both `lldb` and `ASan`.

  - `lldb` gives me the location where the `EXC_BAD_ACCESS` happens, but I don’t get the cause. Especially, I cannot *watch* `shared_pointer<leveldb::DB>`, so there’s no way to check why it gets affected a `nullptr` — even though I feel uncomfortable about the `weak_ptr` `unordered_map`. It’s weird.
  - `ASan` aborts on the first error (caused by a return stack scope error). Even with `ASAN_OPTIONS` set to the appropriate values, it still aborts on error.

phaazon:
Why did I closed that? 😆 Friday’s evening is very Friday!

phaazon:
Hurray!

core/test/integration/CMakeLists.txt
```diff
@@ -33,4 +33,112 @@ target_include_directories(ledger-core-integration-tests PUBLIC ../../../qt-host
 include(CopyAndInstallImportedTargets)
 copy_install_imported_targets(ledger-core-integration-tests crypto ledger-core)
 
-add_test (NAME ledger-core-integration-tests COMMAND ledger-core-integration-tests)
\ No newline at end of file
+#add_test (NAME ledger-core-integration-tests COMMAND ledger-core-integration-tests)
```
KhalilBellakrid: Is there a way of launching a batch without breaking CI (Like refacto some with a *) ?

.gitmodules
```diff
@@ -1,9 +1,11 @@
 [submodule "djinni"]
 	path = djinni
 	url = https://github.com/LedgerHQ/djinni.git
+# temporary fix for ios-nocodesign meanwhile they accept my PR
```
KhalilBellakrid: Could you please create a fork on LedgerHQ ? In case we have other changes or/and their review takes time ?

phaazon: Sure!

core/test/integration/CMakeLists.txt
phaazon: I don’t get what you’re asking for. Do you mean a way to by-pass the tests?

PR #59, opened 2018-12-19
Comment tests to allow CI execute at least compilation and unit tests

teams2ua:


phaazon:
I’m okay with this but we’ll need not to forget to re-introduce them. Maybe create a ticket in Jira for this purpose?

teams2ua:
> I’m okay with this but we’ll need not to forget to re-introduce them. Maybe create a ticket in Jira for this purpose?

Done LLC-101

PR #73, opened 2018-12-27
Add OperationService + tests

teams2ua:


core/src/math/BigInt.cpp
```diff
@@ -366,7 +366,7 @@ namespace ledger {
             }
         }
 
-        BigInt::BigInt(BigInt &&mov) {
+        BigInt::BigInt(BigInt &&mov) noexcept {
```
phaazon: _Décidemment_, we’ve been tweaking this one… :)

PR #75, opened 2019-01-02
Add persistent UTXO sources.

phaazon:
This PR adds the persistence layer implementation for UTXO sources. This is currently based on the blockchain abstraction made by @teams2ua.

We need to provide a proper design schema / RFC / document to explicitly state what’s going on with that database blockchain as I fear it would become bloated quickly — this is not trivial and we might face issue with concurrency if we’re not crystal clear about what’s happening.

Specifically:

  - The DB abstraction works on the concept of a _block_, which is NOT a blockchain’s block like from the bitcoin network. It’s a bucket that has the concept of _block height_ and is used to tag along real blocks with raw and untyped data. Another abstraction provides a typed API so that we can have typed data in those _blocks_.
  - The current implementation of persistent UTXO source also relies on the in-memory implementation to provide an efficient lookup method for the cache. We might want to have a better control on that so that we can disable the memory lookup and always go with a disk / solid state lookup (that might be needed for mobile app, for instance).
  - The tests currently are failing — segfaults — and I need to fix them so that the PR can pass completely.

### DIFF
--- a/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
+++ b/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
@@ -30,7 +30,7 @@ namespace ledger {
                             // no block in the DB so far; persist the source list
                             auto block = std::vector<uint8_t>();
                             serialization::save(sourceList, block);
-                            self->_db->AddBlock(self->_inMemorySource->currentSynchronizedHeight(), block);
+                            self->_db->AddBlock(inMemHeight, block);
                         }
 
                         return sourceList;
--- a/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
+++ b/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
@@ -18,7 +18,7 @@ namespace ledger {
                     return _db->GetLastBlock().template map<UTXOSourceList>(ctx, [=](const Option<std::pair<uint32_t, db::ReadOnlyBlockchainDB::RawBlock>>& lastBlock) {
                         if (lastBlock.hasValue()) {
                             auto dbHeight = std::get<0>(*lastBlock);
-                            auto inMemHeight = self->_inMemorySource->currentSynchronizedHeight();
+                            auto inMemHeight = sourceList.height;
 
                             if (dbHeight < inMemHeight) {
                                 // the in-memory source has advanced; we must persist the new state
--- a/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
+++ b/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
@@ -16,9 +16,10 @@ namespace ledger {
 
                 return _inMemorySource->getUTXOs(ctx).template flatMap<UTXOSourceList>(ctx, [=](const UTXOSourceList& sourceList) {
                     return _db->GetLastBlock().template map<UTXOSourceList>(ctx, [=](const Option<std::pair<uint32_t, db::ReadOnlyBlockchainDB::RawBlock>>& lastBlock) {
+                        auto inMemHeight = sourceList.height;
+
                         if (lastBlock.hasValue()) {
                             auto dbHeight = std::get<0>(*lastBlock);
-                            auto inMemHeight = sourceList.height;
 
                             if (dbHeight < inMemHeight) {
                                 // the in-memory source has advanced; we must persist the new state
--- a/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
+++ b/core/src/wallet/bitcoin/PersistentUTXOSource.cpp
@@ -1,0 +1,42 @@
+#include <bytes/serialization.hpp>
+#include <wallet/bitcoin/PersistentUTXOSource.hpp>
+#include <database/BlockchainDB.hpp>
+
+namespace ledger {
+    namespace core {
+        namespace bitcoin {
+            PersistentUTXOSource::PersistentUTXOSource(
+                const std::shared_ptr<db::BlockchainDB>& blockchainDB,
+                const std::shared_ptr<UTXOSourceInMemory>& inMemorySource
+            ) : _db(blockchainDB), _inMemorySource(inMemorySource) {
+            }
+
+            Future<UTXOSourceList> PersistentUTXOSource::getUTXOs(std::shared_ptr<api::ExecutionContext> ctx) {
+                auto self = shared_from_this();
+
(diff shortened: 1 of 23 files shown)


core/src/wallet/bitcoin/PersistentUTXOSource.cpp
```diff
@@ -0,0 +1,42 @@
...
+                const std::shared_ptr<db::BlockchainDB>& blockchainDB,
+                const std::shared_ptr<UTXOSourceInMemory>& inMemorySource
+            ) : _db(blockchainDB), _inMemorySource(inMemorySource) {
+            }
+
+            Future<UTXOSourceList> PersistentUTXOSource::getUTXOs(std::shared_ptr<api::ExecutionContext> ctx) {
+                auto self = shared_from_this();
+
+                return _inMemorySource->getUTXOs(ctx).template flatMap<UTXOSourceList>(ctx, [=](const UTXOSourceList& sourceList) {
+                    return _db->GetLastBlock().template map<UTXOSourceList>(ctx, [=](const Option<std::pair<uint32_t, db::ReadOnlyBlockchainDB::RawBlock>>& lastBlock) {
```
teams2ua: Not your fault. Just thoughts - GetLastBlock read entire block, that can be big, actually it needs only height, maybe we need to remember last returned value, or add new method to DB (or have another db for this purposes, as you write in PR comment)

core/src/wallet/bitcoin/PersistentUTXOSource.cpp
```diff
@@ -0,0 +1,42 @@
...
+                        if (lastBlock.hasValue()) {
+                            auto dbHeight = std::get<0>(*lastBlock);
+                            auto inMemHeight = self->_inMemorySource->currentSynchronizedHeight();
```
teams2ua: currentSynchronizedHeight should be returned in the getUTXOs, somebody can call getUTXOs in other thread just after our getUTXOs call but before currentSynchronizedHeight(), that may change the value of currentSynchronizedHeight .
In general I think we will have the height returned with actual result in a lot of functions, I don't see other thread-safe mechanism.

core/src/wallet/bitcoin/PersistentUTXOSource.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <database/BlockchainDB.hpp>
+#include <wallet/bitcoin/UTXOSource.hpp>
+#include <wallet/bitcoin/UTXOSourceInMemory.hpp>
+
+namespace ledger {
+    namespace core {
+        namespace bitcoin {
+            class PersistentUTXOSource: public UTXOSource, public std::enable_shared_from_this<PersistentUTXOSource> {
+                std::shared_ptr<db::BlockchainDB> _db; ///< Database used for UTXOs persistence.
+                std::shared_ptr<UTXOSourceInMemory> _inMemorySource; ///< Memory cache.
```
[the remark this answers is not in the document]
teams2ua: I think it will be easier to use this way and harder to write some random data to DB, for now somebody can pass random UTXOSourceInMemory and it will be written to DB.
But this is up to you.

core/src/wallet/bitcoin/PersistentUTXOSource.cpp
teams2ua: Right, it is only height.
Go ahead if you fill like you are on it.

core/src/wallet/common/OperationServiceOnDB.hpp
```diff
@@ -39,9 +39,9 @@ namespace ledger {
                         _unstableBlocksDB->getBlocks(fromBlock, toBlock),
                         _stableBlocksDB->getBlocks(fromBlock, toBlock)
                     };
-                    auto self = shared_from_this();
+                    auto self = this->shared_from_this();
                     return executeAll(_executionContext, futures)
-                        .map<std::vector<Operation>>(_executionContext, [self](const std::vector<std::vector<FilledBlock>>& vectorBlocks) {
+                        .template map<std::vector<Operation>>(_executionContext, [self](const std::vector<std::vector<FilledBlock>>& vectorBlocks) {
```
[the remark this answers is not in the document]
phaazon: I guess appveyor uses the same compiler as yours?

core/src/wallet/bitcoin/PersistentUTXOSource.hpp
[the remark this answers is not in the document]
phaazon: Yep, I thought about that and if we do that, it’ll put more pressure on this class (since it will forward all the arguments to create the `UTXOSourceInMemory` as dependencies). I think I’ll leave it this way.

core/src/wallet/bitcoin/PersistentUTXOSource.cpp
phaazon: Done in the current last commit.

core/src/wallet/bitcoin/PersistentUTXOSource.cpp
phaazon: Done.

core/src/wallet/bitcoin/PersistentUTXOSource.cpp
```diff
@@ -0,0 +1,42 @@
...
+                                // the in-memory source has advanced; we must persist the new state
+                                auto block = std::vector<uint8_t>();
+                                serialization::save(sourceList, block);
+                                self->_db->AddBlock(inMemHeight, block);
+                            }
+                        } else {
+                            // no block in the DB so far; persist the source list
+                            auto block = std::vector<uint8_t>();
+                            serialization::save(sourceList, block);
+                            self->_db->AddBlock(self->_inMemorySource->currentSynchronizedHeight(), block);
```
teams2ua: self->_inMemorySource->currentSynchronizedHeight() should be changed to inMemHeight, right?

core/src/wallet/common/PersistentBlockchainDatabase.hpp
```diff
@@ -80,6 +80,15 @@ namespace ledger {
                     });
                 }
 
+                Future<Option<uint32_t>> getLastBlockHeight() override {
+                    return _persistentDB->GetLastBlock()
```
teams2ua: Now we don't parse RawBlock, that is good, but we still read it.
Let keep this in mind for future.

core/src/wallet/bitcoin/PersistentUTXOSource.cpp
phaazon: Right. :)

phaazon: Done.

PR #76, opened 2019-01-02
Add the possibility to switch off spdlog.

phaazon:
Relates to LLC-102.

### DIFF
--- a/core/src/debug/logger.cpp
+++ b/core/src/debug/logger.cpp
@@ -31,26 +31,42 @@
 
 #include "logger.hpp"
 #include <spdlog/spdlog.h>
+#include <spdlog/sinks/null_sink.h>
 #include "LogPrinterSink.hpp"
 #include "RotatingEncryptableSink.hpp"
 
 namespace ledger {
     namespace core {
 
-        std::shared_ptr<spdlog::logger> logger::create(const std::string &name,
-                                 std::experimental::optional<std::string> password,
-                                 const std::shared_ptr<api::ExecutionContext> &context,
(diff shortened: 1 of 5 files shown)


core/src/debug/logger.cpp
```diff
@@ -31,26 +31,42 @@
...
-            auto logger = spdlog::create(name, {logPrinterSink, rotatingSink});
-            spdlog::drop(name);
-            logger->set_level(spdlog::level::trace);
-            logger->flush_on(spdlog::level::trace);
-            logger->set_pattern("%Y-%m-%dT%XZ%z %L: %v");
-            return logger;
+
+            if (disabled) {
+                auto logger = spdlog::create<spdlog::sinks::null_sink_st>(name);
+                spdlog::drop(name);
```
teams2ua: As I understand from the doc https://github.com/gabime/spdlog/wiki/5.-Logger-registry 

> "The "drop()" function can be used to remove a logger from the registry.
> If no other shared_ptr pointing to the logger exists, the logger will be closed and all its resources will be freed."

Will be the `logger` still alive in the next line?

phaazon: I don’t really know why we do that since that line was already there prior to my PR. :confused:

teams2ua: Ah I see, somebody else should call `spdlog::register_logger(some_logger);` latter, strange we have this line here. OK

phaazon:
I’m still leaving it tagged until I have added some kind of a test to ensure no file gets generated.

phaazon:
I shall merge this as I tested it correctly. I’ll make another PR with the IDL commits.

PR #83, opened 2019-01-09
Fix the logger that truncated the current log.

phaazon:
We were explicitly asking to truncate files on rotations. I’m asking @pollastri-pierre reviewal because it was his code and maybe he had a good reason to do that.


phaazon:
Yeah actually I’m putting this one on **HODL** also because the current implementation should work. It’s weird that we lose log files.

phaazon:
Seen with Live guys: the problem is linked to a workaround they implement for privacy purposes. I also audited the code further and the `_rotate` function is genuine. I’m closing and not merging.

PR #87, opened 2019-01-14
Run the 'generate_doc' job only when a git tag exists (CircleCI).

phaazon:


.circleci/config.yml
```diff
@@ -252,8 +252,9 @@ workflows:
...
       - build_macos_debug
+  documentation:
+    jobs:
       - generate_doc:
           filters:
-            branches:
-              only:
-                - develop
+            tags:
+              only: /^v.*/
```
Vedrillan: You have to add this if you want to run on tags only
```
branches:
    ignore: /.*/
```

phaazon: Thanks! Resolved!

PR #91, opened 2019-01-15
Do not build cmake each time on CI

teams2ua:
cmake_install_old


phaazon:
Why closing! :scream:

PR #96, opened 2019-01-16
Add freshResetAll to reset libcore to a “fresh install”-like state

phaazon:

### DIFF
--- a/core/src/database/DatabaseSessionPool.cpp
+++ b/core/src/database/DatabaseSessionPool.cpp
@@ -59,9 +59,7 @@ namespace ledger {
         }
 
         DatabaseSessionPool::~DatabaseSessionPool() {
-            if (_logger != nullptr) {
-                delete _logger;
-            }
+            delete _logger;
         }
 
         FuturePtr<DatabaseSessionPool>
--- a/core/src/database/DatabaseSessionPool.cpp
+++ b/core/src/database/DatabaseSessionPool.cpp
@@ -28,17 +28,20 @@
  * SOFTWARE.
  *
  */
+
 #include "DatabaseSessionPool.hpp"
 #include "migrations.hpp"
 
 namespace ledger {
     namespace core {
         FuturePtr<DatabaseSessionPool>
-        DatabaseSessionPool::getSessionPool(const std::shared_ptr<api::ExecutionContext> &context,
-                                            const std::shared_ptr<DatabaseBackend>& backend,
-                                            const std::shared_ptr<api::PathResolver>& resolver,
-                                            const std::shared_ptr<spdlog::logger>& logger,
-                                            const std::string& dbName) {
+        DatabaseSessionPool::getSessionPool(
+            const std::shared_ptr<api::ExecutionContext> &context,
+            const std::shared_ptr<DatabaseBackend>& backend,
+            const std::shared_ptr<api::PathResolver>& resolver,
+            const std::shared_ptr<spdlog::logger>& logger,
+            const std::string& dbName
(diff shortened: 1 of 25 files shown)


phaazon:
**HODL**, I missed to expose the `clear` function.

phaazon:
Done, this is now reviewable.

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -28,17 +28,50 @@
...
+                if (_logger != nullptr)
+                    session.set_log_stream(_logger);
+            }
+
+            // Migrate database
+            performDatabaseMigration();
+        }
+
+        DatabaseSessionPool::~DatabaseSessionPool() {
+            if (_logger != nullptr) {
```
teams2ua: This is redundant, it is safe to `delete nullptr;`

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -28,17 +28,50 @@
...
+                delete _logger;
```
teams2ua: Don't you want to assign _logger=nullptr; after delete?

core/src/wallet/pool/WalletPool.cpp
```diff
@@ -106,6 +105,9 @@ namespace ledger {
             _publisher = std::make_shared<EventPublisher>(getContext());
         }
 
+        WalletPool::~WalletPool() {
```
teams2ua: Isn't this the same behavior as default constructor?

core/src/database/DatabaseSessionPool.cpp
teams2ua: Ah this is not a new code... ok then

phaazon: Yeah I also felt it could be removed. I will remove it. Cleaning codebase is important.

core/src/wallet/pool/WalletPool.cpp
phaazon: It is, but… that’s the previous code (I just moved it around). I code remove that and put a `= default` if you prefer (I do prefer, too).

core/src/database/migrations.cpp
```diff
@@ -32,21 +32,41 @@
 
 namespace ledger {
     namespace core {
+        int getDatabaseMigrationVersion(soci::session& sql) {
+            int version = -1;
+
+            try {
+                soci::statement st = (sql.prepare << "SELECT version FROM __database_meta__ WHERE id == 0", soci::into(version));
```
KhalilBellakrid: Could you remove the "==" please ? :trollface:

core/src/database/migrations.cpp
```diff
@@ -280,5 +358,27 @@ namespace ledger {
...
+            sql << "DROP TABLE erc20_operations";
+
+            // ERC20 accounts
+            sql << "DROP TABLE erc20_accounts";
+
+            // ETH operations
+            sql << "DROP TABLE ethereum_operations";
+
+            // ETH transactions
+            sql << "DROP TABLE ethereum_operations";
```
KhalilBellakrid: Probably you wanted to `DROP TABLE ethereum_transactions` here ?

core/src/database/migrations.cpp
phaazon: Hehe, sure. Done.

core/src/database/migrations.cpp
phaazon: Argh, nice catch.

PR #97, opened 2019-01-16
Retrieve block height of currency from DB in transaction construction

KhalilBellakrid:
This causes issues when using current block height stored in memory ... so parsing/serialization of transaction was using the wrong BIP/fork of a currency leading in some issues from Live side

### DIFF
--- a/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
+++ b/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
@@ -531,6 +531,21 @@ namespace ledger {
             });
         }
 
+        // This method is used to get a block height only used to know which
+        // update/fork (activated at certain block height) is used.
+        // This explains why we take LLONG_MAX is we have no block in DB for
+        // certain currency
+        // WARNING: please don't use this method if you need an accurate value
+        // of last block
+        static uint64_t getLastBlockFromDB(soci::session &sql, const std::string &currencyName) {
+            //Get last block from DB
+            auto lastBlock = BlockDatabaseHelper::getLastBlock(sql, currencyName);
+            // If we can not retrieve a last block for currency,
+            // we set the returned block height to LLONG_MAX in order to
+            // activate/apply last BIP/update/fork for this currency.
+            return lastBlock.hasValue() ? static_cast<uint64_t >(lastBlock->height) : LLONG_MAX;
+        }
+
         void BitcoinLikeAccount::broadcastRawTransaction(const std::vector<uint8_t> &transaction,
                                                          const std::shared_ptr<api::StringCallback> &callback) {
             auto self = getSelf();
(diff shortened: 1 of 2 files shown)


core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -531,14 +531,32 @@ namespace ledger {
...
 
+        // This method is used to get a block height only used to know which
+        // update/fork (activated at certain block height) is used.
+        // This explains why we take LLONG_MAX is we have no block in DB for
+        // certain currency
+        // WARNING: please don't use this method if you need an accurate value
+        // of last block
+        static uint64_t getLastBlockFromDB(soci::session &sql, const std::string &currencyName) {
+            //Get last block from DB
+            auto lastBlock= BlockDatabaseHelper::getLastBlock(sql, currencyName);
```
phaazon: Whitespace between `lastBlock` and `=` :)

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -531,14 +531,32 @@ namespace ledger {
...
+            //If we can not retrieve a last block for currency, we apply last BIP/update/fork
+            return lastBlock.hasValue() ? static_cast<uint64_t >(lastBlock.getValue().height) : LLONG_MAX;
```
phaazon: Does it mean we retreive a whole block just for discarding it and use only the height? I’m fine with this as first implementation but I guess it could be more efficient to have a way to get only the height (without getting payload) from the `BlockDatabaseHelper`. That means adding `BlockDatabaseHelper::getLastBlockHeight` or something like that.

phaazon: Also, I _think_ you can change this:

```
lastBlock.getValue().height
```

with this:

```
lastBlock->heigth
```

(there’s a prefix `operator*` for `Optional<_>` but I’m not sure there is for `operator->`)

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
KhalilBellakrid: My bad !

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
KhalilBellakrid: For the payload issue, I agree on the concept, here the idea was just reusing an existing method, more generally we should have a more generic method to interact/retrieve data with/from DB but it should be easier with the new DB abstraction

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -531,14 +531,32 @@ namespace ledger {
...
+        // update/fork (activated at certain block height) is used.
+        // This explains why we take LLONG_MAX is we have no block in DB for
+        // certain currency
+        // WARNING: please don't use this method if you need an accurate value
+        // of last block
+        static uint64_t getLastBlockFromDB(soci::session &sql, const std::string &currencyName) {
+            //Get last block from DB
+            auto lastBlock = BlockDatabaseHelper::getLastBlock(sql, currencyName);
+            //If we can not retrieve a last block for currency, we apply last BIP/update/fork
+            return lastBlock.hasValue() ? static_cast<uint64_t >(lastBlock->height) : LLONG_MAX;
```
teams2ua: [OPT] Define a constant for this, something like HIGHEST_POSSIBLE_BLOCK

KhalilBellakrid: As discussed, I have just updated comments, introducing a new constant will just bring more confusing (specially with a name like HIGHEST_POSSIBLE_BLOCK)

PR #99, opened 2019-01-16
Static lib

teams2ua:
To overcome Windows export symbols limitation on DLL, we will explicitly mark API classes, but in tests we may want any code from lib-core.


core/src/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.12)
```
phaazon: Nice addition.

core/src/CMakeLists.txt
```diff
@@ -46,20 +46,23 @@ endif ()
...
-            ${ledger-core-sources}
-            ${SRC_FILES}
-            ${HEADERS_FILES})
+add_library(ledger-core-interface INTERFACE)
+
+target_compile_definitions(ledger-core-interface INTERFACE ledger_core_EXPORTS)
 
 string(FIND "${CMAKE_OSX_SYSROOT}" "iphone" IS_IOS)
 if(IS_IOS GREATER_EQUAL 0)
+	add_library(ledger-core SHARED
```
phaazon: So for iOS, `ledger-core` is a shared library but for the rest will be object libraries. Why?

core/src/libcore_export.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifndef LEDGER_CORE_DEPRECATED_EXPORT
+#  define LEDGER_CORE_DEPRECATED_EXPORT LIBCORE_EXPORT LEDGER_CORE_DEPRECATED
+#endif
+
+#ifndef LEDGER_CORE_DEPRECATED_NO_EXPORT
+#  define LEDGER_CORE_DEPRECATED_NO_EXPORT LEDGER_CORE_NO_EXPORT LEDGER_CORE_DEPRECATED
+#endif
+
+#if 0 /* DEFINE_NO_DEPRECATED */
```
phaazon: This will never be preprocessed, so I think you can remove it?

teams2ua: It was generated automatically by CMAKE command (that I delete). So I don't see a point to modify it.

core/src/CMakeLists.txt
teams2ua: No, ledger-core will always be a shared library. For other platforms I am creating temporary "object library" ledger-core-obj from which ledger-core and ledger-core-static are build.

core/src/CMakeLists.txt
teams2ua: Need this for proper object library support (include object library to another target with target_link_library)

PR #106, opened 2019-01-18
Adapt (again) derivation address for ETH

KhalilBellakrid:
- ETH address derivation is now like Bitcoin (with respect to BIP32 and BIP44),
- We support also configurable new derivation schemes (as shown in tests), this will allow Live to support legacy derivation paths and custom (weird) ones,
- When asked about new account's creation info, client will receive a structure`{index, owners, derivations}` e.g. if we ask for first ETH account of a wallet with derivation scheme `44'/<coin_type>'/<account>'/<node>/<address>` we will receive: 
`{0, { main }, {" 44'/60'/0' "}}`
- Tests that were added were checked against returned values by a NanoS

### DIFF
--- a/core/src/ethereum/EthereumLikeExtendedPublicKey.cpp
+++ b/core/src/ethereum/EthereumLikeExtendedPublicKey.cpp
@@ -37,6 +37,7 @@
 #include <bytes/BytesReader.h>
 #include <bytes/BytesWriter.h>
 
+#include <utils/hex.h>
 #include <utils/Exception.hpp>
 
 #include <crypto/Keccak.h>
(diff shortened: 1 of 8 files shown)


core/src/ethereum/EthereumLikeExtendedPublicKey.cpp
```diff
@@ -54,25 +55,43 @@ namespace ledger {
             _currency(params), _key(key), _path(path)
         {}
 
-        std::shared_ptr<api::EthereumLikeAddress> EthereumLikeExtendedPublicKey::derive(const std::string & path) {
+        static inline DeterministicPublicKey _derive(int index, const std::vector<uint32_t>& childNums, const DeterministicPublicKey& key) {
+            if (index >= childNums.size()) {
```
teams2ua: [OPT] this function can be not recursive.

PR #110, opened 2019-01-21
Modernize the implementation of DynamicValue to consume (way) less memory.

phaazon:
Also, as a nice side-effect, I made the code in a way so that it’s very easy to switch it to C++17 when we switch.


core/src/collections/DynamicValue.hpp
```diff
@@ -28,47 +28,130 @@
...
+            DynamicValue& operator=(double rhs);
+
+            /// Change the value to a DynamicArray by copying.
+            DynamicValue& operator=(const std::shared_ptr<DynamicArray>& rhs);
 
+            /// Change the value to a DynamicObject by copying.
+            DynamicValue& operator=(const std::shared_ptr<DynamicObject>& rhs);
+
+            /// Try to get the value as if it were a std::string.
+            optional<std::string&> asStr();
```
teams2ua:
```
template<typename T>
optional<T> as() {
   if (data.type() == typeid(T)) return boos::get<T>(data);
   return optional<T>();
}
```

core/src/collections/DynamicValue.hpp
```diff
@@ -28,47 +28,130 @@
...
+            /// Try to get the value as if it were a std::string.
+            optional<const std::string&> asStr() const;
+
+            /// Try to get the value as if it were a std::vector<uint8_t>.
+            optional<std::vector<uint8_t>&> asData();
+            /// Try to get the value as if it were a std::vector<uint8_t>.
+            optional<const std::vector<uint8_t>&> asData() const;
+
+            /// Try to get the value as if it were a bool.
+            optional<bool&> asBool();
```
teams2ua: Why do we need to have a reference here? Why not optional<bool>?

core/src/collections/DynamicValue.hpp
```diff
@@ -28,47 +28,130 @@
...
+            optional<const std::shared_ptr<DynamicArray>&> asArray() const;
+
+            /// Try to get the value as if it were a DynamicObject.
+            optional<std::shared_ptr<DynamicObject>&> asObject();
+            /// Try to get the value as if it were a DynamicObject.
+            optional<const std::shared_ptr<DynamicObject>&> asObject() const;
+
+        private:
+            // Tag representing the current in-use variant
+            api::DynamicType type;
```
teams2ua: Can remove this as well.
define values inside variant in the same order as in DynamicType, for UNDEFINED create a dummy struct Undefined{}; and then use boost::variant::which with conversion to enum.

core/src/collections/DynamicValue.hpp
teams2ua: Look at https://www.boost.org/doc/libs/1_65_0/doc/html/boost_typeindex.html

core/src/collections/DynamicValue.hpp
teams2ua: Semantic of this operation is not clear for me

phaazon: The former mutates `x`, the second one will create an error (because you will get `NONE`). If you want to change the type, you have to use `operator=`. And accessing the value of of an `optional<_>` should be done, in C++, by testing whether the value is there prior. So in your second example, you will get `hasValue() == false`.

teams2ua: I found this dangerous.
```
auto x = DynamicValue(false);
auto y = x.asBool();
return y;
....
y.value = true; // this change x, that may not exist anymore
```

phaazon: This is typical C++ when someone tries to escape a reference from a local scope. It has nothing to do with the `optional` :) And I hope `msvc` / `clang` / `g++` will be able to see such an escaped reference!

core/src/collections/DynamicValue.hpp
phaazon: I know that and I know RTTI is big and heavy and I wish we could get rid of it. However, there is a way to go with it without RTTI and still removing that `api::DynamicType` stuff directly — using the type index in the variant, which is a bit silly since you might have issues if you had types in the middle of the variant.

core/src/collections/DynamicValue.hpp
teams2ua: What is the benefit to have & inside optional? We can mutate DynamicValue directly if we want, & is not free, it still consume the same memory as pointer (so for bool it is even worse).
Taking this in mind and the danger nature of having optional<T&> I would go for more safe way - optional<T>. 
Looks like compilers don't want to help us here - https://ideone.com/4Pt2YH - runtime error.

phaazon: `optional<T&>` gives you “maybe” a reference that you can use to manipulate in a type-safe way the content of the `DynamicValue`. If it doesn’t contain a `T`, you don’t get a reference. Of course we’re talking about C++, so it’s easy to shoot oneself in the feet if you try to use the reference after having changed the `DynamicValue` or if you use it after free — Rust has escape analysis at least.

If you want to “keep a mutable reference” of a dynamic value and change its content without having to continuously check the type at runtime, this method gives you such semantics. It provides a better type-safe layer than accessing it via `int32` or that kind of things — look at the diff code, I removed a lot of dynamic type checking thanks to this function. However, I agree that it doesn’t solve the problem of escape analysis (we have the same problem with `std::vector<_>::data()`, for instance: you can get a pointer to something that has been freed, regrown or even moved to another memory place.

What you seem to be worried about here is the reference, not really the contract of those methods. I get your point but if we forbid people to do this, that means you cannot mutate the content of a `DynamicValue` without copying. If you think we don’t care, then I can remove the reference version, if you prefer, and return by copy.

teams2ua:
Nice!
Why do we have exactly the same serialization for PortableBinaryArchive and BinaryArchive?
should this be one template function?

phaazon:
> Nice!
> Why do we have exactly the same serialization for PortableBinaryArchive and BinaryArchive?
> should this be one template function?

I don’t really know, this is legacy code. I guess we could use templates for those.

PR #119, opened 2019-01-25
Allow several build system on same source dir

teams2ua:
Remove generated CMake files from source directory
To allow building with several build system without cleaning.


teams2ua:
> If now we build secp256k1 in same build dir as ledger-core project we can get rid of this : https://github.com/LedgerHQ/lib-ledger-core/blob/develop/.circleci/build_lib.sh#L109

Removed this, as far as I understand the generated source is not dependent on the platform, so this should be safe.

PR #120, opened 2019-01-25
Some cleaning in generated code+new VS enable

teams2ua:
It doesn't matter what is the version of VS studio
as this is not some limitation of older version that we want
to overcome, but limitation of loader


phaazon:
Good to go to me.

PR #121, opened 2019-01-25
First pass of fixing and documenting the IDL files.

phaazon:
I might open several PRs for that as it has no impact but documentation.

### DIFF
--- a/core/idl/database/database.djinni
+++ b/core/idl/database/database.djinni
@@ -241,6 +241,6 @@ DatabaseBackend = interface +c {
     # @return DatabaseBackend object
     static getSqlite3Backend(): DatabaseBackend;
 
-    # Create an instance of PostgreSQL database.
+    # Create a database backend instance from the given DatabaseEngine implementation.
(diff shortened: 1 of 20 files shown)


core/idl/database/database.djinni
```diff
@@ -240,10 +237,10 @@ DatabaseBackend = interface +c {
     # @return trye if query logging is enabled, false otherwise.
     isLoggingEnabled(): bool;
 
-    # Create an instance of SQLite3 database
+    # Create an instance of SQLite3 database.
     # @return DatabaseBackend object
     static getSqlite3Backend(): DatabaseBackend;
 
-    # Create an instance of PostgreSQL database
+    # Create an instance of PostgreSQL database.
```
pollastri-pierre: Ouch "Copier/Gaufrer" (my bad). It should be something like create a database backend instance from the given DatabaseEngine implementation.

phaazon: I’ll fix it! Thanks for the review 🙇

PR #122, opened 2019-01-28
Provide re-encryption algorithm for preferences (LevelDB)

phaazon:


core/src/preferences/PreferencesBackend.cpp
```diff
@@ -196,8 +223,96 @@ namespace ledger {
...
+            // brand new cipher; create it here
+            auto newSalt = createNewSalt(rng);
+            auto newCipher = Option<AESCipher>(AESCipher(rng, newPassword, newSalt, PBKDF2_ITERS));
+
+            // we will also need a brand new leveldb to implement a somewhat atomic database swap;
+            // the idea is to read from the old leveldb and write to this new one; when we’re done,
+            // we just “swap” the database and remove the old one; this swap is needed so that we
+            // don’t corrupt the old database if we crash or get interrupted while in the middle of
+            // the process of resetting
+            auto tempDBName = _dbName + "__temp_copy__";
```
teams2ua: I think we can do this without additional DB, just push all changes to one batch https://github.com/google/leveldb/blob/master/doc/index.md#atomic-updates , both delete old values and put new then make one Write(batch) that will do the things, this way we push the burden to atomically swap to levelDB.

phaazon: Nice idea. I’ll change it to reflect your idea. It’ll be way more lightweight.

phaazon: Done. That was nice to rewrite the code with this feature. 😄

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -196,8 +211,63 @@ namespace ledger {
             _cipher = Option<AESCipher>::NONE;
         }
 
+        void PreferencesBackend::resetEncryption(
+            const std::shared_ptr<api::RandomNumberGenerator>& rng,
+            const std::string& newPassword,
```
KhalilBellakrid: I thought you changed the order, no ? or just fixed your tests ? 
Because it feel more natural to start with old password then new password ! IMHO

core/src/preferences/PreferencesBackend.hpp
```diff
@@ -95,6 +95,14 @@ namespace ledger {
             /// Data already present in the preferences are not affected.
             void unsetEncryption();
 
+            // Reset the encryption with a new password by first decrypting on the
+            // fly with the old password.
+            void resetEncryption(
```
KhalilBellakrid: Same here !

core/src/preferences/PreferencesBackend.cpp
phaazon: I just fixed the tests, yep. I’ll switch the order!

core/src/preferences/PreferencesBackend.hpp
phaazon: ACK

PR #125, opened 2019-01-29
Fix a bug in Live with the AESCipher (that doesn’t like empty passwor…

phaazon:
…ds).

This fixes the first part of the problem. I’ll open a Jira ticket to fix
AES/PBKDF2 when the key / passphrase is empty.

Also, note: we might want to change the interface of the WalletPool
creation. This would be a breaking change, though. I’ll also open a Jira
ticket for that.

### DIFF
--- a/core/src/wallet/pool/WalletPool.cpp
+++ b/core/src/wallet/pool/WalletPool.cpp
@@ -53,7 +53,16 @@ namespace ledger {
         ): DedicatedContext(dispatcher->getSerialExecutionContext(fmt::format("pool_queue_{}", name))) {
             // General
             _poolName = name;
-            _password = password;
+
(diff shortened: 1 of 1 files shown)


phaazon:
I created some Jiras issues related to this very issue:

- [LLC-154](https://ledgerhq.atlassian.net/browse/LLC-154), about a weird use of the password in `RotatingLogSink` that I found suspicious.
- [LLC-155](https://ledgerhq.atlassian.net/browse/LLC-155), adding the `AESCipher` bug into our backlog.
- [LLC-156](https://ledgerhq.atlassian.net/browse/LLC-156), about reviewing the encoding.

phaazon:
Do you think we also need to backport this fix to all versions until the appearance of the cipher? I’ll make a PR to Live to fix their code for all other versions but I wonder whether we need to backport.

PR #138, opened 2019-02-05
Tweak database engine implementation

pollastri-pierre:
- Small fixes on SQL requests
- Doesn't require for a cursor to know the full size anymore (ability to do some lazy loading)
- Add tests for dynamic columns management


core/src/api/DatabaseResultSet.hpp
```diff
@@ -34,28 +34,22 @@ class DatabaseResultSet {
...
 
     /**
-     * Returns true if the result set has at least one remaining row to get.
-     * @return true if the result set has at least one remaining row to get, false otherwise.
-     */
-    virtual bool hasNext() = 0;
-
-    /**
-     * Move the result set to the next available result. This method may fail if there is now further row to fetch.
+     * Internally move the result set to the next available row. This method may fail if there is now further row to fetch.
```
phaazon: s/now/no

pollastri-pierre: Indeed 👍

core/test/database/MemoryDatabaseProxy.cpp
```diff
@@ -248,6 +243,7 @@ class Results : public api::DatabaseResultSet, public std::enable_shared_from_th
 
     std::list<std::shared_ptr<ResultRow>> _rows;
     std::list<std::shared_ptr<ResultRow>>::iterator _it;
+    int _currentIndex;
```
teams2ua: Don't get what is this member for? It is only used in available()?

core/src/database/proxy_backend/session.cpp
```diff
@@ -63,8 +66,7 @@ proxy_statement_backend *proxy_session_backend::make_statement_backend() {
 }
 
 proxy_rowid_backend *proxy_session_backend::make_rowid_backend() {
-    // TODO: Create ROWID interface
-    return nullptr;
+    throw make_exception(api::ErrorCode::DATABASE_EXCEPTION , "ROWID is not supported by database backend");
```
phaazon: Why not sticking to the `nullptr`?

core/src/database/proxy_backend/soci-proxy.h
```diff
@@ -59,6 +59,7 @@
 #   define SOCI_OVERRIDE override
 #endif
 
+#define SOCI_PROXY_DEBUG
```
phaazon: I’m not sure whether it’s on purpose, but this will make the block below always gets substituted to the same `SP_PRINT` definition.

pollastri-pierre: Oh god yes, I remove that

core/src/database/proxy_backend/session.cpp
pollastri-pierre: Because I prefer that the day we use a rowid by mistake we have an explicit exception and not trying to find the reason of the failure. The fact that rowid is unsupported is documented in the database md file

PR #140, opened 2019-02-06
Fix the generate_interfaces / generateBindingsRN scripts. [skip ci]

phaazon:
Remove already existing files to prevent bad things from happening.

### DIFF
--- a/core/src/api/Account.hpp
+++ b/core/src/api/Account.hpp
@@ -9,7 +9,7 @@
 #include <memory>
 #include <string>
 #ifndef LIBCORE_EXPORT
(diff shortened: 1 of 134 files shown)


core/src/api/Account.hpp
```diff
@@ -9,7 +9,7 @@
 #include <memory>
 #include <string>
 #ifndef LIBCORE_EXPORT
-    #if defined(_MSC_VER)
+    #if defined(_MSC_VER) && _MSC_VER <= 1900
```
KhalilBellakrid: Please update your djinni before regenerating the bindings, it seems that you have an old version !

PR #144, opened 2019-02-08
Add a test for empty key / empty passwords for AESCipher.

phaazon:


core/test/crypto/encryption_test.cpp
```diff
@@ -154,4 +154,38 @@ TEST(Encryption, EncryptDecryptWithCipherHugeText) {
...
-}
\ No newline at end of file
+}
+
+TEST(Encryption, EncryptDecryptWithCipherEmptiness) {
+    auto rng = std::make_shared<OpenSSLRandomNumberGenerator>();
+
+    std::vector<uint8_t> vec(BIG_TEXT.begin(), BIG_TEXT.end());
+    BytesReader input(vec);
+    BytesWriter encrypted;
```
teams2ua: encrtypted can be created inside lambda, so you will not need clear() method on BytesWriter

phaazon: While this is true, it will also make the code perform more memory allocations.

phaazon: I’ll remove `clear` because it’s just tests and no one seems interested in that function.

PR #146, opened 2019-02-12
SQL Cipher fix

KhalilBellakrid:
- Define OPENSSL_LIB as crypto TARGET_FILE_DIR (instead of TARGET_FILE)
- @phaazon this is probably the reason building libcore with SQLCipher was failing ( Pierre had same issue and I fixed by that), I think reason why it was not failing on my side and CI side is because probably it was linking against system's OpenSSL which happen to have right version with all symbols defined !


phaazon:
That should fix the issue as I saw some weird stuff about `FindCryptoPP.cmake` in my logs lately.

PR #147, opened 2019-02-13
Prepare for 2.7.0: XRP integration

KhalilBellakrid:
XRP integration:

- XRP Keychian, address derivation and account creation,
- XRP API explorer and parser,
- XRP Node explorer and parser,
- Accounts synchronization,
- XRP transaction serialization and parsing,
- Tests of above features.


phaazon:
3.0.0? Do we really break anything? :/ If so, we really need to work on a non-breaking scheme for coins!

KhalilBellakrid:
@phaazon there are no breaking changes but it seemed legit to bump major vesion since we add a new coin family to libcore, but indeed we can stick to SemVer and bump only minor one ! I will change that 👍

CHANGELOG.md
```diff
@@ -17,6 +17,13 @@
 - Changed deployement process, now RCs are deployed with format `x.y.z-rc-{COMMIT_HASH}`, this will allow us
   not to always bump the versions when merging PRs and also keeping the binaries deployed to our bucket.
 
+- XRP integration:
+	- XRP Keychian, address derivation and account creation,
```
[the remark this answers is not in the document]
KhalilBellakrid: Nope, I'm already on 4 👍

core/idl/ripple/addresses.djinni
```diff
@@ -0,0 +1,33 @@
+@import "../math/big_int.djinni"
+
+RippleLikeNetworkParameters = record {
+	Identifier: string;
+	MessagePrefix: string;
+	XPUBVersion: binary;
+	AdditionalRIPs: list<string>;
```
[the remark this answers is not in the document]
KhalilBellakrid: Exactly ! Thought about it too when writing it haha 😆

core/src/api/Currency.hpp
```diff
@@ -39,21 +40,25 @@ struct Currency final {
     std::experimental::optional<BitcoinLikeNetworkParameters> bitcoinLikeNetworkParameters;
     /** Optional EthereumLikeNetworkParameters, for more details refer to EthereumLikeNetworkParameters doc */
     std::experimental::optional<EthereumLikeNetworkParameters> ethereumLikeNetworkParameters;
+    /**Optional EthereumLikeNetworkParameters, for more details refer to EthereumLikeNetworkParameters doc */
+    std::experimental::optional<RippleLikeNetworkParameters> rippleLikeNetworkParameters;
```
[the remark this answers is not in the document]
KhalilBellakrid: So for djinni generated files we only have : `std::experimental::optional` and it's the one here: `core/utils/optional.hpp`
And generally, in Libcore we use mainly `utils/Optional`, never used or seen `std::optional` from our side !

core/src/bitcoin/BitcoinLikeExtendedPublicKey.hpp
```diff
@@ -31,28 +31,30 @@
...
-        class BitcoinLikeExtendedPublicKey : public api::BitcoinLikeExtendedPublicKey {
+        using BitcoinExtendedPublicKey = AbstractExtendedPublicKey<api::BitcoinLikeNetworkParameters>;
+        class BitcoinLikeExtendedPublicKey : public BitcoinExtendedPublicKey, public api::BitcoinLikeExtendedPublicKey {
         public:
             BitcoinLikeExtendedPublicKey(const api::Currency& params,
                                          const DeterministicPublicKey& key, const DerivationPath& path = DerivationPath("m/"));
-            virtual std::shared_ptr<api::BitcoinLikeAddress> derive(const std::string &path) override;
-            virtual std::shared_ptr<BitcoinLikeExtendedPublicKey> derive(const DerivationPath& path);
+            std::shared_ptr<api::BitcoinLikeAddress> derive(const std::string &path) override;
+            std::shared_ptr<BitcoinLikeExtendedPublicKey> derive(const DerivationPath& path);
```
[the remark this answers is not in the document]
KhalilBellakrid: Hmm I don't see a reason why do that, we don't have any implicit conversion on `DerivationPath` since its string based ctor is explicit ... unless you have another argument that I missed ? 🤔

core/src/database/migrations.cpp
```diff
@@ -381,5 +381,51 @@ namespace ledger {
...
+
+            sql << "CREATE TABLE ripple_operations("
+                    "uid VARCHAR(255) PRIMARY KEY NOT NULL REFERENCES operations(uid) ON DELETE CASCADE,"
+                    "transaction_uid VARCHAR(255) NOT NULL REFERENCES ripple_transactions(transaction_uid),"
+                    "transaction_hash VARCHAR(255) NOT NULL"
+                    ")";
+        }
+        
+        template <> void rollback<6>(soci::session& sql) {
+            sql << "DROP TABLE ripple_currencies";
```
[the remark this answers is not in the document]
KhalilBellakrid: Yes good catch !

core/src/math/Base58.cpp
```diff
@@ -35,34 +35,44 @@
...
 #include <functional>
-#include <utils/hex.h>
 #include <crypto/Keccak.h>
 
 static const std::string DIGITS = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+static const std::string RIPPLE_DIGITS = "rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz";
 static const ledger::core::BigInt V_58(58);
 
-static void _encode(const ledger::core::BigInt& v, std::stringstream& ss) {
+static std::string getBase58Dictionary(const std::string &networkIdentifier) {
```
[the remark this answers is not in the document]
KhalilBellakrid: It depends how we look at it, because first it is specific to Base58 encoding so it can have it's place in this file, second if we follow same logic we should also get rid of EIP55 encoding which is specific to ETH ! 
What we can do when re-architecturing libcore is having specific implementations per coin family belonging to this family's module

core/src/math/Base58.cpp
```diff
@@ -98,35 +108,45 @@ std::string ledger::core::Base58::encodeWithEIP55(const std::string &address) {
...
+            auto digitIndex = getBase58Dictionary(networkIdentifier).find(c);
             if (digitIndex != std::string::npos) {
                 intData = (intData * V_58) + BigInt((unsigned int)digitIndex);
             } else {
                 throw Exception(api::ErrorCode::INVALID_BASE58_FORMAT, "Invalid base 58 format");
             }
         }
     }
+
+    //For XRP, hash160 is preceeded by a null byte prefix
```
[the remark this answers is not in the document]
KhalilBellakrid: So the code right above is used by BTC & ETH families but prefixing with null byte for XRP does not have same rules, so I preferred put it after for many reasons: 

- avoid introducing regressions in BTC & ETH,
- for the moment, XRP addresses never start with `1` but if it becomes, the code will still be ok, which is not the case if you put it before (not without risking introducing regs),
- To be honest, I found it also more understandable like this !

core/src/wallet/common/api_impl/OperationApi.cpp
```diff
@@ -135,6 +138,13 @@ namespace ledger {
             return std::make_shared<EthereumLikeOperation>(shared_from_this());
         }
 
+        std::shared_ptr<api::RippleLikeOperation> OperationApi::asRippleLikeOperation() {
+            if (getWalletType() != api::WalletType::RIPPLE) {
```
phaazon: This is the kind of thing the new architecture must not have: if we want to downcast (and I hope we’ll settle on another design), the fact the downcast can fail should be conveyed in the API (with a `Result` / `Option` type).

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -272,9 +272,9 @@ namespace ledger {
                                             std::shared_ptr<SynchronizationBuddy> buddy) {
 
                 buddy->logger->info("SYNC BATCHES");
-                //For ethereum like wallets, one account corresponds to one ETH address,
+                //For ethereum and ripple like wallets, one account corresponds to one ETH address,
```
phaazon: […] to one ETH or XRP address, respectively

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -272,9 +272,9 @@ namespace ledger {
...
                 //so ne need to discover other batches
-                auto isETHWallet = buddy->wallet->getWalletType() == api::WalletType::ETHEREUM;
+                auto hasMultipleAddresses = buddy->wallet->getWalletType() == api::WalletType::BITCOIN;
```
phaazon: I can see the complexity growing… 😨

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -290,7 +290,7 @@ namespace ledger {
                     //But we may want to force sync of accounts within KEYCHAIN_OBSERVABLE_RANGE
                     auto discoveredAddresses = currentBatchIndex * buddy->halfBatchSize;
                     auto lastDiscoverableAddress = buddy->configuration->getInt(api::Configuration::KEYCHAIN_OBSERVABLE_RANGE).value_or(buddy->halfBatchSize);
-                    if (!isETHWallet && (!done || (done && hadTransactions) || lastDiscoverableAddress > discoveredAddresses)) {
+                    if (hasMultipleAddresses && (!done || (done && hadTransactions) || lastDiscoverableAddress > discoveredAddresses)) {
```
phaazon: Because you changed this line, I guess it’s time to clean it, especially according to Boole logic!

```
!done || (done && hadTransactions) == !done || hadTransactions
```

😉

core/src/wallet/currencies.hpp
```diff
@@ -71,6 +71,8 @@ namespace ledger {
             extern LIBCORE_EXPORT const api::Currency ETHEREUM;
             extern LIBCORE_EXPORT const api::Currency ETHEREUM_CLASSIC;
             extern LIBCORE_EXPORT const api::Currency ETHEREUM_ROPSTEN;
+            extern LIBCORE_EXPORT const api::Currency ETHEREUM_CLASSIC;
```
phaazon: Duplicated line, _my sir!_

core/src/wallet/pool/database/CurrenciesDatabaseHelper.cpp
```diff
@@ -182,7 +197,24 @@ void ledger::core::CurrenciesDatabaseHelper::getAllCurrencies(soci::session &sql
 
                 break;
             };
-            case api::WalletType::RIPPLE:break;
+            case api::WalletType::RIPPLE: {
+
```
phaazon: Extra empty line here.

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -0,0 +1,414 @@
...
+        }
+
+        int RippleLikeAccount::putTransaction(soci::session &sql,
+                                              const RippleLikeBlockchainExplorerTransaction &transaction) {
+            auto wallet = getWallet();
+            if (wallet == nullptr) {
+                throw Exception(api::ErrorCode::RUNTIME_ERROR, "Wallet reference is dead.");
+            }
+
+            if (transaction.block.nonEmpty())
```
phaazon: The previous `if` at line _104_ uses curly braces for a single statement; I think you should stick to this coding style and use curly braces here as well. On a more general note, I think omitting curly braces can mislead people from time to time so we should stick to using curly braces for block constructs with one-liners.

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -0,0 +1,414 @@
...
+                operation.amount = transaction.value;
+                operation.type = api::OperationType::SEND;
+                operation.refreshUid();
+                if (OperationDatabaseHelper::putOperation(sql, operation)) {
+                    emitNewOperationEvent(operation);
+                }
+                result = FLAG_NEW_TRANSACTION;
+            }
+
+            if (_accountAddress == transaction.receiver) {
```
phaazon: You could use `else if` here to prevent a non-needed branching. Also, question: what happens if none of the branches are active? Is it possible in theory? If not, then we might want to return something indicating that the transaction is ill-formed?

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+        std::string RippleLikeAccount::getRestoreKey() {
+            return _keychain->getRestoreKey();
+        }
+
+        void RippleLikeAccount::broadcastRawTransaction(const std::vector<uint8_t> &transaction,
+                                                        const std::shared_ptr<api::StringCallback> &callback) {
+            _explorer->pushTransaction(transaction).map<std::string>(getContext(),
+                                                                     [](const String &seq) -> std::string {
+                                                                         //TODO: optimistic update
```
phaazon: I suggest to remove the TODO by transforming it into a Jira issue. :) (also, I have no idea what optimistic update is)

core/src/wallet/ripple/database/RippleLikeTransactionDatabaseHelper.cpp
```diff
@@ -0,0 +1,137 @@
...
+                tx.block = block;
+            }
+
+            return true;
+        }
+
+        bool RippleLikeTransactionDatabaseHelper::transactionExists(soci::session &sql,
+                                                                    const std::string &rippleTxUid) {
+            int32_t count = 0;
+            sql << "SELECT COUNT(*) FROM ripple_transactions WHERE transaction_uid = :rippleTxUid", use(rippleTxUid), into(
```
phaazon: While this is technically correct, the performance of this call will depend on the RDBMS’ optimizer (and in the case of SQLite… 🙄). What you could do is simply `SELECT 1` and check if the return result-set is empty or not.

core/src/wallet/ripple/explorers/ApiRippleLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,175 @@
...
+        Future<String>
+        ApiRippleLikeBlockchainExplorer::pushLedgerApiTransaction(const std::vector<uint8_t> &transaction) {
+            throw make_exception(api::ErrorCode::MISSING_DERIVATION, "Missing implementation of pushTransaction method for Ripple API");
+        }
+
+        Future<std::shared_ptr<BigInt>>
+        ApiRippleLikeBlockchainExplorer::getBalance(const std::vector<RippleLikeKeychain::Address> &addresses) {
+
+            if (addresses.size() != 1) {
+                throw make_exception(api::ErrorCode::INVALID_ARGUMENT, "ApiRippleLikeBlockchainExplorer::getBalance can called only with one address");
```
phaazon: There’s a missing verb in that sentence! 😋 . Maybe change the error message to `"[…] can only be called with […]"`.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,282 @@
...
+                DedicatedContext(context),
+                RippleLikeBlockchainExplorer(configuration, {api::Configuration::BLOCKCHAIN_EXPLORER_API_ENDPOINT}) {
+            _http = http;
+            _parameters = parameters;
+        }
+
+
+        Future<std::shared_ptr<BigInt>>
+        NodeRippleLikeBlockchainExplorer::getBalance(const std::vector<RippleLikeKeychain::Address> &addresses) {
+            //TODO: multiple accounts balances ?
```
phaazon: What do you mean?

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,282 @@
...
+                                     "Can only get transactions for 1 address from Ripple Node, but got {} addresses", addresses.size());
+            }
+            NodeRippleLikeBodyRequest bodyRequest;
+            bodyRequest.setMethod("account_tx");
+            bodyRequest.pushParameter("account", addresses[0]);
+            auto requestBody = bodyRequest.getString();
+            return _http->POST("", std::vector<uint8_t>(requestBody.begin(), requestBody.end()))
+                    .template json<TransactionsBulk, Exception>(
+                            LedgerApiParser<TransactionsBulk, RippleLikeTransactionsBulkParser>())
+                    .
```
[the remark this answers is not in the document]
phaazon: WTF happened here? 😆

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        FuturePtr<Block> NodeRippleLikeBlockchainExplorer::getCurrentBlock() const {
+            NodeRippleLikeBodyRequest bodyRequest;
+            bodyRequest.setMethod("ledger");
+            bodyRequest.pushParameter("ledger_index", std::string("validated"));
+            auto requestBody = bodyRequest.getString();
+            return _http->POST("", std::vector<uint8_t>(requestBody.begin(), requestBody.end()))
+                    .template json<Block, Exception>(LedgerApiParser<Block, RippleLikeBlockParser>())
+                    .
```
phaazon: Ditto!

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,282 @@
...
+                            template mapPtr<Block>(getExplorerContext(),
+                                                   [](const Either<Exception, std::shared_ptr<Block>> &result) {
```
phaazon: I like that pattern, and that’s just `Result::unwrap()` in Rust (on a left, it panics, which is somewhat the same as an exception, but we don’t like exceptions in Rust). Maybe we can add that to `Either` and use it here? :)

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.h
```diff
@@ -0,0 +1,172 @@
...
+            NodeRippleLikeBodyRequest() {
+                //Document should be defined as object
+                _document.SetObject();
+                _params = rapidjson::Value(rapidjson::kObjectType);
+                //_params.
+            };
+
+            NodeRippleLikeBodyRequest &setMethod(const std::string &method) {
+                //When setting method always clear everything
+                //_document.Clear();
```
phaazon: Maybe it’s a forgotten line? If not, I think you should remove it.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.h
```diff
@@ -0,0 +1,172 @@
...
+                vParam.SetInt64(value);
+                _params.AddMember(vKeyParam, vParam, allocator);
+                return *this;
+            };
+
+            std::string getString() {
+                rapidjson::Document::AllocatorType &allocator = _document.GetAllocator();
+                rapidjson::Value container(rapidjson::kArrayType);
+                container.PushBack(_params, allocator);
+                //container.AddMember();
```
phaazon: Ditto.

core/src/math/Base58.cpp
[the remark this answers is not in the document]
KhalilBellakrid: And where this function would live ? In same file ?

core/src/ripple/RippleLikeExtendedPublicKey.cpp
```diff
@@ -0,0 +1,97 @@
...
+            return std::make_shared<RippleLikeExtendedPublicKey>(currency, k, p);
+        }
+
+        std::shared_ptr<RippleLikeExtendedPublicKey>
+        RippleLikeExtendedPublicKey::fromBase58(const api::Currency& currency,
+                                                  const std::string& xpubBase58,
+                                                  const Option<std::string>& path) {
+            auto& params = currency.rippleLikeNetworkParameters.value();
+            DeterministicPublicKey k = RippleExtendedPublicKey::fromBase58(currency, params, xpubBase58, path);
+            return std::make_shared<ledger::core::RippleLikeExtendedPublicKey>(currency, k, DerivationPath(path.getValueOr("m")));
```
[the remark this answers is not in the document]
KhalilBellakrid: Hmm it's standard BIPs, no magic, but I will put something for it

core/src/wallet/common/api_impl/OperationApi.cpp
KhalilBellakrid: Yes it would be better having a `Result` type

core/src/math/Base58.cpp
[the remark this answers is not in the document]
phaazon: Ok, after a live discussion, I got why you did that. LGTM for me then.

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
KhalilBellakrid: Why not 👍

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
KhalilBellakrid: Actually, it's the opposite, we have networks/blockchains that will support multiple addresses/account (like UTXO based ones) and others having 1 address/account (like XRP, ETH etc..)

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
KhalilBellakrid: It should explicitly stay like it is because it is straightforward to understand each condition which leads to trigger the next synchro.

core/src/wallet/currencies.hpp
KhalilBellakrid: Ouuuuh nice catch good sir ! 🙇

core/src/wallet/ripple/RippleLikeAccount.cpp
KhalilBellakrid: This is done in purpose, you can go through both branches (you send to yourself) or none of them (if transaction does not belong to you)

core/src/wallet/ripple/RippleLikeAccount.cpp
KhalilBellakrid: I created JIRA task, but I leave the comment (don't forget that we are open source :trollface: )

core/src/wallet/ripple/database/RippleLikeTransactionDatabaseHelper.cpp
KhalilBellakrid: Hmm not really sure about that, first `SELECT 1` will expand column list and then pull all data, second if we are talking about `COUNT(*)` and `COUNT(1)` there is no difference in term of performance, this assumption might be true in old RDBMS

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
KhalilBellakrid: I removed it, those were thoughts at the beginning of implementation :trollface:

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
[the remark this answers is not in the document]
KhalilBellakrid: Oh god damn it ! I think I used Clion's code formatter and he f***ed up everything 😆

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
KhalilBellakrid: Great idea !

core/src/database/migrations.cpp
```diff
@@ -381,5 +381,51 @@ namespace ledger {
             // ETH currencies
             sql << "DROP TABLE ethereum_currencies";
         }
+
+        template <> void migrate<6>(soci::session& sql) {
+
+            sql << "CREATE TABLE ripple_currencies("
+                    "name VARCHAR(255) PRIMARY KEY NOT NULL REFERENCES currencies(name) ON DELETE CASCADE ON UPDATE CASCADE,"
+                    "identifier VARCHAR(255) NOT NULL,"
+                    "xpub_version VARACHAR(255) NOT NULL,"
```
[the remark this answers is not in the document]
KhalilBellakrid: Haha it is back again :trollface:

core/idl/wallet/common/currency.djinni
```diff
@@ -42,4 +43,6 @@ Currency = record {
     bitcoinLikeNetworkParameters: optional<BitcoinLikeNetworkParameters>;
     # Optional EthereumLikeNetworkParameters, for more details refer to EthereumLikeNetworkParameters doc
     ethereumLikeNetworkParameters: optional<EthereumLikeNetworkParameters>;
+    #Optional EthereumLikeNetworkParameters, for more details refer to EthereumLikeNetworkParameters doc
```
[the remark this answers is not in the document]
KhalilBellakrid: Good catch !

PR #149, opened 2019-02-18
Move password change implementation from the API interface to the WalletPool implementation.

phaazon:
Also, support preferences backend password change.

### DIFF
--- a/core/src/wallet/pool/WalletPool.cpp
+++ b/core/src/wallet/pool/WalletPool.cpp
@@ -476,7 +476,20 @@ namespace ledger {
                 self->logger()->debug("Finish erasing data of WalletPool : {}",name);
                 return Future<api::ErrorCode>::successful(api::ErrorCode::FUTURE_WAS_SUCCESSFULL);
             });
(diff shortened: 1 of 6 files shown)


core/src/wallet/pool/WalletPool.cpp
```diff
@@ -476,7 +476,20 @@ namespace ledger {
                 self->logger()->debug("Finish erasing data of WalletPool : {}",name);
                 return Future<api::ErrorCode>::successful(api::ErrorCode::FUTURE_WAS_SUCCESSFULL);
             });
+        }
 
+        Future<api::ErrorCode> WalletPool::changePassword(
```
pollastri-pierre: It's kind of weird to have a future which should return an error code as a success, why don't you use ledger::core::Unit type ?

phaazon: Yep, I agree. As discussed on slack, this is due to a limitation of our current djinni generation code that doesn’t enjoy `Unit` very much… :(

PR #150, opened 2019-02-18
Fix/get fees stacktrace

phaazon:
That should fix the stack overflow bug. I need to test a bit so don’t merge to fast. ;)

screenshot 2019-02-18 at 16 12 43


phaazon:
Also, the script is not completely usable for long-running devs (I will update it so that it uses already checked out node bindings / Live Desktop, otherwise it’s a bit boring to use right now).

phaazon:
`HODL`, I really can’t figure out why the transactions are behaving like they’ve got Parkinson.

phaazon:
screenshot 2019-02-19 at 11 07 33

phaazon:
There’s no comment yet and no one has reviewed. Because it’s a nasty bug and I have to inquire more, I’ll close this PR and will open a new one when I have more hindsight and something that fixes the bug.

PR #151, opened 2019-02-20
Implement ERC20 token balance history retrieval.

phaazon:


core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -121,6 +179,48 @@ namespace ledger {
...
+        ) {
+            auto localAccount = _account.lock();
+            soci::session sql (localAccount->getWallet()->getDatabase()->getPool());
+            soci::rowset<soci::row> rows = (sql.prepare << "SELECT op.uid, op.ethereum_operation_uid, op.account_uid,"
+                    " op.type, op.hash, op.nonce, op.value,"
+                    " op.date, op.sender, op.receiver, op.input_data,"
+                    " op.gas_price, op.gas_limit, op.gas_used, op.status"
+                    " FROM erc20_operations AS op"
+                    " WHERE op.account_uid = :account_uid AND (op.date BETWEEN ':start_date_uid' AND ':end_date_uid')"
+                    " ORDER BY op.date ASC", soci::use(_accountUid));
```
phaazon: You dummy, missing some `soci::use` there!

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -72,15 +73,72 @@ namespace ledger {
...
+
+        api::ERC20LikeBalanceHistory ERC20LikeAccount::getBalanceHistoryFor(
+            const std::chrono::system_clock::time_point& startDate,
+            const std::chrono::system_clock::time_point& endDate,
+            api::TimePeriod period
+        ) {
+            // guard against bad arguments
+            if (startDate >= endDate) {
+                throw make_exception(
+                    api::ErrorCode::INVALID_DATE_FORMAT,
```
teams2ua: INVALID_ARGUMENT more suitable, but we didn't distinguish them anyway...

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -72,15 +73,72 @@ namespace ledger {
...
+                    "Start date should be strictly greater than end date"
+                );
+            }
+
+            std::vector<std::shared_ptr<api::BigInt>> balances;
+            auto currentBalance = BigInt::ZERO;
+            auto nextDate = DateUtils::incrementDate(startDate, period);
+            auto operations = getSortedOperationsRange(startDate, endDate);
```
teams2ua: You will always have Balance 0 at startDate

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
phaazon: I just do it the way it’s done everywhere else (i.e. Bitcoin and Ethereum).

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
phaazon: Yes. (I don’t know what else to do, is there a problem with that? :D)

teams2ua: I think this is not users expect, we should calculate what is the balance at startPoint as the first element of balances array. Either it is one request to DB or manually - in the same loop.

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
teams2ua: ok

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
phaazon: If you compute the balance at `startPoint`, you’re actually not using `startPoint` because it means `startPoint = 0`. I don’t really see why Live would need to compute the balance with a `startPoint` actually (they only know the `endDate` to me, which represents the pointer at which the balance should be displayed). Actually, with hindsight, I don’t really get neither of the solution. I’m confused. :D

phaazon: Talked on slack, I’ll fix that. :)

phaazon: @teams2ua that should be fixed now. Thanks for spotting that issue out!

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -126,6 +190,47 @@ namespace ledger {
             return result;
         }
 
+        std::vector<std::shared_ptr<api::ERC20LikeOperation>>
+        ERC20LikeAccount::getSortedOperationsRange(
+            const std::chrono::system_clock::time_point& endDate
+        ) {
+            auto localAccount = _account.lock();
+            soci::session sql (localAccount->getWallet()->getDatabase()->getPool());
+            soci::rowset<soci::row> rows = (sql.prepare << "SELECT op.uid, op.ethereum_operation_uid, op.account_uid,"
```
teams2ua: [OPT] we only need type and value, no need to ask for all the fields.

phaazon: What do you mean?

teams2ua: SELECT op.date, op.value, op.type FROM erc20_operations ....

phaazon: If you do that, you cannot return a correct and coherent `ERC20Operation`.

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -126,6 +190,47 @@ namespace ledger {
...
+                    " op.type, op.hash, op.nonce, op.value,"
+                    " op.date, op.sender, op.receiver, op.input_data,"
+                    " op.gas_price, op.gas_limit, op.gas_used, op.status"
+                    " FROM erc20_operations AS op"
+                    " WHERE op.account_uid = :account_uid AND op.date <= ':end_date_uid'"
+                    " ORDER BY op.date ASC", soci::use(_accountUid), soci::use(endDate));
+            std::vector<std::shared_ptr<api::ERC20LikeOperation>> result;
+            for (auto& row : rows) {
```
KhalilBellakrid: Could you please refacto this code with the one used in `ERC20LikeAccount::getOperations` ?

phaazon: I still think this is a bit a pity to throw away some code that would reduce the network’s bandwidth, the client’s CPU time spending sorting data, etc., solely for refactoring concerns, but seems @teams2ua and you seem to have made your mind about that and that I’m the only one thinking otherwise, I’ll remove that code and switch to `getOperations` then.

teams2ua: I fill you pain, but I believe this change will allow us to have even more improved performance in the future, if we will need this. And it gives us less code (=less bugs) right now.

core/idl/wallet/ethereum/erc20.djinni
```diff
@@ -1,59 +1,72 @@
...
     # Maximum number of decimals (used to show the human representation of the currency).
-	numberOfDecimal: i32;
+    numberOfDecimal: i32;
+}
+
+# ERC20-like balance history.
+ERC20LikeBalanceHistory = record {
+    # Period used when representing the history.
+    period: TimePeriod;
+    # Start date of the history.
```
teams2ua: Can you please include into the comments are startDate and endDate included?

phaazon: Sure.

core/idl/wallet/ethereum/erc20.djinni
```diff
@@ -1,59 +1,72 @@
...
     # Address of the smart contract for this ERC20.
-	contractAddress: string;
+    contractAddress: string;
     # Maximum number of decimals (used to show the human representation of the currency).
-	numberOfDecimal: i32;
+    numberOfDecimal: i32;
+}
+
+# ERC20-like balance history.
+ERC20LikeBalanceHistory = record {
```
teams2ua: What is the purpose of this record?
The client knows first three parameters anyway, so he is only interested in the list

core/src/utils/DateUtils.hpp
```diff
@@ -28,12 +28,13 @@
  * SOFTWARE.
  *
  */
-#ifndef LEDGER_CORE_DATEPARSER_HPP
-#define LEDGER_CORE_DATEPARSER_HPP
+
+#pragma once
```
teams2ua: thanks

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -74,15 +75,95 @@ namespace ledger {
...
+            return api::ERC20LikeBalanceHistory(
+                api::TimePeriod::DAY,
+                startDate,
+                endDate,
+                balances
+            );
+        }
+
+        BigInt ERC20LikeAccount::accumulateBalanceWithOperation(
+            BigInt balance,
```
teams2ua: const BigInt& balance, we can avoid additional copy here.

core/idl/wallet/ethereum/erc20.djinni
phaazon: Initially, this was intended to handle missing information and clamping (i.e. when you ask for something that is not available, the algorithm should clamp back). Though I will remove it as it complicates types for a situation people might not be interested in.

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
phaazon: The joy of not having a move-as-default language…

PR #153, opened 2019-02-22
Fix generateBindingsRN.sh to add the Java interface.

phaazon:


tools/generateBindingsRN.sh
```diff
@@ -1,11 +1,14 @@
 #!/usr/bin/env bash
 
 PACKAGE_NAME=ledgercore
+
+JAVA_IFACE=api/core/java
```
KhalilBellakrid: Why don't you generate files directly in `DEST_ANDROID_JAVA_IFACE` ?

phaazon: As I haven’t written that code, I don’t know why it’s done this way. Currently, that Java code is generated when we do a `make` IIRC, not when we call that script.

KhalilBellakrid: Hmm I don't really get it, `make` is not generating any code and specially not Java code, plus `api/core/java` is not used anymore, so you can safely use `DEST_ANDROID_JAVA_IFACE`
Or have I miss something ? 🤔

phaazon: Hm, I’ll try to see whether I can remove it from our side.

phaazon: Currently I’m on a bug fix but as soon as it’s fixed I’ll address this comment and update. In the meantime, I’m putting this on `HODL`.

phaazon: Ok, this cannot be fixed right now because those files are generated by `generate_interfaces.sh`, not this script. So let’s merge that and if we want to clean those files, we should open another PR as this is a bigger refactoring concern (and that refactoring is not about fixing, so let’s fix first and refactor after) @KhalilBellakrid

KhalilBellakrid: `JAVA_IFACE` is not used anywhere, can you get please get rid of it ?

phaazon: It’s used at the bottom of the script.

phaazon: Ok, so we remove `JAVA_IFACE` and we generate directly the iface into the `DEST_ANDROID_JAVA_IFACE` on the djinni command.

PR #154, opened 2019-02-22
Check pointer after weak_ptr locking

teams2ua:
Check pointers after weak_ptr.lock()


core/src/debug/RotatingEncryptableSink.cpp
```diff
@@ -125,6 +127,9 @@ namespace ledger {
 
         void RotatingEncryptableSink::_rotate() {
             auto resolver = _resolver.lock();
+			if (!resolver) {
```
KhalilBellakrid: What happened with all your indentations ? it seems broken everywhere 😞

phaazon:
(also, indentation is broken)

core/src/debug/RotatingEncryptableSink.cpp
teams2ua: Oh.... let me check

teams2ua:
> Can you provide more hindsight about why you need this? i.e. if there’s an issue with a use-after-free, this PR band-aids the consequences, it doesn’t fix the cause.
It is better, because now we can have a stack-trace not just a crash. Will allow us identify problem earlier. Maybe it will help us point the problem that some people have with android

phaazon:
Fair enough then. Thanks for doing this! I’ll approve after you’ve addressed El @KhalilBellakrid’s comment.

PR #159, opened 2019-02-26
Fix android build on MacOSX.

phaazon:
I guess the include/darwin stuff is fixed on some implementation (it’s
not even linked to Mojave vs. not Mojave). Looks like that it will
depend on which implementation of JNI you have. In my case, I have the
default one, which has that darwin include dir that needs to be added.

I feel like my soul is gone and my sanity evaporated.


core/src/CMakeLists.txt
```diff
@@ -17,21 +17,10 @@ file(GLOB ledger-core-jni-sources
...
-#                list(APPEND JAVA_JVM_LIBRARY $ENV{JAVA_HOME}/jre/lib/server/libjvm.dylib)
-#            else()
-#                list(APPEND JAVA_INCLUDE_PATH2 $ENV{JAVA_HOME}/include/linux)
-#                list(APPEND JAVA_AWT_LIBRARY $ENV{JAVA_HOME}/jre/lib/i386/libjawt.so)
-#                list(APPEND JAVA_JVM_LIBRARY $ENV{JAVA_HOME}/jre/lib/i386/server/libjvm.so)
-#            endif()
-#        endif ()
-#    endif()
+    if (APPLE)
+        list(APPEND JAVA_INCLUDE_PATH2 $ENV{JAVA_HOME}/include/darwin)
```
KhalilBellakrid: Discussion we had: you don't normally need this !

tools/build_android/android_build.sh
```diff
@@ -21,6 +25,6 @@ if [ "$ARCH" == "armeabi-v7a" ]; then
     export TOOLCHAIN_NAME='android-ndk-r16b-api-21-x86-clang-libcxx'
 fi
 
-cmake -DCMAKE_BUILD_TYPE:STRING=Release -DTARGET_JNI=ON -DCMAKE_TOOLCHAIN_FILE=${POLLY_ROOT}/${TOOLCHAIN_NAME}.cmake ../lib-ledger-core
+cmake . -DCMAKE_BUILD_TYPE:STRING=Release -DTARGET_JNI=ON -DCMAKE_TOOLCHAIN_FILE=${POLLY_ROOT}/${TOOLCHAIN_NAME}.cmake
```
KhalilBellakrid: I think that this one should be `libcore_path`, since we are supposed to launch it from build dir, or you will have to change the line below since it's building current from directory

core/src/CMakeLists.txt
phaazon: Yep we don’t; I’m removing it.

PR #160, opened 2019-02-26
SQLCipher on windows

teams2ua:
Generate one big sqlite.c from sqlcipher repo as described https://sqlite.org/src/doc/trunk/README.md
Link with our own crypto lib.


.circleci/build_lib.sh
```diff
@@ -106,8 +106,6 @@ export POLLY_ROOT=`pwd`/toolchains/polly
 ###
 # Clean
 ###
-echo "=====>Cleaning SQLCipher"
```
phaazon: @KhalilBellakrid why was that needed, again?

CMakeLists.txt
```diff
@@ -20,21 +20,9 @@ list(APPEND INCLUDE_DIRECTORIES core/test/include/)
...
-# step of SQLCipher project (because C compiler's path contains spaces)
-if (NOT MSVC AND NOT MINGW)
-    set(SQLCIPHER TRUE CACHE BOOL "Enable SQLCipher encryption support")
-endif()
-
-if (SQLCIPHER)
-    add_definitions("-DSQLCIPHER")
-    add_definitions("-DSQLITE_HAS_CODEC")
-endif()
+add_definitions("-DSQLITE_HAS_CODEC")
```
phaazon: Please keep the `SQLCIPHER` variable so that we can disable it if we want to.

core/lib/CMakeLists.txt
```diff
@@ -14,11 +14,10 @@ add_subdirectory(snappy)
 add_subdirectory(leveldb)
 
 add_subdirectory(spdlog)
-if (SQLCIPHER)
-    include(ProjectSQLiteCipher)
-else()
-    add_subdirectory(sqlite3)
-endif()
+add_subdirectory(sqlcipher)
```
phaazon: Ditto.

core/lib/soci_sqlite3/CMakeLists.txt
```diff
@@ -14,23 +14,5 @@ add_library(soci_sqlite3 STATIC
...
-            find_library(FOUNDATION Foundation)
-            target_link_libraries(soci_sqlite3 ${FOUNDATION})
-            find_library(SECURITY Security)
-            target_link_libraries(soci_sqlite3 ${SECURITY})
-        endif()
-    endif()
-else()
-    target_include_directories(soci_sqlite3 PUBLIC ../sqlite3 )
-endif()
+target_link_libraries(soci_sqlite3 PUBLIC ${SQLITE_LIB})
```
phaazon: Ditto.

core/test/database/CMakeLists.txt
```diff
@@ -16,13 +16,7 @@ add_executable(ledger-core-database-tests main.cpp pool_tests.cpp query_filters_
...
 
 target_link_libraries(ledger-core-database-tests gtest gtest_main)
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    # We need this otherwise link fails (for Linux),
-    # this is needed since we build static library for tests,
-    # not needed if we link against dynamic
-    # TODO: needs more investigation
-    target_link_libraries(ledger-core-database-tests "${CMAKE_SOURCE_DIR}/core/lib/sqlcipher/lib/${CMAKE_STATIC_LIBRARY_PREFIX}sqlcipher${CMAKE_STATIC_LIBRARY_SUFFIX}")
-endif()
+
```
phaazon: Extra whitespace.

core/test/database/database_soci_proxy_tests.cpp
```diff
@@ -80,6 +80,7 @@ class SociProxyBaseTest : public ::testing::Test {
         _backend = std::make_shared<ProxyBackend>(engine);
         _backend->enableQueryLogging(true);
         auto date = DateUtils::toJSON(std::chrono::system_clock::now());
+        std::remove(date.begin(), date.end(), ':');
```
phaazon: Why is that needed?

core/test/database/sqlcipher_tests.cpp
```diff
@@ -63,16 +64,6 @@ TEST(SQLCipherTest, SanityCheck) {
                 soci::use(value_0), soci::use(value_1);
         session.close();
     }
-    {
```
phaazon: I don’t get why you removed that from the test suite.

core/test/database/sqlcipher_tests.cpp
```diff
@@ -104,12 +95,43 @@ TEST(SQLCipherTest, SanityCheck) {
...
         session.close();
+    }
+}
 
-        session.open(*factory, parameters);
+// Preventing terminating in environments that don't properly define __cplusplus macros
+// TODO: remove this check after we migrate to VS2017 and provide /Zc:__cplusplus option 
+// https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+#if __cplusplus >= 201103L
+TEST(SQLCipherTest, ThrowIfWrongPassword) {
```
phaazon: Hm, I don’t know how to feel about that. Tests shouldn’t be compiler-dependent like that. Also, what is that CPP conditional expression for? I don’t get the part with VS2017.

CMakeLists.txt
teams2ua: If we want to stop encrypt, all we need to do is remove `add_definitions("-DSQLITE_HAS_CODEC")`
then sqlcipher will act exactly like sqlite.

core/test/database/database_soci_proxy_tests.cpp
teams2ua: ":" can't be part of file name on windows, so the test fails.

.circleci/build_lib.sh
KhalilBellakrid: It was needed because we were building this external project before in-source, so when building for another target, it was necessary to clean it, otherwise build would fail, but now I guess we are good to go

core/test/database/sqlcipher_tests.cpp
teams2ua: It was not removed it was moved to separate test that can only be run on not broke compilers. Will elaborate inside the next comment.

core/lib/CMakeLists.txt
teams2ua: no more sqlite3 - we have other implementation of this functionality in sqlcipher.

core/lib/soci_sqlite3/CMakeLists.txt
teams2ua: Ditto `¯\_(ツ)_/¯`

core/test/database/sqlcipher_tests.cpp
teams2ua: soci_core throws exceptions on the destructors. Starting from C++11 throwing exception in destructor should terminate program, unless the destructor was marked as `noexcept(false)` this is what soci_core is doing:
```
#if __cplusplus >= 201103L
#define SOCI_ONCE_TEMP_TYPE_NOEXCEPT noexcept(false)
#else
#define SOCI_ONCE_TEMP_TYPE_NOEXCEPT
#endif

and latter 

class SOCI_DECL once_temp_type
{
public:
...    
    ~once_temp_type() SOCI_ONCE_TEMP_TYPE_NOEXCEPT;
....
```
According to the article, all Visual Studios up to VS2017 15.7 defining __cplusplus as 199711L 
In other words Live Desktop on Windows will terminate if someone will try open encrypted DB with wrong password, don't know what to do here... either use another db driver, or migrate completely to VS2017 ...tuft choose

CMakeLists.txt
phaazon: Then just surround it with a proper variable as it was before. I don’t want to have to edit a file while `-D` on the `cmake` command line invocation can do it.

teams2ua: If this option is not set, Windows build of sqlcipher is failing.

PR #161, opened 2019-02-27
Secp files in build not in src

teams2ua:
This will allow to build lib-core from one source directory with different compilers without cleaning src


phaazon:
I’m merging this as I really need to figure out how to make that Android build work. :(

PR #168, opened 2019-03-09
Add getStatus method on EthereumLikeTransaction

KhalilBellakrid:
- Once merged, I will re-generate bindings and submit corresponding PRs to repos

### DIFF
--- a/core/idl/wallet/ethereum/ethereum_like_wallet.djinni
+++ b/core/idl/wallet/ethereum/ethereum_like_wallet.djinni
@@ -34,6 +34,8 @@ EthereumLikeTransaction = interface +c {
     getValue(): Amount;
     # Get binary data payload.
     getData(): optional<binary>;
+    # Get status of transaction: equals to 1 if succeeded, 0 otherwise
+    getStatus(): i32;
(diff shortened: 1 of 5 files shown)


core/idl/wallet/ethereum/ethereum_like_wallet.djinni
```diff
@@ -34,6 +34,8 @@ EthereumLikeTransaction = interface +c {
     getValue(): Amount;
     # Get binary data payload.
     getData(): optional<binary>;
+    # Get status of transaction: equals to 1 if succeeded, 0 otherwise
+    getStatus(): i32;
```
phaazon: Maybe we can have a better encoding of status? Maybe have a small algebraic type (that would get enum-like foreign implementation in JS for instance) for all the possible status?

KhalilBellakrid: I thought about that but we don't have the exhaustive list of possible status, for the moment explorer is only giving `0` or `1` so a type for that might be over killed (I avoided boolean just to not be forced to change if they return other status than 0/1)

PR #170, opened 2019-03-14
LLC-183: Implement updateWalletConfig on WalletPool

KhalilBellakrid:
Wallet's configuration is provided to Wallet Pool when instantiating this same wallet, then persisted in our DB.
So when recovering a wallet we use same persisted configuration, so basically there is no way to update the configuration, except deleting the wallet and recreating it (SFYL)
We provide now a way to update the wallet's configuration, only fields present in the new configuration will be updated, others remain intact !

**Note**: user should be careful how to use this method:
- A new instance of wallet should be fetched from the Wallet Pool (e.g. through `getWallet` method) to get a wallet instance with the updated configuration,
- User should NOT update some fields (e.g. derivation scheme, keychain engine ...) that are defining the wallet itself (we can imagine in the future having of blacklisted fields and return an Error Code for that)

### DIFF
--- a/core/src/bitcoin/bech32/Bech32Parameters.h
+++ b/core/src/bitcoin/bech32/Bech32Parameters.h
@@ -76,7 +76,7 @@ namespace ledger {
             };
             extern LIBCORE_EXPORT const Bech32Struct getBech32Params(const std::string &networkIdentifier);
             extern LIBCORE_EXPORT const std::vector<Bech32Struct> ALL;
-            static bool insertParameters(soci::session& sql, const Bech32Struct &params);
+            bool insertParameters(soci::session& sql, const Bech32Struct &params);
         }
     }
 }
--- a/core/src/collections/DynamicObject.cpp
+++ b/core/src/collections/DynamicObject.cpp
@@ -184,5 +184,48 @@ namespace ledger {
                 }
             }
         }
+
+        std::shared_ptr<api::DynamicObject> DynamicObject::updateWithConfiguration(const std::shared_ptr<DynamicObject> &configuration) {
+            for (auto &key : configuration->getKeys()) {
+                auto type = configuration->getType(key).value_or(api::DynamicType::UNDEFINED);
+                switch (type) {
+                    case api::DynamicType::STRING:
+                        put(key, configuration->get<std::string>(key).value());
+                        break;
+
+                    case api::DynamicType::DATA:
+                        put(key, configuration->get<std::vector<uint8_t>>(key).value());
+                        break;
+
+                    case api::DynamicType::BOOLEAN:
+                        put(key, configuration->get<bool>(key).value());
+                        break;
+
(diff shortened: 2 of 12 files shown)


teams2ua:
Lets split wallet configuration to two parts - immutable one, that can't change after wallet creation
(keychain scheme etc.), and the part that is set by client afrer wallet building (either getWallet or createWallet) for example explorer URL, batch size. This way we will not need "update" method, so we will be safe of client modify something he shouldn't.
This change will require client change as well, so it is not fast. We can put this in the "future libcore" plan.

core/src/bitcoin/bech32/Bech32Parameters.h
```diff
@@ -76,7 +76,7 @@ namespace ledger {
             };
             extern LIBCORE_EXPORT const Bech32Struct getBech32Params(const std::string &networkIdentifier);
             extern LIBCORE_EXPORT const std::vector<Bech32Struct> ALL;
-            static bool insertParameters(soci::session& sql, const Bech32Struct &params);
+            extern LIBCORE_EXPORT bool insertParameters(soci::session& sql, const Bech32Struct &params);
```
teams2ua: Why you change this?
Now you expose to the client of the library soci::session&, I think we need to avoid this, as it compexify the interface, and this is not ABI safe

core/src/collections/DynamicObject.cpp
```diff
@@ -184,5 +184,48 @@ namespace ledger {
                 }
             }
         }
+
+        std::shared_ptr<api::DynamicObject> DynamicObject::updateWithConfiguration(const std::shared_ptr<DynamicObject> &configuration) {
```
teams2ua: [FAR] We should redo all such method with visitors.

core/test/integration/wallet_tests.cpp
```diff
@@ -153,4 +153,31 @@ TEST_F(WalletTests, CreateAccountBug) {
     };
     loop(0);
     dispatcher->waitUntilStopped();
+}
+
+TEST_F(WalletTests, ChangeWalletConfig) {
```
teams2ua: Thanks for tests :)

core/src/bitcoin/bech32/Bech32Parameters.h
KhalilBellakrid: To fix link issues, it is now used in another build unit, knowing that `static` method when declared under a namespace have internal linkage.

core/src/collections/DynamicObject.cpp
KhalilBellakrid: Agreed ! 👍

core/src/bitcoin/bech32/Bech32Parameters.h
teams2ua: So just remove static, `bool insertParameters(soci::session& sql, const Bech32Struct &params);`

PR #171, opened 2019-03-14
LLC-181: Wrap optimistic update after tx broadcast in a Try 

KhalilBellakrid:
Avoid throwing when a tx is successfully broadcasted to a blockchain, if optimistic update fails, this is not a real issue because next synchro will trigger the update of local DB

### DIFF
--- a/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
+++ b/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
@@ -553,60 +553,69 @@ namespace ledger {
                 //Store newly broadcasted tx in db
                 //First parse it
                 auto txHash = seq.str();
+                auto optimisticUpdate = Try<int>::from([&] () -> int {
+                    //Get last block from DB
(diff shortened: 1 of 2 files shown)


core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -553,60 +553,69 @@ namespace ledger {
                 //Store newly broadcasted tx in db
                 //First parse it
                 auto txHash = seq.str();
+                auto optimisticUpdate = Try<int>::from([self, transaction, txHash] () -> int {
```
teams2ua: Why just not try{}catch(){}, instead of creating new object, copy self, transaction, txHash and get a mess with call stack during debugging.

KhalilBellakrid: I understand your concern about the stack, but I will use `Try` class to comply with what we mostly do but I will capture by reference in this context it's ok to do so

PR #183, opened 2019-03-26
Fix race condition on AbstractAccount.events

teams2ua:


core/src/wallet/common/AbstractAccount.cpp
```diff
@@ -187,16 +187,22 @@ namespace ledger {
         void AbstractAccount::emitEventsNow() {
             auto self = shared_from_this();
             run([self] () {
-                for (auto& event : self->_events) {
+                std::list<std::shared_ptr<api::Event>> events;
+                {
+                    std::lock_guard<std::mutex> lock(self->_eventsLock);
+                    std::swap(events, self->_events);
+                    events = self->_events;
```
phaazon: So first you swap `events` and `self->_events`, which is not really useful since `events` is empty`. I guess you want to remove that `std::swap` line?

phaazon: Also, IIRC, you use a mutex to make a deep copy of current events?

teams2ua: This is a trick to not hold a locked mutex for too long, I can do
```
std::lock_guard<std::mutex> lock(self->_eventsLock);
for (auto& event : self->_events) {
    self->_publisher->post(event);
}
self->_events.clear();
```
But this will hold the mutex for all that time, instead I "copy" all the events to local list and clear self->_events in one operation under lock. As this is swap, it should be fast (if event implement move).
Do you think this is overkill?

phaazon: 1. `self->_events` contains stuff and `events` is empty.
2. After `std::swap`, `events` contains stuff and `self->_events` is empty.
3. After assignation, `events` is empty.
4. Both empty.

teams2ua: Ah, nice catch. Thank you!

phaazon: You’re welcome :)

core/src/wallet/common/AbstractAccount.cpp
```diff
@@ -187,16 +187,21 @@ namespace ledger {
         void AbstractAccount::emitEventsNow() {
             auto self = shared_from_this();
             run([self] () {
-                for (auto& event : self->_events) {
+                std::list<std::shared_ptr<api::Event>> events;
+                {
+                    std::lock_guard<std::mutex> lock(self->_eventsLock);
```
phaazon: This is a solution, but if the list of events is large, this is not a real solution and a lock-free solution, ring buffer or immutable slice lock (i.e. you take the size and prevent other threads to delete elements until you’re done) would be faster. However, I think it’s fine.

teams2ua: Didn't get how the size of the list affect this approach, swap is constant time.

PR #185, opened 2019-03-26
Add the RippleLikeMemo type and support parsing it.

phaazon:


core/src/wallet/ripple/RippleLikeMemo.h
```diff
@@ -0,0 +1,48 @@
...
+namespace ledger {
+    namespace core {
+        // A single memo in a Ripple transaction.
+        struct RippleLikeMemo {
+            std::string data;
+            std::string fmt;
+            std::string ty;
+
+            RippleLikeMemo() = default;
+            RippleLikeMemo(std::string const& data_, std::string const& fmt_, std::string const& ty_);
```
teams2ua: Why do you need this constructor? In worse case it can be initialized as {"a", "b", "c"};

phaazon: This file is going to be removed because I had to add this type as an IDL record (so it’ll get generated automatically).

core/src/database/migrations.cpp
```diff
@@ -442,5 +442,18 @@ namespace ledger {
         template <> void rollback<7>(soci::session& sql) {
             sql << "DROP TABLE bech32_parameters";
         }
+
+        template <> void migrate<8>(soci::session& sql) {
+            sql << "CREATE TABLE ripple_memos("
```
KhalilBellakrid: Could you please add a cascade delete here ?

KhalilBellakrid:
Great job ! 
As discussed, for parsing raw XRP tx, please provide some tests for parsing raw txs with memos 
And also for existing test of raw tx parsing, please add some additional checks on values we get (like value, date, receiver, sender ...)
Thank you !

core/src/database/migrations.cpp
phaazon: Sure I can!

phaazon: Done, on the `ripple_transactions` field, `ON DELETE CASCADE`.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -43,6 +43,29 @@
...
     namespace core {
+        // a helper to write VLE fields
+        bool writeLengthPrefix(BytesWriter& writer, uint64_t size) {
+            // encoding here: https://developers.ripple.com/serialization.html#length-prefixing
+            if (size <= 192) {
+                writer.writeByte(size & 0xFF);
+
+                return true;
+            } else if (size <= 12480) {
+                uint16_t x = 193 + ((((size >> 8) & 0x00FF) - 193) << 8) + (size & 0x00FF);
```
teams2ua: for me this code do something different from the reference one here https://github.com/ripple/rippled/blob/fa57859477441b60914e6239382c6fba286a0c26/src/ripple/protocol/impl/Serializer.cpp#L313
For example for size = 193
ref:
Byte 0 = 193
Byte 1 = 0
ours:
Byte 0 = 64
Byte 1 = 130

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -43,6 +43,29 @@
...
+                writer.writeByte((x >> 8) & 0xFF);
+                writer.writeByte(x & 0xFF);
+
+                return true;
+            } else if (size <= 918744) {
+                uint32_t x = 12481 + (((size >> 16) & 0x0000FF - 241) << 16) + (size & 0x00FF00) + (size & 0x0000FF);
```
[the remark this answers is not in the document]
phaazon: Yep, I guess it is. I already optimized that expression a lot, and I guess it can be optimized again! Thanks!

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
phaazon: That code of theirs seems broken. Since they do a `length -= 193`, the second byte AND mask is wrong compared to the official formula they provide in the official documentation.

>  If the field contains 193 to 12480 bytes of data, the first two bytes indicate the length of the field with the following formula:
> 
> 193 + ((byte1 - 193) * 256) + byte2

So if `length` is two bytes long, `length & 0xFF00` is the first byte padded by 8 bits to the left, to show up, and `length & 0x00FF` is the second one. In order to get the first byte, since it’s padded, we need to move it on 8 bits by moving it 8-bit right. Hence:

```
193 + ((byte1 - 193) * 256) + byte2
193 + ((((length >> 8) & 0x00FF) - 193) << 8) + (length & 0x00FF)
```

You end up with my exact formula. I don’t know why they do it this way, though. Something I’m interested in is: do you see how to change the unit tests to see whether the implementation from this PR is wrong? If I did it wrong, the unit tests should show it :P

teams2ua: I think in the formula from documentation `193 + ((byte1 - 193) * 256) + byte2` `byte1` and `byte2` is the resulting bytes, not the bytes of `length`s and with this understanding they code is correct.

phaazon: Yeah actually that’s weird, it looks like which “byte1” or “byte2” comes first is ambiguous. Obviously, if `193` has defined as `byte1 << 8 + byte2`, `byte1` is `0`, so `(byte1 - 193)` is a mess. I need to check further, then.

phaazon: (The interesting part is that they do signed arthimethic while I do unsigned one, which might be why there’s a difference; I might switch to their implementation because when I switch my implementation to signed arithmetic (basically, replace the & 0xFF with * 256), I end up with the same result).

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -51,16 +51,25 @@ namespace ledger {
...
             } else if (size <= 918744) {
-                uint32_t x = 12481 + (((size >> 16) & 0x0000FF - 241) << 16) + (size & 0x00FF00) + (size & 0x0000FF);
-                writer.writeByte((x >> 16) & 0xFF);
-                writer.writeByte((x >> 8) & 0xFF);
-                writer.writeByte(x);
+                //     l         = 12481 + ((byte1 - 241) * 65536) + (byte2 * 256) + byte3
+                // <=> l - 12481 = ((byte1 - 241) * 65536) + (byte2 * 256) + byte3
+                // <=> byte3 = (l - 12481) & 0xFF
+                // <=> byte2 = ((l - 12481) >> 8) & 0xFF
+                // <=> byte1 = ((l - 12481) >> 16) & 0xFF
```
teams2ua: byte1 = ((l - 12481) >> 16) & 0xFF + 241 ?

phaazon: Ah yeah. And I forgot the 193 above too.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -43,6 +43,38 @@
...
+
+                return true;
+            } else if (size <= 918744) {
+                //     l         = 12481 + ((byte1 - 241) * 65536) + (byte2 * 256) + byte3
+                // <=> l - 12481 = ((byte1 - 241) * 65536) + (byte2 * 256) + byte3
+                // <=> byte3 = (l - 12481) & 0xFF
+                // <=> byte2 = ((l - 12481) >> 8) & 0xFF
+                // <=> byte1 = 241 + ((l - 12481) >> 16) & 0xFF
+                size -= 12481;
+                writer.writeByte(241 + (size >> 16) & 0xFF);
```
teams2ua: According to https://en.cppreference.com/w/cpp/language/operator_precedence `&` has lower priority then +, so the result will be (241 + (size>>16) ) & 0xFF

phaazon: Fixed.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -234,6 +266,40 @@ namespace ledger {
...
+            if (!_memos.empty()) {
+                writer.writeByte(0xF9); // STI_ARRAY (Memos); Type Code = 15, Memos; Field ID = 9
+
+                for (auto& memo : _memos) {
+                    writer.writeByte(0xEA); // STI_OBJECT (Memo); Type Code = 10
+
+                    if (!memo.ty.empty()) {
+                        writer.writeByte(0x7C); // STI_VL (MemoType), 12
+                        auto written = writeLengthPrefix(writer, memo.ty.size());
+                        assert(written);
```
teams2ua: Is this assert from cassert http://www.cplusplus.com/reference/cassert/assert/  ? If yes, what would happened in release mode if the error occured?

core/src/database/migrations.cpp
teams2ua: When we will do select from this table, the memos for one transaction can appear in any order (sql give no garanties that select return records in the same order as it was inserted). But memos field is an array, that means the order of elemets is important, because binary representation of transaction will be different depends of the order of the element, and this is bad. I think we need to have index field here as well.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
phaazon: It is.

In release mode, all `assert` calls disappear.

core/src/database/migrations.cpp
phaazon: [The documentation doesn’t state the order is important](https://developers.ripple.com/transaction-common-fields.html#memos-field). If we really, really want to order them, I think we could simply add primary key (it doesn’t have any right now).

Do we care about the order here?

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
teams2ua: So will have some unknow behavior in this case. Instead we can throw nice exception.

core/src/database/migrations.cpp
teams2ua: In the description of Array Field from https://developers.ripple.com/serialization.html#internal-format . `Arrays contain several object fields in their native binary format in a specific order.` Not sure exactly what does this mean, but IMO this is not good that binary serialization of Transaction from DB can be different, depending on how DB is implemented.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
phaazon: If I were to redo the whole libcore parsing, it wouldn’t have asserts nor exceptions. But proper error handling. In this case, I think it’s okay (the rest of the serializer, for all other currencies, don’t even look at the values read back).

I don’t want to spend a week to redo all the parsers / serializers just because we know they could be enhanced. I think it’s time for another time. :)

core/src/database/migrations.cpp
phaazon: Hm, yeah, that might be an issue then. So I guess I can either add an index field, or just a primary key?

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
teams2ua: ok

core/src/database/migrations.cpp
teams2ua: Maybe we can store binary representation of the whole Memo as one binary field. As we have already implement serialization/deserialization.

phaazon: Nah I think it’s overkill. I’ll go with one of your other solutions. :)

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -43,6 +43,38 @@
 
 namespace ledger {
     namespace core {
+        // a helper to write VLE fields
+        bool writeLengthPrefix(BytesWriter& writer, uint64_t size) {
+            // encoding here: https://developers.ripple.com/serialization.html#length-prefixing
+            if (size <= 192) {
+                writer.writeByte(size & 0xFF);
```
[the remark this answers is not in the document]
phaazon: In theory yes. In practice, we must be careful.

According to [this], about standard conversions:

> Otherwise, if the conversion is between numeric types of the same kind (integer-to-integer or floating-to-floating), the conversion is valid, but the value is implementation-specific (and may not be portable).

In my case, it was there mostly to exhibit what part of `size` we’re interested in. We can clearly see here that we are passing 8-bit thanks to the `0xFF` constant.

You know how I feel about that kind of coercions… If you really think we should get rid of that and, I can remove it for sure.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -43,6 +43,38 @@
...
+
+                return true;
+            } else if (size <= 12480) {
+                //     l       = 193 + ((byte1 - 193) * 256) + byte2
+                // <=> l - 193 = ((byte1 - 193) * 256) + byte2
+                // <=> byte2 = (l - 193) & 0xFF
+                // <=> byte1 = 193 + ((l - 193) >> 8) & 0xFF
+                size -= 193;
+                writer.writeByte(193 + ((size >> 8) & 0xFF));
```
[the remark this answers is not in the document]
phaazon: Same remark.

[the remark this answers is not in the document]
phaazon: (also, it applies only because of the test line 53; otherwise, it would be explicitly needed)

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -43,6 +43,38 @@
...
+
+                return true;
+            } else if (size <= 918744) {
+                //     l         = 12481 + ((byte1 - 241) * 65536) + (byte2 * 256) + byte3
+                // <=> l - 12481 = ((byte1 - 241) * 65536) + (byte2 * 256) + byte3
+                // <=> byte3 = (l - 12481) & 0xFF
+                // <=> byte2 = ((l - 12481) >> 8) & 0xFF
+                // <=> byte1 = 241 + ((l - 12481) >> 16) & 0xFF
+                size -= 12481;
+                writer.writeByte(241 + ((size >> 16) & 0xFF));
```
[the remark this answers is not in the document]
phaazon: Same remark too.

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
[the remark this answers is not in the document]
KhalilBellakrid: I was saying that to simplify the code :trollface: 
I understand, you want emphasize on passing only 8-bit ... alright then

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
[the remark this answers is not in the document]
KhalilBellakrid: That's exactly why I was saying that 😉

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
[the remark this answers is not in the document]
phaazon: I think I could compromise to go into your direction would be to `static_cast<uint8_t>`. It would then prevent a useless mask (if the compiler is smart enough it will remove it… but we don’t actually know) and the `static_cast` should be enough to me to emphasize on the 8-bit.

phaazon:
We shall merge it now. 😋

core/src/wallet/ripple/database/RippleLikeTransactionDatabaseHelper.cpp
```diff
@@ -48,9 +48,12 @@ namespace ledger {
             rowset<row> rows = (sql.prepare << "SELECT  tx.hash, tx.value, tx.time, "
                     " tx.sender, tx.receiver, tx.fees, tx.confirmations, "
                     "block.height, block.hash, block.time, block.currency_name "
+                    "memo.data, memo.fmt, memo.ty "
                     "FROM ripple_transactions AS tx "
                     "LEFT JOIN blocks AS block ON tx.block_uid = block.uid "
-                    "WHERE tx.hash = :hash", use(hash));
+                    "LEFT JOIN ripple_memos AS memo ON memo.transaction_uid = tx.transaction_uid "
```
teams2ua: If you have one transaction and 10 fields in memo, this `select` will return you 10 rows, that will looks like 10 transactions. Right?

phaazon: Not really. Just look at the line 59. It doesn’t return 10 transactions. It returns 10 rows. Those rows are being consumed in `inflateTransaction`.

teams2ua: I see.

phaazon: To explain further, from what I understand, it returns several rows that are _related_ to the same transaction — i.e. because of the `WHERE` — clause. And then the transaction is built by looping over the rows and setting fields there.

It’s always good to have eyes looking at that kind of code to be sure it’s not fucked up, so thank you. :)

PR #187, opened 2019-03-28
Patch ERC20LikeAccount::getBalanceHistoryFor.

phaazon:
This commit introducces a currency-agnostic history balance getter. Yay.


core/src/wallet/common/BalanceHistory.hpp
```diff
@@ -0,0 +1,112 @@
...
+            std::vector<std::shared_ptr<CastValue>> getBalanceHistoryFor(
+                std::chrono::system_clock::time_point const& startDate,
+                std::chrono::system_clock::time_point const& endDate,
+                api::TimePeriod precision,
+                OperationIt operationIt,
+                OperationIt operationEnd,
+                std::function<std::chrono::system_clock::time_point (typename OperationIt::value_type&)> opDate,
+                std::function<api::OperationType (typename OperationIt::value_type&)> opType,
+                std::function<Value (typename OperationIt::value_type&)> opValue,
+                std::function<Option<Value> (typename OperationIt::value_type&)> opFees
```
teams2ua: You only need one function opBalanceChange() with 
```
 if (operationDate <= upperDate) {
                        switch (opType(operation)) {
                            case api::OperationType::RECEIVE: {
                                sum = sum + opValue(operation);
                                break;
                            }
                            case api::OperationType::SEND: {
                                sum = sum - (opValue(operation) + opFees(operation).getValueOr(Value()));
                                break;
                            }
                        }
                    }
```
inside.

phaazon: I don’t get your comment here.

teams2ua:
```
std::vector<std::shared_ptr<CastValue>> getBalanceHistoryFor(
                std::chrono::system_clock::time_point const& startDate,
                std::chrono::system_clock::time_point const& endDate,
                api::TimePeriod precision,
                OperationIt operationIt,
                OperationIt operationEnd,
                std::function<std::chrono::system_clock::time_point (typename OperationIt::value_type&)> opDate,
>>>>       std::function<Value (typename OperationIt::value_type&)> opBalanceChange
            ) {
                if (startDate >= endDate) {
                    throw make_exception(api::ErrorCode::INVALID_DATE_FORMAT,
                                         "Start date should be strictly greater than end date");
                }

                auto lowerDate = startDate;
                auto upperDate = DateUtils::incrementDate(startDate, precision);
                std::vector<std::shared_ptr<CastValue>> values;
                Value sum;

                while (lowerDate <= endDate && operationIt != operationEnd) {
                    auto operation = *operationIt;
                    auto operationDate = opDate(operation);

                    while (operationDate > upperDate && lowerDate < endDate) {
                        lowerDate = DateUtils::incrementDate(lowerDate, precision);
                        upperDate = DateUtils::incrementDate(upperDate, precision);
                        values.emplace_back(std::make_shared<ValueImpl>(sum));
                    }
                    if (operationDate <= upperDate) {
>>>>>                sum += opBalanceChange(operation);
                    }
                    ++operationIt;
                }

                while (lowerDate < endDate) {
                    lowerDate = DateUtils::incrementDate(lowerDate, precision);
                    values.emplace_back(std::make_shared<ValueImpl>(sum));
                }

                return values;
            }
```

phaazon: That makes the function harder to use (you need to pass it a date), and it moves the invariant out of the function to make it implemented by the client of this agnostic getBalanceHistory, right?

teams2ua: I have even better idea- just don't path any functions, let compiler found appropriate functions for specific currency, we just will need to define opBalanceChange and opDate in the namespace where Operation is defined.
1. This will work faster - all resolutions happened on compile time.
2. This will be easy to use - no need to provide something in parameters.

KhalilBellakrid:
Noiiiice !! 👍 
We will need also to replace the ones used by ETH and BTC to get balance history !

core/src/wallet/common/BalanceHistory.hpp
```diff
@@ -0,0 +1,112 @@
...
+#include <api/OperationType.hpp>
+#include <api/TimePeriod.hpp>
+#include <async/Future.hpp>
+#include <utils/DateUtils.hpp>
+#include <utils/Option.hpp>
+
+namespace ledger {
+    namespace core {
+        namespace agnostic {
+            // Get a balance history absade on a time window (inclusive) and currency operations.
```
phaazon: I’m absade.

core/src/wallet/common/BalanceHistory.hpp
teams2ua: > That makes the function harder to use (you need to pass it a date), and it moves the invariant out of the function to make it implemented by the client of this agnostic getBalanceHistory, right?

Fix the code in my upper comment, now I see where is our misunderstanding about date comes from ¯\_(ツ)_/¯

phaazon: I’m not a huge fan of this proposal, as it moves some logic inherent to the computation of sums in the caller, while it’s currently done in the callee. The `switch` case on the `api::OperationType` would be done every time the function is called (possibly written wrong), and some thing for the fees.

Also, I still don’t get the motivation. The current code is really simple to both read and use (you just pass a few lambdas, all implemented with a oneliner). To me, this request looks a bit like overengineering (or please provide me with more details about why you think we should _really_ do what you propose). Can we have your opinion on that, @KhalilBellakrid.

teams2ua: I am not pushing, I will approve,
My motivation is:
If we allow compiler to find opBalanceChange we:
1. Simplify life of the caller - no need to provide any parameters.
2. Make this code faster, as less copy will be made (pushing objects to function is not free), will allow compiler to do more optimisation like inlining.
3. Makes process of adding new currencies more isolated - "integrator" will write opBalanceChange function in the same namespace / file where he define Transaction for the new currency, no need to find where the call to getBalanceHistory is made.
About opBalanceChange instead of 3 function, the motivation is:
1. For some currencies there maybe another mechanism of computing Fees, or they may have exotic operation types. We will make less constraints on the currencies (so getBalanceHistory will be applicable in more scenarious).
Hope this is clear :)

teams2ua: @KhalilBellakrid interested in hearing you oppinion as well, as we will have this dillema in the future as well.

KhalilBellakrid: Hmm indeed both approaches offer something here, but I have to admit that it would be better to go with @teams2ua proposal, the idea of reducing number of params is always interesting, also capturing less objects in lambdas and more importantly leave the logic of balance computation to the method's user, specially that we can have really weird/exotic cases (as Andrii said) typically we will try be precautious about balance computation as an example when we will have the staking feature we will have (probably) transactions with values but those are not moving any funds but just locking them (this is a wild guess but better be safe) ... so it will be easier to support this kind of funny behaviours  !

phaazon: I don’t think that reducing the number of parameters is always interesting (in this case it’s not to me, or please explain me why you think it’s interesting, especially such documented and easy parameters that are used only internally). Especially, those parameters **are in the contract of the function**, while the template solution uses duck typing and then hides the interface.

Anyway, this is generating too much bikeshedding to me, so I’ll remove them and use a similar approach to what suggested @teams2ua. If I understood your concerns correctly, my current implementation is not “easy to use” (I disagree, just look at the implementation) and it might be slower than a generic version (which is an assumption, not a fact, and if we pass a function known at static type, it’s very likely that the compiler will optimize that away).

That’s a lot of assumptions and as I said, I don’t want to bikeshed too much. My personal, engineering vision of things is: make it work fast, then measure and if you think it _must_ be faster, then make it faster. The high-order version wouldn’t generate slow call (especially if optimized by the compiler) but the current discussion has been running for days for something we don’t even have measures / benchmarks of.

I’ll make the compiler-version stuff, then.

teams2ua: @phaazon The performance arguments was "optional", the main point was to have one function opBalanceChange instead of 3 opType, opFees and opValues as it limit the applicability of that algorithm.

KhalilBellakrid: @phaazon the main reason is your solution 'impose' user how to compute the balance and it's not adapted for potential type of ops we might have later (as said above) this replacement will also reduce the number of arguments which is interesting, I'm a user of this method I have 1 function to implement and I know it computes the balance, in your previous implementation, I have to provide 3 functions and I should read your algorithm (even if it's well documented) because I have to check if my use case is covered (for example I know that I have exotic way of computing balance or ops)

phaazon: Yes, for the argument of moving the logics to the client, I agree. Hence, I could have removed some `std::function` and leave only two. The current code still requires you to implement those functions (elsewhere, in a struct).

PR #188, opened 2019-03-29
Downgrade fmt to 5.3.0 

KhalilBellakrid:
Commit tag because of crash occuring on Android `armeabi-v7a` (bug occurs before `442fa1b`)
We probably should open an issue on `fmt`


phaazon:
What bug exactly?

KhalilBellakrid:
@teams2ua that's why I spent some time trying to detect more or less which commit was incriminated :trollface: 
@phaazon there was a crash occurring on PR on Live mobile only on Android `armeabi-v7a` ...

PR #189, opened 2019-03-29
Fix soci usages

teams2ua:
Argument of use() shouldn't be temporaly object.

### DIFF
--- a/core/src/bitcoin/bech32/Bech32Parameters.cpp
+++ b/core/src/bitcoin/bech32/Bech32Parameters.cpp
@@ -99,13 +99,16 @@ namespace ledger {
                         strGenerator.push_back(bigIntG.toString());
                     }
                     strings::join(strGenerator, generator, separator);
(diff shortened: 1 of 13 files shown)


KhalilBellakrid:
Could you please also post your analysis about this issue (or code in `soci` where you found this) ? 
It's weird that we didn't have any issue with it until now ...

teams2ua:
> Could you please also post your analysis about this issue (or code in `soci` where you found this) ?
> It's weird that we didn't have any issue with it until now ...
https://github.com/SOCI/soci/blob/c15b178a44b99ed3ff7fd953837fb97f6314abb7/include/soci/use.h#L27 , here is guy have similar issue
https://github.com/SOCI/soci/issues/700

phaazon:
From what I understand, C++ scope analysis is fucked as hell (we would expect it to create a temporary variable _before_ calling a function for each arguments, calling the function and then destroy the temp objects). Looks like it doesn’t always do that.

Change goods to me.

PR #190, opened 2019-03-29
No throw in function definitions

teams2ua:
Remove a lot of warnings and prepare for C++17 (throw(Something) is removed from C++)


teams2ua:
Should be on develop

PR #191, opened 2019-03-29
Preparing optional for C++17

teams2ua:
std::optional is part of C++17, so to have less confusion with legitime std::exception namespace, put it to very custom namespace, in coming PRs will just do std::ledger_exp::optional = std::optional

### DIFF
--- a/core/src/utils/optional.hpp
+++ b/core/src/utils/optional.hpp
@@ -101,7 +101,7 @@
 
 namespace std{
 
-    namespace experimental{
+    namespace ledger_exp{
 
 // BEGIN workaround for missing is_trivially_destructible
 # if defined TR2_OPTIONAL_GCC_4_8_AND_HIGHER___
@@ -113,8 +113,8 @@ namespace std{
 # elif defined TR2_OPTIONAL_DISABLE_EMULATION_OF_TYPE_TRAITS
     // leave it: the user doesn't want it
 # else
-	template <typename T>
-	using is_trivially_destructible = std::has_trivial_destructor<T>;
+    template <typename T>
+    using is_trivially_destructible = std::has_trivial_destructor<T>;
 # endif
 // END workaround for missing is_trivially_destructible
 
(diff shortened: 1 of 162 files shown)


core/src/utils/optional.hpp
```diff
@@ -101,7 +101,7 @@
 
 namespace std{
 
-    namespace experimental{
+    namespace ledger_exp{
```
phaazon: I think we should remove that from `std` if we decide to change the namespace. I’m not comfortable with intruding in other namespaces… 😕

teams2ua:
> I think this PR would be the right place to decide what we do about `Option` vs. `optional`. It’s been extremely confusing. 😞
The purpouse of this PR was to allow me compile under C++17, but agree we can do better here by switching directly to std::optional (just didn't want to have problem in case there is behavirioal change).
My thoughts about Option and optional - we should have only std::optional objects everywhere and make functions from Option static that accept std::optional as argument, so we will preserve the helpfull part of Option.

core/src/utils/optional.hpp
teams2ua: Yeah, you right. 
I will try to completely move to std::optional.
Looks like it should be safe, I will upgrade compilers on CI in coming PR
https://en.cppreference.com/w/cpp/compiler_support

teams2ua: @phaazon I will do this change in two steps. First I would like temporary merge this commit, it will allow me safely upgrade compilers on CI. Second step I will switch to std::optional.
This way will be easy to understand what exactly cause problems, if they happen.

teams2ua:
> I really don’t like the namespace nesting. Invading the standard library’s namespace is a bit odd here. Why not just sticking to `ledger::…`?

This is "temporary" change, we should get rid of it once C++17.
We have Optional in ledger::core, may add confusion as well.

teams2ua:
Close for now, will reopen later when it will make sense.

PR #192, opened 2019-03-29
Remove throw(Exception) from function definition

teams2ua:
Remove a lot of warnings and prepare for C++17 (throw(Something) is removed from C++)

### DIFF
--- a/core/src/utils/DerivationPath.cpp
+++ b/core/src/utils/DerivationPath.cpp
@@ -50,7 +50,7 @@ namespace ledger {
 
         }
 
-        std::vector<uint32_t> DerivationPath::parse(const std::string &path){
+        std::vector<uint32_t> DerivationPath::parse(const std::string &path) {
             std::string currentNode = "";
             bool hardened = false;
             bool lastCharWasZero = false;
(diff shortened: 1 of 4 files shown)


core/src/utils/DerivationPath.cpp
```diff
@@ -115,12 +115,12 @@ namespace ledger {
             return (uint32_t) _path.size();
         }
 
-        uint32_t DerivationPath::getLastChildNum() const throw(Exception) {
+        uint32_t DerivationPath::getLastChildNum() const{
```
phaazon: Missing space before {.

core/src/utils/DerivationPath.cpp
```diff
@@ -50,7 +50,7 @@ namespace ledger {
 
         }
 
-        std::vector<uint32_t> DerivationPath::parse(const std::string &path) throw(Exception) {
+        std::vector<uint32_t> DerivationPath::parse(const std::string &path){
```
phaazon: Missing space before {.

core/src/utils/DerivationPath.cpp
teams2ua: added

core/src/utils/DerivationPath.cpp
teams2ua: added

PR #194, opened 2019-04-01
Fix encryption to pass RecryptClearValues.

phaazon:


core/src/preferences/PreferencesBackend.cpp
```diff
@@ -194,17 +194,47 @@ namespace ledger {
...
+                auto saltData = std::vector<uint8_t>(salt.cbegin(), salt.cend());
+                auto noCipher = Option<AESCipher>::NONE;
+                putPreferencesChange(batch, noCipher, PreferencesChange(PreferencesChangeType::PUT_TYPE, saltKey, saltData));
+
+                // iterate through all records already present and update them
+                auto it = std::unique_ptr<leveldb::Iterator>(_db->NewIterator(leveldb::ReadOptions()));
+                for (it->SeekToFirst(); it->Valid(); it->Next()) {
+                    // read the clear value
+                    auto value = it->value().ToString();
+                    auto plain = std::vector<uint8_t>(std::begin(value), std::end(value));
```
teams2ua: sometimes you use std::begin(x), sometimes x.cbegin(), lets steek to one style + we shouldn't explicitly specify a namespace `std::` in this case, the compiler should resolve the name based on the argument (so fo collections from stl it would be std and for our own collections it woulb be our own begin, that can be in ledger::core name)

phaazon: I’ll stick to the postfix version for now, nice catch.

phaazon:
@teams2ua comment addressed, thanks!

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -194,17 +194,47 @@ namespace ledger {
...
+                _cipher = AESCipher(rng, password, salt, PBKDF2_ITERS);
+
+                // add the salt to the batch to be written
+                auto saltData = std::vector<uint8_t>(salt.cbegin(), salt.cend());
+                auto noCipher = Option<AESCipher>::NONE;
+                putPreferencesChange(batch, noCipher, PreferencesChange(PreferencesChangeType::PUT_TYPE, saltKey, saltData));
+
+                // iterate through all records already present and update them
+                auto it = std::unique_ptr<leveldb::Iterator>(_db->NewIterator(leveldb::ReadOptions()));
+                for (it->SeekToFirst(); it->Valid(); it->Next()) {
```
teams2ua: Will this loop overwrite the salt data that was put before?

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -194,17 +194,47 @@ namespace ledger {
...
-            }
 
-            // create the AES cipher
-            _cipher = AESCipher(rng, password, salt, PBKDF2_ITERS);
+                // create the AES cipher
+                _cipher = AESCipher(rng, password, salt, PBKDF2_ITERS);
+
+                // add the salt to the batch to be written
+                auto saltData = std::vector<uint8_t>(salt.cbegin(), salt.cend());
+                auto noCipher = Option<AESCipher>::NONE;
```
teams2ua: Make noCipher static

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -194,17 +194,47 @@ namespace ledger {
             unsetEncryption();
 
             auto emptySalt = std::string("");
-            auto pref = getPreferences("__core");
-            auto salt = pref->getString(ENCRYPTION_SALT_KEY, emptySalt);
+            auto saltKey = std::vector<uint8_t>(ENCRYPTION_SALT_KEY.cbegin(), ENCRYPTION_SALT_KEY.cend());
```
teams2ua: Make saltKey static.

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -183,7 +183,7 @@ namespace ledger {
 
         std::string PreferencesBackend::createNewSalt(const std::shared_ptr<api::RandomNumberGenerator>& rng) {
             auto bytes = rng->getRandomBytes(128);
-            return std::string(std::begin(bytes), std::end(bytes));
+            return std::string(bytes.begin(), bytes.end());
         }
 
         void PreferencesBackend::setEncryption(
```
teams2ua: setEncryption is called in resetEncryption, so we iterate throw the DB twice in worste case

phaazon: That happens if someone calls that function on a non-encrypted DB. I’ll add a check for that.

teams2ua: I think if you refactor it a little bit, as I point in other comment, there will be no possibility to iterate twice.

core/src/preferences/PreferencesBackend.cpp
phaazon: Why? 🙂 You want to keep around memory that is likely to be used only once in a whole application lifetime?

core/src/preferences/PreferencesBackend.cpp
phaazon: Same remark as above. That function is likely to be called more than twice and will be called at initialization of a wallet / application / whatever. I might be wrong, but keeping around in memory an `Option<AESCipher>::NONE` seems overkill to me.

core/src/preferences/PreferencesBackend.cpp
phaazon: I think it might. I’m refactoring a bit to put the salt stuff after the loop (I’m curious why the tests pass if the salt got overwritten). Nice catch, I’ll inquire.

core/src/preferences/PreferencesBackend.cpp
teams2ua: We have this code in several places, so I want to be sure we always use the same key. This is just few bytes...

core/src/preferences/PreferencesBackend.cpp
teams2ua: Same argument as I put bellow, this is like a constant. We create enums, for a good reason, here it is the same, how many bytes this is? + the function will work faster :) ( this arguments is the same powerful as the one about memory)

core/src/preferences/PreferencesBackend.cpp
phaazon: Sure, I will!

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -183,87 +183,123 @@ namespace ledger {
...
-                pref->editor()->putString(ENCRYPTION_SALT_KEY, salt)->commit();
-            }
-
-            // create the AES cipher
-            _cipher = AESCipher(rng, password, salt, PBKDF2_ITERS);
+            // setting encryption is akin to resetting with an old password that is empty
+            resetEncryption(rng, "", password);
         }
 
         void PreferencesBackend::unsetEncryption() {
```
teams2ua: Something is not clear for me
1. What would happen, if we have encrypted data and aomebody calls unsetEncryption? 
2. What is the proper way to remove encryption?

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -183,87 +183,123 @@ namespace ledger {
...
             auto saltKey = std::vector<uint8_t>(ENCRYPTION_SALT_KEY.cbegin(), ENCRYPTION_SALT_KEY.cend());
-            auto saltChange = PreferencesChange(PreferencesChangeType::PUT_TYPE, saltKey, newSaltBytes);
-            auto noCipher = Option<AESCipher>::NONE;
-            putPreferencesChange(batch, _cipher, PreferencesChange(PreferencesChangeType::DELETE_TYPE, saltKey, {}));
-            putPreferencesChange(batch, noCipher, PreferencesChange(
-                PreferencesChangeType::PUT_TYPE, saltKey, newSaltBytes
-            ));
+            putPreferencesChange(batch, noCipher, PreferencesChange(PreferencesChangeType::DELETE_TYPE, saltKey, {}));
+
+            if (newCipher.hasValue()) {
```
teams2ua: I think we always need to set this value. Image we have encrypted data + not encrypted salt. Then we decide to change password to "". So we encrypt ALL the values - we did decrypt(data) and decrypt (not encrypted salt). Now we have garbage in our salt.

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -289,6 +325,19 @@ namespace ledger {
...
         }
 
+        std::string PreferencesBackend::getEncryptionSalt() {
+            auto saltKey = std::vector<uint8_t>(ENCRYPTION_SALT_KEY.cbegin(), ENCRYPTION_SALT_KEY.cend());
+
+            // move _cipher around to prevent trying to decrypt the salt
+            auto cipher = std::move(_cipher);
+            _cipher = Option<AESCipher>();
+
+            auto salt = get(saltKey).value_or("");
```
teams2ua: So maybe use raw function instead of get() , to not dancing with _cipher ?

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -183,87 +183,123 @@ namespace ledger {
...
             const std::string& oldPassword,
             const std::string& newPassword
         ) {
-            // turn on encryption with the old password first
-            setEncryption(rng, oldPassword);
+            Option<AESCipher> noCipher;
+            auto newCipher = noCipher;
+            auto salt = getEncryptionSalt();
+
+            if (oldPassword.empty()) {
```
teams2ua: This nasting make me crazy :)

core/src/preferences/PreferencesBackend.cpp
phaazon: 1. The code is pretty straightforward: it disables the the AES cipher and you read garbage. This is “only” needed for unit tests. If you have a better idea I would love to switch to your idea (is there a way to test a private method without going nuts with `friend` shitty C++ concepts? :D)
2. `resetEncryption(old, "")`.

phaazon: I.e. if 2. is too weird, I can add a helper function that does it.

core/src/preferences/PreferencesBackend.cpp
phaazon: :+1:

core/src/preferences/PreferencesBackend.cpp
teams2ua: To be more specific, I think finding cipher for new and old password are independent from each other.

phaazon: In the first place, I had independent blocks for old / new but I switched to this nesting for the reason that it’s less code and encapsulates the logic more tightly.

core/src/preferences/PreferencesBackend.cpp
phaazon: We always need to set a `newCipher`? No we don’t: when decrypting from a ciphertext DB to a plaintext one. In this case, the salt must done be written, either.

phaazon: If you change password to `""`, it will _decrypt_ all the values, not encrypt. The salt will be removed and you will get back to a plaintext db.

teams2ua: I meant we always need to overwrite the salt, because during the loop we watsed it.

teams2ua: I ok, saw the deletion now. My bad

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -183,87 +183,123 @@ namespace ledger {
...
+                if (!newPassword.empty()) {
+                    if (salt.empty()) {
+                        salt = createNewSalt(rng);
+                        newCipher = Option<AESCipher>(AESCipher(rng, newPassword, salt, PBKDF2_ITERS));
+                    } else {
+                        // we want to enable encryption with a given password (new) without decrypting the
```
teams2ua: If we leave the old data not encrypted, how would we know that this particular value is not encrypted and we should skip decryption?

core/src/preferences/PreferencesBackend.cpp
```diff
@@ -183,87 +183,123 @@ namespace ledger {
...
         ) {
-            // turn on encryption with the old password first
-            setEncryption(rng, oldPassword);
+            Option<AESCipher> noCipher;
+            auto newCipher = noCipher;
+            auto salt = getEncryptionSalt();
+
+            if (oldPassword.empty()) {
+                if (!newPassword.empty()) {
+                    if (salt.empty()) {
```
teams2ua: If salt is not empty but password IS, then it means the provided "old" password is probably wrong, right?

phaazon: What case are you talking about? Which password?

teams2ua: oldPassword == "", newPassword != "", salt != "". Is this possible?

core/src/preferences/PreferencesBackend.cpp
teams2ua: After discussion in slack, I release there are use of this function to initially set password when staring application

PR #195, opened 2019-04-02
Implement migration process for bitcoinLikeNetworkParameters

KhalilBellakrid:
This should fix issues with XST because they removed timestamp in their new version of transactions


phaazon:
Thanks for doing that! I’ll read it through, I’m interested!

myselfmyself:
Issue not resolved.  Tried again, same problem.  Get variant called with unexpected parameters

gre:
This is not deployed anywhere yet so it's normal that the fix is not shipped yet.

we'll need on our side to:

- merge this PR
- have a new lib-ledger-core release
- bump lib-ledger-core version in ledger-live-desktop
- do a new ledger-live-desktop release

myselfmyself:
Any updates here?  Looks like work was begun on it but never updated further.  I understand that these things take time, but my XST has been stuck for going on 5 months now.

KhalilBellakrid:
@myselfmyself we have provided this fix on libcore side, we have an other one pending on another dependency and we are waiting in parallel for a new version of XST app which supports new transactions format of XST, so the fix should definitely be in one of the next versions of Live, we will keep you updated about it !
Sorry for the inconvenience !

Kryptoxic:
> @myselfmyself we have provided this fix on libcore side, we have an other one pending on another dependency and we are waiting in parallel for a new version of XST app which supports new transactions format of XST, so the fix should definitely be in one of the next versions of Live, we will keep you updated about it !
> Sorry for the inconvenience !

Is the app being developed by the Stealth team or is it developed by Ledger?

KhalilBellakrid:
I don't have the info about who developed the app initially but the support will be provided by us !

Kryptoxic:
I see. Thanks for letting me know. If you need, I can reach out to the Stealth team via telegram and get them to get in touch with you. Just let me know.

KhalilBellakrid:
For the moment I don't think there is a need for it, but we will let you know if we need to reach them ! 
Thanks for your help ! 👍

PR #196, opened 2019-04-02
Fix XRP transaction parsing

KhalilBellakrid:
Compute and set transaction ID (hash) for signed transactions


core/test/crypto/digest_test.cpp
```diff
@@ -46,13 +47,26 @@ TEST(Digests, SHA256_Strings_to_String) {
     EXPECT_EQ(SHA256::stringToHexHash("abcdefghbcdefghicdefghijdefghijkefghijklfghijklmghijklmnhijklmno"), "2ff100b36c386c65a1afc462ad53e25479bec9498ed00aa5a04de584bc25301b");
 }
 
+// Tested with crypto module on NodeJS implementation (SFYL)
```
phaazon: :ohno:

core/src/crypto/SHA256.cpp
```diff
@@ -32,28 +32,31 @@
...
+        std::vector<uint8_t> SHA256::dataToBytesHash(const void *data, size_t size) {
+            uint8_t hash[SHA256_DIGEST_LENGTH];
+            SHA256_CTX sha256;
+            SHA256_Init(&sha256);
+            SHA256_Update(&sha256, data, size);
+            SHA256_Final(hash, &sha256);
+            return std::vector<uint8_t >(hash, hash + SHA256_DIGEST_LENGTH);
+        }
+        
+        std::vector<uint8_t> SHA256::stringToBytesHash(const std::string &input) {
```
teams2ua: Is this a really string consisting of characters and numbers, or this just a container for binary data? If second, then this function will have problem with \0

PR #197, opened 2019-04-03
 Fix PreferencesBackend password change (empty new passwords now disable encryption).

phaazon:
Depends on #194.


phaazon:
# Once #194 is merged, rebase this one and it should be mergeable.

PR #198, opened 2019-04-05
Implement encryption disabling for SQLite3. 

phaazon:


phaazon:
@KhalilBellakrid what do you mean change? Also, I told you on slack to review it per-commit; there’s only two (don’t read the middle one, because it’s the one changing the line endings).

KhalilBellakrid:
@phaazon alright 👍

PR #200, opened 2019-04-05
Soci C++17-ready

teams2ua:


teams2ua:
> This is nice.
> 
> However, I fear the technical debt here, because we are patching a dependency of ours. Isn’t there a way to push that PR to [soci](https://github.com/SOCI/soci) instead, and remove that soci cloned in order to add it as a dependency in our git modules?

I am doubt they will accept it, https://github.com/SOCI/soci/issues/688 looks like they give up on C++11/14 support

phaazon:
No if the CI is okay that should be enough. I think we should fix that CI problem, it’s bothering. :'(

pollastri-pierre:
Yep, I'm more up to get rid of SOCI. It will also be the time to determine if we keep SQL database in the future.

PR #201, opened 2019-04-05
Update Android NDK

teams2ua:
To not use libncurses during building
and to apply recent bugfixes.

### DIFF
--- a/.circleci/build_lib.sh
+++ b/.circleci/build_lib.sh
@@ -60,16 +60,16 @@ function command_ios {
 
 function command_android {
   echo "Set Android NDK variable"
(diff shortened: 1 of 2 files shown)


teams2ua:
> I’m mostly okay with the changes but I think we should have the all the CI testing that part (we have that weird GitHub CI bug again?!)

Yeap, looks like. https://circleci.com/gh/teams2ua/lib-ledger-core/tree/new_android_ndk Do you want me to repush with fake change?

phaazon:
No if the CI is okay that should be enough. I think we should fix that CI problem, it’s bothering. :'(

PR #202, opened 2019-04-05
Remove boost package installation

teams2ua:


phaazon:
IIRC @pollastri-pierre might still need boost.

teams2ua:
> IIRC @pollastri-pierre might still need boost.

We have header-only boost in test/libs/boost.
I remove package-installed boost  to avoid ambiguity, once we will have a need of builded libs from boost, but we will need to work out some common approach.

KhalilBellakrid:
First, why are we doing this ? 
Second , if we really need this it's not ideal to have a `lib` depending on `test/lib` ...

teams2ua:
> First, why are we doing this ?
> Second , if we really need this it's not ideal to have a `lib` depending on `test/lib` ...

I remove this in order to not have ambiguity (+ we don't need this), some libraries may do find_boost() in CMakeLists.txt so they will use the boost installed as a package, but other code will use directly included headers.
Yes, I think should move this to lib, can be done in separate review.

pollastri-pierre:
👍 I'll do a PR if I need to add boost to our dependencies

PR #203, opened 2019-04-05
Switch to C++17

teams2ua:


core/src/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ if(IS_IOS GREATER_EQUAL 0)
             ${HEADERS_FILES})
     target_link_libraries(ledger-core PUBLIC ledger-core-interface)
     set(CMAKE_SHARED_LINKER_FLAGS "-Wall")
+    set(CMAKE_CXX_FLAGS "-fno-aligned-allocation")
```
phaazon: This is not related to C++17. Why are you doing this?

teams2ua: C++17 adds aligned allocations, but ARM doesn't support it, so to make the iOS build works I need to do this, in case some library will use this feature (I am looking at you stlab)

phaazon: Roger that. Can you then make that setting depends on ARM architecture, so that he doesn’t leak to other architectures as well?

teams2ua:
Postponed until we start include libc.so to linux distributions of ledger-live

PR #204, opened 2019-04-05
Add stlab [skip ci]

teams2ua:
Change use internal execution context inside Wallet and in all
eraseDataSince operation


core/src/wallet/common/AbstractWallet.cpp
```diff
@@ -174,16 +177,16 @@ namespace ledger {
             return _currency;
         }
 
-        Future<int32_t> AbstractWallet::getAccountCount() {
+        stlab::future<int32_t> AbstractWallet::getAccountCount() {
             auto self = shared_from_this();
-            return async<int32_t>([self]() -> int32_t {
+            return stlab::async(stlab::default_executor, [self]() -> int32_t {
```
pollastri-pierre: This will work on Ledger Live due to a side effect but none of the currency code are supposed to work on thread pools they all assume you're on a sequential queue

teams2ua:
Postponed until multithreding future comes

PR #207, opened 2019-04-15
Jdk 11

teams2ua:
Update djinni sbt + scala to work woth newer JDK
Update JDK to have a package on debian buster
Update debian to Buster to have a newer gcc
Update gcc to have C++17

### DIFF
--- a/.circleci/build_lib.sh
+++ b/.circleci/build_lib.sh
@@ -101,7 +101,7 @@ fi
 
 }
 
-export JAVA_HOME="/usr/lib/jvm/java-8-openjdk"
+export JAVA_HOME="/usr/lib/jvm/java-11-openjdk-amd64/"
 export POLLY_ROOT=`pwd`/toolchains/polly
 ###
 # Clean
--- a/core/src/crypto/AESCipher.cpp
+++ b/core/src/crypto/AESCipher.cpp
@@ -32,6 +32,7 @@
 #include "AESCipher.hpp"
 #include "PBKDF2.hpp"
 #include "AES256.hpp"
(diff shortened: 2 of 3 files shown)


.circleci/build_lib.sh
```diff
@@ -101,7 +101,7 @@ fi
 
 }
 
-export JAVA_HOME="/usr/lib/jvm/java-8-openjdk"
+export JAVA_HOME="/usr/lib/jvm/java-11-openjdk-amd64"
```
phaazon: That `amd64` brings concerns. Why are we now platform-dependent?

core/src/crypto/AESCipher.cpp
```diff
@@ -57,7 +57,6 @@ namespace ledger {
 
                 // Create an IVt
                 auto IV = _rng->getRandomBytes(AES256::BLOCK_SIZE);
-                assert(IV.size() == AES256::BLOCK_SIZE);
```
phaazon: Nope. You can’t do that. 🙅‍♂️ That assert is needed.

teams2ua: ok, put it back

.circleci/build_lib.sh
teams2ua: Don't know what to do here, this is the path the package (apt-get install openjdk-11-jdk) is installed... What would you recommend ?

core/src/crypto/AESCipher.cpp
phaazon: 🔞

PR #210, opened 2019-04-16
Add getBlockHeight method on erc20 operations

KhalilBellakrid:
Related issue : https://ledgerhq.atlassian.net/browse/LLC-210

### DIFF
--- a/core/idl/wallet/ethereum/erc20.djinni
+++ b/core/idl/wallet/ethereum/erc20.djinni
@@ -58,4 +58,7 @@ ERC20LikeOperation = interface +c {
     getOperationType(): OperationType;
     # Get opration status : pending or confirmed.
     getStatus(): i32;
+    # Get block height on which operation was included.
+    # @return Optional 64-bit integer, height of block in which operation was validated
+    getBlockHeight(): optional<i64>;
 }
--- a/core/src/wallet/ethereum/ERC20/ERC20LikeOperation.h
+++ b/core/src/wallet/ethereum/ERC20/ERC20LikeOperation.h
@@ -47,7 +47,7 @@ namespace ledger {
                                const Operation &operation,
                                const api::Currency &currency);
 
-            ERC20LikeOperation() {};
+            ERC20LikeOperation() : _blockHeight(0){};
             
             std::string getHash() override;
             std::shared_ptr<api::BigInt> getNonce() override;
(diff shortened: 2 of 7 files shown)


core/idl/wallet/ethereum/erc20.djinni
```diff
@@ -58,4 +58,7 @@ ERC20LikeOperation = interface +c {
     getOperationType(): OperationType;
     # Get opration status : pending or confirmed.
     getStatus(): i32;
+    # Get block height on which operation was included.
+    # @return Optional 64-bit integer, height of block in which operation was validated
+    getBlockHeight(): optional<i64>;
```
phaazon: Hm. Is it really needed to use an optional here? Do we care about a blockchain with no block vs. a blockchaini with only a single genesis block?

Also, djinni doesn’t have unsigned integers?!

phaazon: If not, I know what I have to do.

core/src/wallet/ethereum/ERC20/ERC20LikeOperation.h
```diff
@@ -47,7 +47,9 @@ namespace ledger {
                                const Operation &operation,
                                const api::Currency &currency);
 
-            ERC20LikeOperation() {};
+            ERC20LikeOperation() {
+                _blockHeight = 0;
```
phaazon: May you move that into the `ERC20LikeOperation` ctor’s initialization list? Tom Hanks!

core/idl/wallet/ethereum/erc20.djinni
KhalilBellakrid: This is to allow clients (Live or Vault) to differentiate between them because they have a pending status, otherwise it will be problematic for them 
And yes no unsigned integers on djinni ... SFOL

core/src/wallet/ethereum/ERC20/ERC20LikeOperation.h
```diff
@@ -47,7 +47,7 @@ namespace ledger {
                                const Operation &operation,
                                const api::Currency &currency);
 
-            ERC20LikeOperation() {};
+            ERC20LikeOperation() : _blockHeight(0){};
```
phaazon: The semicolon is not needed here, but I’ll approve as it’s not harmful.

PR #213, opened 2019-04-18
Fix encryption

KhalilBellakrid:
changing password from empty to non-empty password was crashing the app ...

### DIFF
--- a/core/src/wallet/pool/WalletPool.cpp
+++ b/core/src/wallet/pool/WalletPool.cpp
@@ -78,10 +78,11 @@ namespace ledger {
                 _pathResolver
             );
 
+            _rng = rng;
             // Encrypt the preferences, if needed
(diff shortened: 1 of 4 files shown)


core/src/wallet/pool/WalletPool.cpp
```diff
@@ -78,10 +78,11 @@ namespace ledger {
                 _pathResolver
             );
 
+            _rng = rng;
             // Encrypt the preferences, if needed
             if (!_password.empty()) {
-                _externalPreferencesBackend->setEncryption(rng, _password);
-                _internalPreferencesBackend->setEncryption(rng, _password);
+                _externalPreferencesBackend->setEncryption(_rng, _password);
```
phaazon: 😱 How did that change anything? Both points to the same object, right?!

PR #214, opened 2019-04-18
Handle multiple ERC20 transfer events in same ETH transaction

KhalilBellakrid:
Support multiple transfer events under same ETH transaction, this might be the case with DEXes


core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -145,7 +145,6 @@ namespace ledger {
                     return balance + value;
 
                 case api::OperationType::SEND:
-                default:
```
phaazon: C++ and its lack of exhaustive patterns… :(

PR #215, opened 2019-04-23
Fix Ethereum: ERC20 balance

KhalilBellakrid:
Handle transactions not belonging to ETH account that creates transfer events related to current synchronizing ETH account

### DIFF
--- a/core/idl/wallet/common/wallet.djinni
+++ b/core/idl/wallet/common/wallet.djinni
@@ -43,6 +43,7 @@ DerivationPath = interface +c {
 OperationType = enum {
     send;
     receive;
+    none;
 }
 
 # Level of trust for transactions.
--- a/core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
+++ b/core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
@@ -119,6 +119,8 @@ namespace ledger {
                         case api::OperationType::SEND:
                             sum = sum - value;
                             break;
+                        default:
+                            break;
                     }
                 }
             };
@@ -146,6 +148,8 @@ namespace ledger {
 
                 case api::OperationType::SEND:
                     return balance - value;
+                default:
+                    return balance;
             }
         }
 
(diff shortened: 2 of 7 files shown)


core/idl/wallet/common/wallet.djinni
```diff
@@ -43,6 +43,7 @@ DerivationPath = interface +c {
 OperationType = enum {
     send;
     receive;
+    none;
```
phaazon: What does it mean for an operation to be of no type?

core/src/wallet/ethereum/ERC20/ERC20LikeAccount.cpp
```diff
@@ -146,6 +148,8 @@ namespace ledger {
 
                 case api::OperationType::SEND:
                     return balance - value;
+                default:
```
phaazon: I’m not even sure we use that function anymore?

core/idl/wallet/common/wallet.djinni
teams2ua: Why we don't want to make this kind of operation as receive?

KhalilBellakrid: As explained in comments, we can have transactions made by someone else which leads to side effects e.g. changing our balance of erc20 token.
It's not a receive because it's not a value transfer toward our address, plus later we might have other side effects that affect account's address without any notion of transfer (e.g. internal transactions in general)

phaazon: Hm, I don’t get that _made by someone else_ part. If it’s a transfer, can we not make that an `OperationType::Transfer` type?

KhalilBellakrid: Basically I can interact with a smart contract (so transaction from `me` to `smartContract`) and this smart contract is sending events or making side effects on your account without being a transfer any kind
At the end from our clients side most important is to know if it's a receive, send or `else` they don't care about a specific type (it's mostly important for tx history, balance history)

PR #216, opened 2019-04-26
Support exotic derivation paths

KhalilBellakrid:
Live is supporting some weird derivation paths as: `44'/60'/14'/5'/16`, `44'/60'/0'/0' ...`

### DIFF
--- a/core/src/utils/DerivationScheme.cpp
+++ b/core/src/utils/DerivationScheme.cpp
@@ -77,6 +77,23 @@ namespace ledger {
                 }
                 _scheme.push_back(node);
             }
+
+            // Account & node level both should be set to be used by keychain.
+            // Account level always comes before node level
+            auto itAccountLevel = std::find_if(_scheme.begin(), _scheme.end(), [](const DerivationSchemeNode &node) {return node.level == DerivationSchemeLevel::ACCOUNT_INDEX;});
+            auto itNodeLevel = std::find_if(itAccountLevel, _scheme.end(), [](const DerivationSchemeNode &node) {return node.level == DerivationSchemeLevel::NODE;});
+            if (itNodeLevel == _scheme.end() && _scheme.size() <= 5) {
+                static std::vector<DerivationSchemeLevel> schemeLevel {DerivationSchemeLevel::UNDEFINED,
+                                                                       DerivationSchemeLevel::COIN_TYPE,
+                                                                       DerivationSchemeLevel::ACCOUNT_INDEX,
+                                                                       DerivationSchemeLevel::NODE,
+                                                                       DerivationSchemeLevel::ADDRESS_INDEX};
+                for (size_t i = 0; i < _scheme.size(); i++ ) {
+                    if (_scheme[i].level == DerivationSchemeLevel::UNDEFINED) {
+                        _scheme[i].level = schemeLevel[i];
+                    }
+                }
+            }
         }
 
         DerivationScheme::DerivationScheme(const std::vector<DerivationSchemeNode> &nodes) {
--- a/core/src/wallet/ethereum/keychains/EthereumLikeKeychain.cpp
+++ b/core/src/wallet/ethereum/keychains/EthereumLikeKeychain.cpp
@@ -161,25 +161,24 @@ namespace ledger {
             if (_address.empty()) {
 
                 _localPath = getDerivationScheme()
-                        .setAccountIndex(getAccountIndex())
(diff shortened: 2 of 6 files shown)


core/src/utils/DerivationScheme.cpp
```diff
@@ -77,6 +77,23 @@ namespace ledger {
...
+            auto itAccountLevel = std::find_if(_scheme.begin(), _scheme.end(), [](const DerivationSchemeNode &node) {return node.level == DerivationSchemeLevel::ACCOUNT_INDEX;});
+            auto itNodeLevel = std::find_if(itAccountLevel, _scheme.end(), [](const DerivationSchemeNode &node) {return node.level == DerivationSchemeLevel::NODE;});
+            if (itNodeLevel == _scheme.end() && _scheme.size() <= 5) {
+                std::vector<DerivationSchemeLevel> schemeLevel {DerivationSchemeLevel::UNDEFINED,
+                                                                DerivationSchemeLevel::COIN_TYPE,
+                                                                DerivationSchemeLevel::ACCOUNT_INDEX,
+                                                                DerivationSchemeLevel::NODE,
+                                                                DerivationSchemeLevel::ADDRESS_INDEX};
+                for (size_t i = 0; i < _scheme.size(); i++ ) {
+                    if (_scheme[i].level == DerivationSchemeLevel::UNDEFINED) {
```
teams2ua: Why do we compare with `DerivationSchemeLevel::UNDEFINED` here, shouldn't we just overwrite it alltogether?
It can create to some strange scheme with several nodes with the same type, for ex.
`{ACCOUN_INDEX, UNDEFINED, UNDEFINED}` will be transformed to `{ACCOUN_INDEX, COIN_TYPE, ACCOUNT_INDEX}`

core/src/utils/DerivationScheme.cpp
```diff
@@ -77,6 +77,23 @@ namespace ledger {
                 }
                 _scheme.push_back(node);
             }
+
+            // Account & node level both should be set to be used by keychain.
+            // Account level always comes before node level
+            auto itAccountLevel = std::find_if(_scheme.begin(), _scheme.end(), [](const DerivationSchemeNode &node) {return node.level == DerivationSchemeLevel::ACCOUNT_INDEX;});
+            auto itNodeLevel = std::find_if(itAccountLevel, _scheme.end(), [](const DerivationSchemeNode &node) {return node.level == DerivationSchemeLevel::NODE;});
+            if (itNodeLevel == _scheme.end() && _scheme.size() <= 5) {
+                std::vector<DerivationSchemeLevel> schemeLevel {DerivationSchemeLevel::UNDEFINED,
```
teams2ua: This can be make static, to avoid allocation each call.

core/src/utils/DerivationScheme.cpp
KhalilBellakrid: So yeah I thought about this kind of weird cases, but they don't make any sense, a "valid" derivation scheme should always respect a pattern of type : `<purpose>/<coin_type>/<account_index>/<node>/<account_index>` even for exotic ones you always have at least `<purpose>/<coin_type>/<account_index>`, but if we want to support some `non-sense` derivation path then I agree we should change condition or how do we correct the derivation scheme ! 
WDYT @pollastri-pierre ?

teams2ua: @KhalilBellakrid Can you please briefly describe what is the behaviour and why we have this algo in the first place. The questions that is not clear to me are - do we try to correct the scheme or just check that it is correct? How do we know that second node in `44'/60'/0'/0'` is actually account and not something else? Is this just a one case that was historicaly there and we want to support it, or this is general approach for weird derivations?

KhalilBellakrid: Yes of course, so the context is the Live's one, they have to have to support some derivation schemes that were chosen by some other wallets (MEW, Ledger Legacy ...) and main ones were: `44'/60'/14'/5'/16` and `44'/60'/0'/0'` in both cases Derivation Scheme levels are unknown, but I wanted to make it enough generic for cases where some nodes have a placeholder instead of fix value (e.g. <coin_type> instead of 60 ...) and as I said in my (code) comments, from keychain side most important part is to have both `account_index` and `node` levels set ! 
Again, if user choose a completely insane derivation scheme even if we support it on that part of code it won't really work elsewhere ...

gre:
can't wait to have this in so i can confirm all the cases works ;)

PR #217, opened 2019-04-29
Add support for destination tags for XRP

KhalilBellakrid:
This is a required feature for Live ! 
For more infos about destination tag: 
https://developers.ripple.com/rippleapi-reference.html#payment
https://support.kraken.com/hc/en-us/articles/360000184443-What-is-the-Destination-Tag-for-XRP-Deposits-
Corresponding task: https://ledgerhq.atlassian.net/browse/LLC-222
Note: we don't retrieve/parse the destination tag from received transactions in purpose, because it is useless/not used.

### DIFF
--- a/core/idl/wallet/ripple/ripple_like_wallet.djinni
+++ b/core/idl/wallet/ripple/ripple_like_wallet.djinni
@@ -38,6 +38,8 @@ RippleLikeTransaction = interface +c {
     getMemos(): list<RippleLikeMemo>;
     # Add a memo to a transaction.
     addMemo(memo: RippleLikeMemo);
+    # An arbitrary unsigned 32-bit integer that identifies a reason for payment or a non-Ripple account
+    getDestinationTag(): optional<i64>;
 }
 
 #Class representing a Ripple Operation
@@ -80,6 +82,9 @@ RippleLikeTransactionBuilder = interface +c {
     # Add a memo.
     # @return A reference on the same builder in order to chain calls.
     addMemo(memo: RippleLikeMemo): RippleLikeTransactionBuilder;
+
+    # An arbitrary unsigned 32-bit integer that identifies a reason for payment or a non-Ripple account
+    setDestinationTag(tag: i64): RippleLikeTransactionBuilder;
 
     # Build a transaction from the given builder parameters.
     build(callback: Callback<RippleLikeTransaction>);
--- a/core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
+++ b/core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
@@ -194,6 +194,13 @@ namespace ledger {
             auto sequence = _sequence->toString(16);
             writer.writeBeValue<uint32_t>(static_cast<const uint32_t>(_sequence->intValue()));
 
+            if (_destinationTag.hasValue()) {
+                //1 byte Destination tag:   Type Code = 2, Field Code = 14
+                writer.writeByte(0x2E);
+                //4 bytes Destination tag
+                writer.writeBeValue<uint32_t>(static_cast<const uint32_t>(_destinationTag.getValue()));
+            }
+
             //2 bytes LastLedgerSequence Field ID:   Type Code = 2, Field Code = 27
             writer.writeByteArray({0x20, 0x1B});
             //LastLedgerSequence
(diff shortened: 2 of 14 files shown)


core/idl/wallet/ripple/ripple_like_wallet.djinni
```diff
@@ -38,6 +38,8 @@ RippleLikeTransaction = interface +c {
     getMemos(): list<RippleLikeMemo>;
     # Add a memo to a transaction.
     addMemo(memo: RippleLikeMemo);
+    # An arbitrary unsigned 32-bit integer that identifies a reason for payment or a non-Ripple account
```
phaazon: s/or/for

phaazon: Also, you only need 32 bit.

core/idl/wallet/ripple/ripple_like_wallet.djinni
```diff
@@ -81,6 +83,9 @@ RippleLikeTransactionBuilder = interface +c {
     # @return A reference on the same builder in order to chain calls.
     addMemo(memo: RippleLikeMemo): RippleLikeTransactionBuilder;
 
+    # An arbitrary unsigned 32-bit integer that identifies a reason for payment or a non-Ripple account
+    setDestinationTag(tag: i64): RippleLikeTransactionBuilder;
```
phaazon: `i32` should be enough. 🙂

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -351,12 +358,21 @@ namespace ledger {
...
+
         std::vector<api::RippleLikeMemo> RippleLikeTransactionApi::getMemos() {
             return _memos;
         }
 
         void RippleLikeTransactionApi::addMemo(api::RippleLikeMemo const& memo) {
             _memos.push_back(memo);
         }
+
+        std::experimental::optional<int64_t> RippleLikeTransactionApi::getDestinationTag() {
```
phaazon: Can you move that function nearby `setDestinationTag`, please?

core/src/wallet/ripple/transaction_builders/RippleLikeTransactionBuilder.cpp
```diff
@@ -158,6 +162,14 @@ namespace ledger {
             auto bigIntSequence = BigInt::fromHex(hex::toString(sequence));
             tx->setSequence(bigIntSequence);
 
+            auto destinationTagCode = reader.peek();
```
phaazon: I think a better way to do this would be to `reader.read` the bytes and `switch` or have branching with `if` / `else if`. The problem here is that if `destinationTagCode ≠ 0x2E`, you’re going to throw the peek away for nothing at line 174.

core/idl/wallet/ripple/ripple_like_wallet.djinni
KhalilBellakrid: I took the line from Ripple's doc :trollface: 
For 32-bit remark, we don't have unsigned 32-bit in Djinni so ...

core/idl/wallet/ripple/ripple_like_wallet.djinni
KhalilBellakrid: Same as before

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
KhalilBellakrid: Of course !

core/src/wallet/ripple/transaction_builders/RippleLikeTransactionBuilder.cpp
KhalilBellakrid: Agreed on that, I complied with has been done before but I can change that !

KhalilBellakrid: Actually I just checked the next field and it has 2 bytes, so we might lose in readability, here is the result of it : 
```
        auto nextCode = reader.readNextByte();
            switch (nextCode) {
                case 0x2E:
                    //1 byte Destination tag:   Type Code = 2, Field Code = 14
                    auto bigIntTag = BigInt::fromHex(hex::toString(reader.read(4)));
                    tx->setDestinationTag(static_cast<const uint32_t>(bigIntTag.toUint64()));
                    break;
                default:
                    break;
            }

            //2 bytes LastLedgerSequence Field ID:   Type Code = 2, Field Code = 27
            reader.readNextByte(); 
            //LastLedgerSequence
            auto ledgerSequence = reader.read(4);
```
So we lose the 2 bytes thing of LastLedgerSequence (refer to comment right above about field & type code)
@phaazon what do you think ?

PR #219, opened 2019-05-04
Fix encryption of plain database

KhalilBellakrid:
It seems that opening an existing DB with `sqlite_key_v2` does not rewrite it with encryption ... 
So we are forced to go through `sqlcipher_export` ...
Now test `AccountCreationTest.ChangePassword` is producing an enrypted database !
Related issue: https://ledgerhq.atlassian.net/browse/LLC-232

### DIFF
--- a/core/lib/soci_sqlite3/session.cpp
+++ b/core/lib/soci_sqlite3/session.cpp
@@ -25,30 +25,85 @@ using namespace sqlite_api;
 namespace // anonymous
 {
 
-// helper function for hardcoded queries
-void execute_hardcoded(sqlite_api::sqlite3* conn, char const* const query, char const* const errMsg)
-{
-    char *zErrMsg = 0;
-    int const res = sqlite3_exec(conn, query, 0, 0, &zErrMsg);
-    if (res != SQLITE_OK)
+    // helper function for hardcoded queries
+    void execute_hardcoded(sqlite_api::sqlite3* conn, char const* const query, char const* const errMsg)
     {
-        std::ostringstream ss;
-        ss << errMsg << " " << zErrMsg;
-        sqlite3_free(zErrMsg);
-        throw sqlite3_soci_error(ss.str(), res);
+        char *zErrMsg = 0;
+        int const res = sqlite3_exec(conn, query, 0, 0, &zErrMsg);
+        if (res != SQLITE_OK)
+        {
+            std::ostringstream ss;
+            ss << errMsg << " " << zErrMsg;
+            sqlite3_free(zErrMsg);
+            throw sqlite3_soci_error(ss.str(), res);
+        }
     }
-}
 
-void check_sqlite_err(sqlite_api::sqlite3* conn, int res, char const* const errMsg)
-{
-    if (SQLITE_OK != res)
+    void check_sqlite_err(sqlite_api::sqlite3* conn, int res, char const* const errMsg)
     {
-        const char *zErrMsg = sqlite3_errmsg(conn);
-        std::ostringstream ss;
-        ss << errMsg << zErrMsg;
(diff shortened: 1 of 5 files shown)


core/src/database/SQLite3Backend.cpp
```diff
@@ -69,13 +68,13 @@ namespace ledger {
                 throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Database should be initiated before changing password.");
             }
 
-            std::string db_params;
+            auto db_params = fmt::format("dbname=\"{}\" ", _dbResolvedPath) + fmt::format("key=\"{}\" ", oldPassword);
             if (!newPassword.empty()) {
                 // change password if a new password is provided
-                db_params = fmt::format("dbname=\"{}\" ", _dbResolvedPath) + fmt::format("key=\"{}\" ", oldPassword) + fmt::format("new_key=\"{}\" ", newPassword);
+                db_params += fmt::format("new_key=\"{}\" ", newPassword);
```
phaazon: Just a small comment: your change will make that string to allocate twice while the previous code allocated only once. While doing that?

core/test/integration/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ add_test (NAME ledger-core-integration-AccountBlockchainObservationTests.AutoRec
 add_test (NAME ledger-core-integration-AccountBlockchainObservationTests.EmitNewBlock COMMAND ledger-core-integration-tests --gtest_filter="AccountBlockchainObservationTests.EmitNewBlock")
 add_test (NAME ledger-core-integration-AccountCreationTest.CreateBitcoinAccountWithInfo COMMAND ledger-core-integration-tests --gtest_filter="AccountCreationTest.CreateBitcoinAccountWithInfo")
 add_test (NAME ledger-core-integration-AccountCreationTest.CreateBitcoinAccountWithInfoOnExistingWallet COMMAND ledger-core-integration-tests --gtest_filter="AccountCreationTest.CreateBitcoinAccountWithInfoOnExistingWallet")
+add_test (NAME ledger-core-integration-AccountCreationTest.ChangePassword COMMAND ledger-core-integration-tests --gtest_filter="AccountCreationTest.ChangePassword")
```
phaazon: Dammit. :D

core/lib/soci_sqlite3/session.cpp
```diff
@@ -125,17 +180,28 @@ sqlite3_session_backend::sqlite3_session_backend(
     post_connection(dbname, connection_flags, timeout, synchronous);
 
     //Set password
-    if (!passKey.empty())
+    if (!passKey.empty() || !newPassKey.empty())
     {
-        res = sqlite3_key_v2(conn_, dbname.c_str(), passKey.c_str(), strlen(passKey.c_str()));
-        check_sqlite_err(conn_, res, "Failed to encrypt database. ");
+        if (!passKey.empty()){
```
phaazon: I need to read all those tests and have a look at the generated conditional graph. That seems like a lot of redundant testing.

core/src/database/SQLite3Backend.cpp
KhalilBellakrid: +1, but here I did this to clearly see/be sure that we use same (base) params plus their differences, to me it's an "optim" that we can make once the feature is bug free

core/lib/soci_sqlite3/session.cpp
KhalilBellakrid: Yes please do !
Again here it's more about easily and clearly see the flow, let's make it green then let's make it gold ! :trollface:

phaazon:
LGTM.

PR #220, opened 2019-05-19
Tezos integration

KhalilBellakrid:
### Keychain:
- Tezos key derivation is based on Edward's curve (https://github.com/satoshilabs/slips/blob/master/slip-0010.md), 
so we don't have public derivations and since Libcore does not support private ones, the only supported derivation scheme, right now, is `44'/<coin_type>'/<account_index>'/<node_index>'`,
- As a result, "derived" address from keychain is directly a hash of the public key, 
- Used hashing algorithm is `Blake2b`,
- We support construction of accounts from both `Secp256k1` (`Base58`) extended public keys (prefixed with `xpub`) or from `Ed25519` (`Base58`) extended public keys (prefixed with `edpk`).
### Wallet logic: 
- Originated accounts are considered as sub-accounts of implicit accounts that originated them,
### Explorer:
- The explorer used by the default (for the moment) is Tzscan (used endpoint: https://api6.tzscan.io/v3),
- **TODO**: we still miss some endpoints or could improve some others: 
	- Missing endpoints to : 
		- Get storage limit,
		- Get gas limit, 
		- Get counter,
		- Get and push raw transactions,
	- Improvements:
		- Improve call to get an address's transactions: can have an exposed call to get all types of operations at once, for the moment we have to specify type as parameter for the `GET` request,
		- TBD: have a call to query operations from list of addresses (could be useful).
### Synchronizer:
- Synchronization of originated accounts comes right after the originator's account is synced, 
### Observer:
- **TODO**: Implement WS from backend side,
### Transactions:
- For the moment we only support `Transaction`, `Reveal`, `Origination` and `Delegation` operations, those are the ones relevant for us,
 - Parsing and serialization of transaction is based on Zarith encoder/decoder (for more details refer to: http://tezos.gitlab.io/master/api/p2p.html#n-t 
 implementation is located here : `core/src/bytes/zarith`),
 - In Tezos, it is possible to have multiple operations in one transaction (e.g. `Reveal` + `Origination` or `Reveal` + `Transaction` ...), 
 since each operation has its own properties (fees, gas limit, storage limit ...) that user should set, so we probably won't handle multi-operation transactions.
 This will be discussed with Libcore's users (e.g. Live team) to see if it is relevant to support multi-operation transactions.
 - **TODO**: 
 	- Handle parsing/serialization of signed transactions,
 	- Handle remaining types of operations,
 	- TBD: handle multi-operation transaction,
### Delegation:	
- To delegate simply build a `Delegation` transaction with address to which delegate as receiver.  
### Parameters:
- BIP44 Coin type : `1729`,
- Used address prefixes: 
	- Implicit (prefix `tz1`): `{0x06, 0xA1, 0x9F}`,
	- Originated (prefix `KT1`): `{0x02, 0x5A, 0x79}`,
- Default explorer's endpoint: https://api6.tzscan.io/v3
### Tests:
- Added tests:
	- Address derivation: `core/test/tezos`,
	- Zarith encoder: `core/test/bytes/zarith.cpp`,
	- Account creation and synchronization: `core/test/integration/synchronization/tezos_synchronization.cpp`,
	- Transaction construction: `core/test/integration/transactions/tezos_transaction_tests.cpp`,
- TODO: 
	- Still need to test:
		- Broadcast of signed transactions,
		- Test E2E, 
### References:
- https://gitlab.com/tezos
- http://tezos.gitlab.io/master/index.html
- https://github.com/obsidiansystems/ledger-app-tezos
- https://gitlab.com/tezos/tezos/merge_requests/994
- https://tzscan.io/api


CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+#### References:
+- https://gitlab.com/tezos
+- http://tezos.gitlab.io/master/index.html
+- https://github.com/obsidiansystems/ledger-app-tezos
+- https://gitlab.com/tezos/tezos/merge_requests/994
+- https://tzscan.io/api
+
+## 2.7.0 
+
+> 2019/05/19
```
phaazon: I don’t see the tag in the git repository. Is that normal?

phaazon: Nevermind, it appeared.

core/src/crypto/BLAKE.cpp
```diff
@@ -54,6 +54,15 @@ namespace ledger {
             return hash;
         }
 
+        std::vector<uint8_t> BLAKE::blake2b(const std::vector<uint8_t>& data, size_t outLength) {
```
phaazon: I suggest to change the interface here to make it less coupled on the container. The internal representation only needs a buffer, so imposing vector is a bit too strong. See my other comment on the use of this function.

core/src/crypto/DeterministicPublicKey.cpp
```diff
@@ -88,6 +89,15 @@ namespace ledger {
             return  std::vector<uint8_t>(keccak.end() - 20, keccak.end());
         }
 
+        std::vector<uint8_t> DeterministicPublicKey::getPublicKeyBlake2b(bool isED25519) const {
+            auto key = _key;
+            if (!isED25519) {
+                //Remove 0x02
+                key.erase(key.begin());
```
phaazon: This is a O(n) operation that could have been prevented if you made `BLAKE::blake2b` use a slice / iterator instead of using a `std::vector`.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
-## 2.7.0 (on-going)
-
-> ????/??/??
+## 2.8.0 (on-going)
```
phaazon: Please respect the style of the rest of the document by using whitespaces between lines starting with a sharp (`#`) and indenting the dashes (`-`) for bullet lists, even for the first level. Mixing styles in such a document will lead to confusion and it’s better to keep things clean. 🙂

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+
+### Tezos Integration:
+#### Keychain:
+- Tezos key derivation is based on Edward's curve (https://github.com/satoshilabs/slips/blob/master/slip-0010.md),
```
phaazon:
```
[Edward's curve](https://github.com/satoshilabs/slips/blob/master/slip-0010.md)
```

Better to use a hyperlink here.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
-## 2.7.0 (on-going)
-
-> ????/??/??
+## 2.8.0 (on-going)
+
+### Tezos Integration:
+#### Keychain:
+- Tezos key derivation is based on Edward's curve (https://github.com/satoshilabs/slips/blob/master/slip-0010.md), 
+we don't have public derivations and since Libcore does not support private ones, only derivation scheme which is supported right now is `44'/<coin_type>'/<account_index>'/<node_index>'`,
```
phaazon: I don’t get the structure of this sentence, maybe you want to rephrase it a bit?

CHANGELOG.md
phaazon: Same comment for the endings of the sentences in bullet lists: the whole document uses `periods`, not `commas`; please switch to periods.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+- Hence, resulting "derived" address from a keychain is directly a hash of the public key, 
+- Used hashing algorithm is `Blake2b`,
```
phaazon: Please provide a link to what this algorithm is (probably the documentation / source code you used to implement it in our lib).

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+- We support construction of accounts from both `Secp256k1` (`Base58`) extended public keys (prefixed with `xpub`) or from `Ed25519` (`Base58`) extended public keys (prefixed with `edpk`).
+#### Wallet logic: 
+- Originated accounts are considered as sub-accounts of implicit accounts that originated them,
+#### Explorer:
+- The explorer used by the default (for the moment) is Tzscan (used endpoint: https://api6.tzscan.io/v3),
```
phaazon: *used by default*, or *The default explorer*.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+- **TODO**: we still miss some endpoints or could improve some others:
```
phaazon: This has nothing to do in a changelog; this belongs to the issue tracker, or a roadmap. I think we should not pollute the changelog, that will be read by people who need to migrate their dependencies to a newer one, with things that are not done: it’s going to confuse them a lot and it will bring zero value to them.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+#### Observer:
+- **TODO**: Implement WS from backend side,
+#### Transactions:
+- For the moment we only support `Transaction`, `Reveal`, `Origination` and `Delegation` operations, those are the ones relevant for us,
+ - Parsing and serialization of transaction is based on Zarith encoder/decoder (for more details refer to: http://tezos.gitlab.io/master/api/p2p.html#n-t 
+ implementation is located here : `core/src/bytes/zarith`),
+ - In Tezos, it is possible to have multiple operations in one transaction (e.g. `Reveal` + `Origination` or `Reveal` + `Transaction` ...), 
+ since each operation has it's own properties (fees, gas limit, storage limit ...) that user should set, so we probably won't handle multi-operation transactions.
+ This will be discussed with Libcore's users (e.g. Live team) to see if it is relevant to support multi-operation transactions.
+ - **TODO**:
```
phaazon: Ditto -> issue tracker, please.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+- For the moment we only support `Transaction`, `Reveal`, `Origination` and `Delegation` operations, those are the ones relevant for us,
+ - Parsing and serialization of transaction is based on Zarith encoder/decoder (for more details refer to: http://tezos.gitlab.io/master/api/p2p.html#n-t 
+ implementation is located here : `core/src/bytes/zarith`),
+ - In Tezos, it is possible to have multiple operations in one transaction (e.g. `Reveal` + `Origination` or `Reveal` + `Transaction` ...), 
+ since each operation has it's own properties (fees, gas limit, storage limit ...) that user should set, so we probably won't handle multi-operation transactions.
+ This will be discussed with Libcore's users (e.g. Live team) to see if it is relevant to support multi-operation transactions.
+ - **TODO**: 
+ 	- Handle parsing/serialization of signed transactions,
+ 	- Handle remaining types of operations,
+ 	- TBD: handle multi-operation transaction,
```
phaazon: Issue tracker.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+	- Implicit (prefix `tz1`): `{0x06, 0xA1, 0x9F}`,
+	- Originated (prefix `KT1`): `{0x02, 0x5A, 0x79}`,
+- Default explorer's endpoint: https://api6.tzscan.io/v3
+#### Tests:
+- Added tests:
+	- Address derivation: `core/test/tezos`,
+	- Zarith encoder: `core/test/bytes/zarith.cpp`,
+	- Account creation and synchronization: `core/test/integration/synchronization/tezos_synchronization.cpp`,
+	- Transaction construction: `core/test/integration/transactions/tezos_transaction_tests.cpp`,
+- TODO:
```
phaazon: Issue tracker.

CHANGELOG.md
```diff
@@ -1,6 +1,68 @@
...
+- Added tests:
+	- Address derivation: `core/test/tezos`,
+	- Zarith encoder: `core/test/bytes/zarith.cpp`,
+	- Account creation and synchronization: `core/test/integration/synchronization/tezos_synchronization.cpp`,
+	- Transaction construction: `core/test/integration/transactions/tezos_transaction_tests.cpp`,
+- TODO: 
+	- Still need to test:
+		- Broadcast of signed transactions,
+		- Test E2E, 
+#### References:
```
phaazon: Please remove this paragraph. It has nothing to do in a changelog either. You can move it to the codebase, for instance, which is a place where people will want to get some references, but not in a changelog.

core/idl/tezos/addresses.djinni
```diff
@@ -0,0 +1,53 @@
...
+	getNetworkParameters(): TezosLikeNetworkParameters;
+	# Encodes to Base58.
+	# @return The Base58 encoding
+	toBase58(): string;
+}
+
+# The xPUB definition for Tezos.
+TezosLikeExtendedPublicKey = interface +c {
+    # Derive an address from an xPUB and a path.
+	derive(path: string): TezosLikeAddress;
```
phaazon: Incorrect indentation (till line 52).

core/idl/wallet/common/currency.djinni
```diff
@@ -43,6 +46,8 @@ Currency = record {
     bitcoinLikeNetworkParameters: optional<BitcoinLikeNetworkParameters>;
     # Optional EthereumLikeNetworkParameters, for more details refer to EthereumLikeNetworkParameters doc
     ethereumLikeNetworkParameters: optional<EthereumLikeNetworkParameters>;
-    #Optional EthereumLikeNetworkParameters, for more details refer to EthereumLikeNetworkParameters doc
+    #Optional RippleLikeNetworkParameters, for more details refer to RippleLikeNetworkParameters doc
```
phaazon: Nice catch.

core/lib/openssl/crypto/blake/blake2b.c
```diff
@@ -0,0 +1,347 @@
...
+ */
+
+#include <assert.h>
+#include <string.h>
+#include <openssl/crypto.h>
+
+#include "blake2b.h"
+
+static const uint64_t blake2b_IV[8] =
+        {
```
phaazon: May you reduce the indentation, please?

core/lib/openssl/crypto/blake/blake2b.c
```diff
@@ -0,0 +1,347 @@
...
+                0x6a09e667f3bcc908ULL, 0xbb67ae8584caa73bULL,
+                0x3c6ef372fe94f82bULL, 0xa54ff53a5f1d36f1ULL,
+                0x510e527fade682d1ULL, 0x9b05688c2b3e6c1fULL,
+                0x1f83d9abfb41bd6bULL, 0x5be0cd19137e2179ULL
+        };
+
+static const uint8_t blake2b_sigma[12][16] =
+        {
```
phaazon: Same. 😆

core/src/crypto/BLAKE.cpp
teams2ua: Iterators would be C++ way to do this. But it will differ from all other functions in this file that accept vector<uint8_t>, imo this should be separate commit that change vector<> to iterators, if we eventually want this.

core/src/crypto/DeterministicPublicKey.cpp
teams2ua: `auto key = _key;` is O(n) as well.

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,164 @@
...
+             * us know there is still something to read, else 0
+             * To sum up, if the resulting bytes are "chained" together (little endian) and we drop
+             * the most significant bit of each byte, we get the input ("chained") bytes
+             * e.g.
+             * Input:           ad d9                  10101101 11011001  (I)
+             * Output:          d9 db 02      11011001 11011011 00000010
+             * LE:              02 db d9      00000010 11011011 11011001
+             * Drop 1st bit:            (0)0000010 (1)1011011 (1)1011001  == (I)
+            */
+            static std::vector<uint8_t> zSerialize(const std::vector<uint8_t> &inputData) {
```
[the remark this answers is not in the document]
teams2ua: I came up with this code, that do the same, but is shorter.
This is mostly the question of taste.
```
// stream of 8-bit bytes to 7-bits bytes
template <typename Iterator, typename OutputIterator>
void encode(Iterator start, Iterator finish, OutputIterator out) {
    uint8_t rest = 0;
    while (start != finish) {
        for (int i = 0; (i < 7) && (start != finish); ++i, ++start) {
            uint8_t currentByte = *start;
            *out++ = 0x80 | (currentByte << i) | rest;
            rest = currentByte >> (7 - i);
        }
        *out++ = rest | (start != finish ? 0x80 : 0x00);
    }
}

std::vector<uint8_t> zSerializeNumber(const std::vector<uint8_t>& number) {
    std::vector<uint8_t> result;
    encode(rbegin(number), rend(number), std::back_inserter(result));
    return result;
}
```

core/src/crypto/BLAKE.cpp
phaazon: Problems with iterators is that they do not ensure contiguity (unless we use a `ContiguousIterator` concept?).

KhalilBellakrid: +1 @teams2ua , that's why the method has this signature 
@phaazon don't forget that this methods are used internally only with byte arrays

core/src/crypto/DeterministicPublicKey.cpp
KhalilBellakrid: Nice catch, this was my "blind" test phase but forgot to remove it, I was willing to have an offset on `BLAKE::blake2b` that would solve the issue !

core/idl/tezos/addresses.djinni
KhalilBellakrid: Good catch !

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,164 @@
...
+                // Reverse because result corresponds to Little Endian
+                std::vector<uint8_t> data{inputData};
```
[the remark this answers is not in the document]
KhalilBellakrid: Indeed, it was the first iteration of the algorithm !

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,164 @@
...
+
+                    // We shift current byte
+                    byte <<= offset;
+
+                    // Take shifted bits from previous byte
+                    byte |= id - delay > 0 ? data[id - delay - 1] >> (7 - (offset - 1)) : 0x00;
+
+                    // This boolean will let us know if shifting last byte of inputData
+                    // will generate a new byte (with remaining shifted bytes)
+                    bool isLast = data[id - delay] == data.back();
```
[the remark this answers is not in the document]
KhalilBellakrid: Niiice catch !

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,164 @@
...
+                std::vector<uint8_t> data(reader.readUntilEnd());
+                reader.reset();
+                reader.read(cursor);
+
+                std::vector<uint8_t> result;
+                size_t id = 0, delay = 0;
+                uint8_t offset = 0;
+                while (id + delay < data.size()) {
+                    auto byte = data[id + delay];
+                    auto isLast = byte == data.back();
```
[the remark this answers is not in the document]
KhalilBellakrid: Yes I agree !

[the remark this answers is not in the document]
teams2ua: You didn't fix this.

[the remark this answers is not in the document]
KhalilBellakrid: Alright, I see what happened here, I fixed it but in the wrong file ...  Navigation on my IDE was taking me to `cmake-build-debug/include/ledger/core/bytes/zarith/zarith.h` instead of `core/bytes/zarith/zarith.h` ... :trollface:

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,160 @@
...
+
+                    // Push to result modified byte
+                    result.push_back(byte);
+                    if (needAdditionalByte) {
+                        result.push_back(inputData[id + delay] >> (7 - offset));
+                    }
+
+                    // Update offset: if offset % 7 == 0 then following byte
+                    // is not shifted i.e. offset = 0
+                    offset = offset > 0 && !(offset % 7) ? 0 : offset + 1;
```
teams2ua: `offset = (offset + 1) % 7;` is more clear IMO

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,160 @@
...
+
+                    // If we shifted with [id/7] > 7 then we have a "delay"
+                    // meaning that we should read values from "delayed" byte
+                    delay = ((inputData.size() - 1) - id) / 7;
```
teams2ua: 1. Looks like there is a bug here:
id = 0;
delay  = 0;
inputData.size() = 8;
we just processed inputData[id + delay] -> inputData[0],
now delay = ((8 - 1) - 0) / 7 -> delay became 1, in the next line id became -1, so next iteration we will still proceed inputaData[0] so we processed this byte twice.
2. I think you don't need `delay` at all.

core/src/bytes/zarith/zarith.h
KhalilBellakrid: It's not the equivalent (e.g. offset = 7)

core/src/bytes/zarith/zarith.h
KhalilBellakrid: 1. Actually here the issue is that we should `break` after processing `id + delay == 0`, but regarding your remark, in case `inputData .size() > 8` e.g. `inputData .size() = 9`, for `(id , delay, offset) = (1, 0, 7)`, in the first iteration, `inputData[1]` will be shifted by 7, so basically in `result` you push a byte `1<7 most significant bits from inputData[2]>`, the second iteration, with `(id , delay, offset) = (0, 1, 0)`  will push `x<7 less significant bits from inputData[1]>` (`x` depends if most significant bit of `inputData[1]` is a 0 or 1 that's when we add an additional byte)
2. I think we need `delay` for cases when `inputData.size() > 8` 
We can discuss that on Friday or Monday it will be easier! 😉

core/src/bytes/zarith/zarith.h
teams2ua: Ah, I meant `offset = (offset +1) % 8;`

core/src/bytes/zarith/zarith.h
teams2ua: 1. Lets look at this example { 0xFF, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 };
the code will get output 81 80 80 80 80 80 80 80 **FF FF** 01,  that can't be right - too many new ones.
2. You only use  `inputData[id + delay]` and `inputData[id + delay + 1]`, that can be replaced by `inputData[idx]` and `inputData[idx + 1]` and a new way of changing `idx`.

KhalilBellakrid: So yeah this is an issue which is explained but my comment above, "Actually here the issue is that we should `break` after processing `id + delay == 0`" 
Btw, I added a test with test vector above ! 👍

teams2ua: Thanks for adding the test.
But the bug remains there, and it is not related to the last byte, the problem is that we proceed with  the same bytes twice when we increment `delay` and decrement `id`
Try this vector `{ 0x01, 0x00, 0xFF, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 }` it will again give the answer with two many "ones".
BTW you use `size_t` for `id`, `id` can became negative, so it start to overflow, there were no errors because when we add `delay` to overflowed `id` it overflows again and became normal number, but I found this dangerous anyway.

KhalilBellakrid: Oh yeah nice catch for the example ! 👍 
For the overflow thing, if you take into account the `break` condition that I added above and test on `id + delay` we have, it's not dangerous, we care about `id + delay` reason why, I agree, we should replace it by only `id` which is also simplifying the condition to increment `id` 
I think now it should be fine, WDYT ? 🤔 
Thanks for the review BTW ! 🙏

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,162 @@
...
+                    bool needAdditionalByte = isFirst && inputData[id] >> (7 - offset);
+
+                    // We set first bit to 1 to know that there is a coming byte
+                    if (!isFirst || needAdditionalByte) {
+                        byte |= 0x80;
+                    }
+
+                    // Push to result modified byte
+                    result.push_back(byte);
+                    if (needAdditionalByte) {
```
teams2ua: This will be true only once during the loop, why checking it every iteration? would be nice to move it outside the loop.

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,162 @@
...
+                    // Take shifted bits from previous byte
+                    byte |= id < inputData.size() - 1 ? inputData[id + 1] >> (7 - (offset - 1)) : 0x00;
+
+                    // This boolean will let us know if shifting last byte of inputData
+                    // will generate a new byte (with remaining shifted bytes)
+                    bool isFirst = id == 0;
+                    bool needAdditionalByte = isFirst && inputData[id] >> (7 - offset);
+
+                    // We set first bit to 1 to know that there is a coming byte
+                    if (!isFirst || needAdditionalByte) {
```
teams2ua: this will be false only once. Lets always do byte |= 0x80 and then outside the loop `result[result.size()-1] &= 0xEF;`

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,162 @@
...
+                        byte |= 0x80;
+                    }
+
+                    // Push to result modified byte
+                    result.push_back(byte);
+                    if (needAdditionalByte) {
+                        result.push_back(inputData[id] >> (7 - offset));
+                    }
+
+                    if (isFirst) {
```
teams2ua: This will be true only once, but planet will be heated each iteration.

core/src/bytes/zarith/zarith.h
teams2ua: For me it looks good now, just few comments that should speed up a little bit and increase readability.

core/src/bytes/zarith/zarith.h
```diff
@@ -0,0 +1,162 @@
...
+                        break;
+                    }
+
+                    // Update offset
+                    offset = (offset + 1) % 8;
+
+                    // Update index if not a period or first iteration
+                    if (offset > 0 || id == inputData.size() - 1) {
```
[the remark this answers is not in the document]
KhalilBellakrid: Not working for first iteration !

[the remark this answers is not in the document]
teams2ua: It will, because offset will be already 1.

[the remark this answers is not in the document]
KhalilBellakrid: True we do that above !

core/src/bytes/zarith/zarith.h
KhalilBellakrid: I'm not a huge fan of doing then undoing things ... but what can't we do for optims ?! :trollface: 
By the way, here it's `result[result.size()-1] &= 0x7F;`

core/src/wallet/tezos/TezosLikeAccount.h
```diff
@@ -0,0 +1,161 @@
...
+            };
+        };
+
+        class TezosLikeOriginatedAccount;
+        class TezosLikeAccount : public api::TezosLikeAccount, public AbstractAccount {
+        public:
+
+            static const int FLAG_TRANSACTION_IGNORED = 0x00;
+            static const int FLAG_NEW_TRANSACTION = 0x01;
+            static const int FLAG_TRANSACTION_UPDATED = 0x01 << 1;
```
phaazon: I think it’s okay to use `0x02` :D

core/src/wallet/tezos/TezosLikeAccount.h
```diff
@@ -0,0 +1,161 @@
...
+            void getEstimatedGasLimit(const std::string & address, const std::shared_ptr<api::BigIntCallback> & callback) override ;
+
+            void getStorage(const std::string & address, const std::shared_ptr<api::BigIntCallback> & callback) override;
+
+            std::vector<std::shared_ptr<api::TezosLikeOriginatedAccount>> getOriginatedAccounts() override;
+
+            void addOriginatedAccounts(soci::session &sql, const std::vector<TezosLikeOriginatedAccountDatabaseEntry> &originatedEntries);
+
+        private:
+            std::shared_ptr<TezosLikeAccount> getSelf();
```
phaazon: Please go to the new line to separate attributes from private methods.

core/src/wallet/tezos/TezosLikeWallet.cpp
```diff
@@ -0,0 +1,228 @@
...
+    namespace core {
+
+        const api::WalletType TezosLikeWallet::type = api::WalletType::ETHEREUM;
+
+        TezosLikeWallet::TezosLikeWallet(const std::string &name,
+                                         const std::shared_ptr<TezosLikeBlockchainExplorer> &explorer,
+                                         const std::shared_ptr<TezosLikeBlockchainObserver> &observer,
+                                         const std::shared_ptr<TezosLikeKeychainFactory> &keychainFactory,
+                                         const TezosLikeAccountSynchronizerFactory &synchronizer,
+                                         const std::shared_ptr<WalletPool> &pool, const api::Currency &network,
```
phaazon: Please break this line in two after the comma.

core/src/wallet/tezos/TezosLikeAccount.h
KhalilBellakrid: I wanted to group all same flags for different types of operations i.e.  `FLAG_NEW_TRANSACTION = 0x01` , `FLAG_NEW_DELEGATION = 0x02` ... then for updates same i.e. `FLAG_TRANSACTION_UPDATED = 0x10` , `FLAG_DELEGATION_UPDATED = 0x20` ...

phaazon: Ok, but `0x01 << 1` is not `0x10`. 🙂

KhalilBellakrid: :trollface:  really ? 
My answer above is about rewriting the whole section taking into account all type of transactions with logic above ! Will do it in next update !

KhalilBellakrid: Finally I used directly the Operation type to have the result of "Update" and only kept one flag for ignored transactions ! 
WDYT @phaazon ? 
And thanks again for the review ! 👍

phaazon: LGTM

PR #221, opened 2019-05-20
Add missing declaration in migrations

KhalilBellakrid:


KhalilBellakrid:
Sorry guys, I will merge even if there is only 1 approval :trollface:

PR #223, opened 2019-05-20
Change x86 android toolchain to x86_64 one

KhalilBellakrid:
Related to : https://ledgerhq.atlassian.net/browse/LLC-219

PR #233, opened 2019-06-04
Build Android x86 32bit binaries in addition to x86_64

MortalKastor:
#223 brought Android x86_64 binaries, but didn't keep 32bit ones.

But from [Google Play announcement](https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html):
> apps using native code must provide a 64-bit version (in addition to the 32-bit version)

Also Android expects 32bit x86 libs to be in the `x86` folder (and x86_64 libs to be in a `x86_64` folder).

This should fix both issues, but I wasn't able to test.

### DIFF
--- a/.circleci/build_lib.sh
+++ b/.circleci/build_lib.sh
@@ -68,8 +68,10 @@ function command_android {
     export TOOLCHAIN_NAME='android-ndk-r18b-api-21-armeabi-v7a-clang-libcxx'
   elif [ "$ARCH" == "arm64-v8a" ]; then
     export TOOLCHAIN_NAME='android-ndk-r18b-api-21-arm64-v8a-clang-libcxx'
-  else
+  elif [ "$ARCH" == "x86_64" ]; then
     export TOOLCHAIN_NAME='android-ndk-r18b-api-21-x86-64-clang-libcxx'
+  else
+      export TOOLCHAIN_NAME='android-ndk-r18b-api-21-x86-clang-libcxx'
   fi
   #This is useful for SQLCipher/config.guess
   export LIBC=gnu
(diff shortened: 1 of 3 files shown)


KhalilBellakrid:
Thanks for the PR ! 
@gre we had a talk where you asked to get rid of 32 bits, do you confirm guys that we need both ?

gre:
we need both. I don't think i ever mentionned to remove 32bits 🤔  I was thinking that google would eventually drop it, but apparently no, it's not in their plan https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html

> We are not making changes to our policy on 32-bit support. Play will continue to deliver apps to 32-bit devices. This requirement means that apps with 32-bit native code will need to have an additional 64-bit version as well.

KhalilBellakrid:
I mean I remember this conversation about 32-bit being dropped and since for us it's an additional job on CI, I removed it ! Anyway CI is going to be loooong !

gre:
:sorry: 🌍

MortalKastor:
> CI is going to be loooong

I'm no CircleCI expert but, we may save quite a lot of time by using [Docker Layer Caching](https://circleci.com/docs/2.0/docker-layer-caching/) or [CircleCI's own caching](https://circleci.com/docs/2.0/caching/) to skip the whole env setup part.

gre:
does this mean we need a new react native bindings bump too?

MortalKastor:
> does this mean we need a new react native bindings bump too?

As part of LL-1406, I'm doing the necessary changes to lib-ledger-core-react-natgive-bindings

PR #234, opened 2019-06-04
Fix versioning of published JAR

KhalilBellakrid:
I don't know why version was not changed inside of the `sbt` file, now we load directly the env variable so it should be fine !

### DIFF
--- a/.circleci/build.sbt
+++ b/.circleci/build.sbt
@@ -2,7 +2,7 @@ name := "scala-lib-core"
 
 organization := "co.ledger"
 
-version := ""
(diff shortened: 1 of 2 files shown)


.circleci/build.sbt
```diff
@@ -2,7 +2,7 @@ name := "scala-lib-core"
 
 organization := "co.ledger"
 
-version := ""
+version := sys.env.getOrElse("JAR_VERSION", "")
```
phaazon: :D

KhalilBellakrid: Meh !

PR #236, opened 2019-06-04
Install openjdk8 on Debian/Stretch

KhalilBellakrid:
Fixing the Linux CI


KhalilBellakrid:
I tried to install `openjdk11` with: 
`echo 'deb http://ftp.debian.org/debian stretch-backports main' | tee /etc/apt/sources.list.d/stretch-backports.list
apt-get update
apt-get -t stretch-backports install openjdk-11-jdk -y`
Since we use an unstable distribution it's breaking a lot of other packages ... so I rolled back to `openjdk8` (SFOL), btw I couldn't figure out the reason why we upgraded to `openjdk11` ...

PR #240, opened 2019-06-13
Bitcoin SegWit fee fix

rdbr:
The `estimateSize` function is not implemented correctly which causes SegWit transaction fees to be too high. See: https://github.com/LedgerHQ/ledger-live-desktop/issues/1714 and https://github.com/LedgerHQ/ledger-live-desktop/issues/2001

This pull request fixes the wrong calculation. I do recommend to re-evaluate the min/max estimation parameters, because the max estimate is a bit below the https://blockstream.info virtual size calculation of a real transaction.

It is also not clear if the `marker` and `flag` should be moved from the `noWitness` to `minWitness` and `maxWitness`. https://bitcoincore.org/en/segwit_wallet_dev/#transaction-fee-estimation suggests it should be moved and Electrum also adds it to the witness data. I have left it as it was for now.

### DIFF
--- a/core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
+++ b/core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
@@ -210,7 +210,7 @@ namespace ledger {
 
         api::EstimatedSize
         BitcoinLikeTransactionApi::estimateSize(std::size_t inputCount, std::size_t outputCount, bool hasTimestamp,
-                                                bool useSegwit, bool isNativeSegwit = false) {
+                                                bool useSegwit, bool isNativeSegwit) {
             // TODO Handle outputs and input for multisig P2SH
             size_t maxSize, minSize, fixedSize = 0;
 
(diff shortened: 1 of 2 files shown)


KhalilBellakrid:
Thanks for the PR, I will have a look at it ! 👍

KhalilBellakrid:
So indeed there is an issue while computing fees, we only need to replace https://github.com/LedgerHQ/lib-ledger-core/blob/develop/core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp#L239 by : 
```
minSize = (noWitness * 3 + minWitness) / 4;
maxSize = (noWitness * 3 + maxWitness) / 4;
```
After checking this part of the code, we also have to tweak the method to differentiate between Segwit and Native Segwit, we should pass the keychain engine to this method so we can compute the input size, we will end up with something like: 
```
size_t inputSize = isNativeSegwit ? 41 : 63;
```
Instead of `inputSize = 59` ...
Because for `P2SH` input has size of 63 ( `32 PrevTxHash + 4 Index + 23 scriptPubKey + 4 sequence`) and for Native Segwit it has a size of 41 (`32 PrevTxHash + 4 Index + 1 null byte + 4 sequence`)

Could you please adapt so we can be more accurate on fee computation ? 

Thanks !

rdbr:
I've fixed the estimation constants and added support for Native Segwit with a backwards-compatible isNativeSegwit parameter. The other changes were of course code-cleanups.

So, now all occurrences of `estimateSize` should be modified:
```
BitcoinLikeTransactionApi.cpp
BitcoinLikeStrategyUtxoPicker.cpp
```

KhalilBellakrid:
Thank you for your quick action !

PR #241, opened 2019-06-18
Add Terracoin support

TheSin-:
Adding Terracoin support to the core lib.

I'm just finishing up the app-btc additions but this is required so it's usable.
https://github.com/LedgerHQ/ledger-app-btc/pull/118

Let me know if I missed anything or would like anything else reviews sot hat we can make TRC use on ledger official.

### DIFF
--- a/core/src/wallet/bitcoin/networks.cpp
+++ b/core/src/wallet/bitcoin/networks.cpp
@@ -411,6 +411,21 @@ namespace ledger {
                             {}
                     );
                     return STAKENET;
+                } else if (networkName == "terracoin") {
+                    static const api::BitcoinLikeNetworkParameters TERRACOIN(
+                            "trc",
+                            {0x00},
+                            {0x05},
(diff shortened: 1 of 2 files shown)


TheSin-:
Just force pushed the commit now that it is signed

phaazon:
Thank you for your contribution. Please feel free to read the [contribution guidelines](https://github.com/LedgerHQ/lib-ledger-core/blob/master/CONTRIBUTING.md); it has plenty of information you might be interested with. Currently, we cannot accept such PRs and we cannot give a date when we will eventually.

Closing according to [this](https://github.com/LedgerHQ/lib-ledger-core/blob/master/CONTRIBUTING.md#about-coin-integration).

TheSin-:
I'm sorry I did read that I'm not sure what I did not follow?

Is it due to the version cause I felt like setting a version assumed I'd be next to be accepted which might not be true, which would then invalidate the next version.

I don't want to resubmit just to have it closed again so could I get some clarity please so I can make a proper PR?

PR #254, opened 2019-06-27
Retrieve cointype from derivation scheme

KhalilBellakrid:
We always use coinType from currency, now if the derivation scheme is setting the coinType it will be used, if not currency's coinType is used as fallback.
Note: this is needed for Live because they scan ETC accounts with `coinType = 60` which is the one of ETH !
Related to : https://ledgerhq.atlassian.net/browse/LLC-278

### DIFF
--- a/core/src/wallet/ethereum/EthereumLikeWallet.cpp
+++ b/core/src/wallet/ethereum/EthereumLikeWallet.cpp
@@ -64,6 +64,7 @@ namespace ledger {
             _observer = observer;
             _keychainFactory = keychainFactory;
             _synchronizerFactory = synchronizer;
+            _coinType = scheme.getCoinType() ? scheme.getCoinType() : network.bip44CoinType;
         }
 
         bool EthereumLikeWallet::isSynchronizing() {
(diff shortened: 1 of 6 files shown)


phaazon:
Hm, I think this will get even easier after the segregation because the factories will be taken out of the abstract core library. LGTM though.

core/src/wallet/ethereum/EthereumLikeWallet.cpp
```diff
@@ -64,6 +64,7 @@ namespace ledger {
             _observer = observer;
             _keychainFactory = keychainFactory;
             _synchronizerFactory = synchronizer;
+            _coinType = scheme.getCoinType() ? scheme.getCoinType() : network.bip44CoinType;
```
teams2ua: Ok, I was not look at this close enought, I was hopping to path coinType as a constructor argument.
But probably it will envolve more changes...

KhalilBellakrid: Exactly !

PR #276, opened 2019-07-26
Improve UTXO picking strategy DEEP_OUTPUTS_FIRST performances

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-301

### DIFF
--- a/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
+++ b/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
@@ -68,6 +68,11 @@ namespace ledger {
                 if (tx.block.nonEmpty()) {
                     sql << "UPDATE bitcoin_transactions SET block_uid = :uid WHERE hash = :tx_hash",
                             use(blockUid), use(tx.hash);
+                    for (auto &out : tx.outputs) {
+                        auto blockHeight = tx.block.getValue().height;
+                        sql << "UPDATE bitcoin_outputs SET block_height = :height WHERE transaction_hash = :tx_hash",
+                                use(blockHeight), use(tx.hash);
+                    }
                 }
                 return btcTxUid;
             } else {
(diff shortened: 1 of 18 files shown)


core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
```diff
@@ -215,20 +221,25 @@ namespace ledger {
...
                 //of same transaction (filter on account_uid won't solve the issue because
                 //bitcoin_outputs going to external accounts have NULL account_uid)
-                if (outputRow.get<std::string>(4) == BitcoinLikeTransactionDatabaseHelper::createBitcoinTransactionUid(accountUid, out.hash)) {
+                if (outputRow.get<std::string>(4) == btcTxUid) {
                     BitcoinLikeBlockchainExplorerOutput output;
                     output.index = (uint64_t) outputRow.get<int>(0);
                     output.value.assignScalar(outputRow.get<long long>(1));
                     output.script = outputRow.get<std::string>(2);
                     output.address = outputRow.get<Option<std::string>>(3);
+                    if (outputRow.get_indicator(4) != i_null) {
```
phaazon: I’m not sure this is needed if you add a non-null column in SQL (that would require a reset of the database on the client, though). Up to you.

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -97,63 +97,48 @@ namespace ledger {
...
-                    //Fix: use uniform initialization
-                    RichUTXO curr_richutxo{block_height, utxo[index]};
-                    richutxo->emplace_back(std::move(curr_richutxo));
-
-                    buddy->logger->info("Go {} on {}", index + 1, utxo.size());
-                    return go(context, index + 1, buddy, utxo, richutxo);
-                });
-            };
+            for (auto &u : utxo) {
+                uint64_t block_height = u->getBlockHeight().value_or(std::numeric_limits<uint64_t>::max());
```
phaazon: Woah that’s such a change, thank you! 😗

core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
KhalilBellakrid: Yeah that's why I didn't do so, I prefer not to force clients to reset their DB

PR #279, opened 2019-08-05
Changed the ownership mechanism of LevelDB in PreferencesBackend.

phaazon:
Instead of having a non-owning unordered_map and lots of owning
`shared_ptr` everywhere in the codebase, we now have an owning table and
weak_ptr everywhere. The idea is that, if we get a
`shared_ptr<ledger::core::PreferencesBackend>`, the underlying database is
handled via a weak_ptr. Which means that you have to lock it in order to
use the database, which allows us to destroy the database and observe it
everywhere.

Fixes LLC-198

### DIFF
--- a/core/src/preferences/PreferencesBackend.cpp
+++ b/core/src/preferences/PreferencesBackend.cpp
@@ -51,7 +51,7 @@ namespace ledger {
             : type(t), key(k), value(v) {
         }
 
-        std::unordered_map<std::string, std::weak_ptr<leveldb::DB>> PreferencesBackend::LEVELDB_INSTANCE_POOL;
+        std::unordered_map<std::string, std::shared_ptr<leveldb::DB>> PreferencesBackend::LEVELDB_INSTANCE_POOL;
         std::mutex PreferencesBackend::LEVELDB_INSTANCE_POOL_MUTEX;
 
         PreferencesBackend::PreferencesBackend(const std::string &path,
@@ -62,11 +62,11 @@ namespace ledger {
             _db = obtainInstance(_dbName);
         }
 
-        std::shared_ptr<leveldb::DB> PreferencesBackend::obtainInstance(const std::string &path) {
+        std::weak_ptr<leveldb::DB> PreferencesBackend::obtainInstance(const std::string &path) {
             std::lock_guard<std::mutex> lock(LEVELDB_INSTANCE_POOL_MUTEX);
             auto it = LEVELDB_INSTANCE_POOL.find(path);
             if (it != LEVELDB_INSTANCE_POOL.end()) {
-                auto db = it->second.lock();
+                auto db = it->second;
                 if (db != nullptr)
                     return db;
             }
@@ -83,13 +83,18 @@ namespace ledger {
             auto instance = std::shared_ptr<leveldb::DB>(db);
             std::weak_ptr<leveldb::DB> weakInstance = instance;
 
-            LEVELDB_INSTANCE_POOL[path] = weakInstance;
+            LEVELDB_INSTANCE_POOL[path] = instance;
 
-            return instance;
+            return weakInstance;
         }
 
-        void PreferencesBackend::commit(const std::vector<PreferencesChange> &changes) {
-            auto db = _db;
+        bool PreferencesBackend::commit(const std::vector<PreferencesChange> &changes) {
+            auto db = _db.lock();
+
+            if (db == nullptr) {
+              return false;
+            }
+
             leveldb::WriteBatch batch;
             leveldb::WriteOptions options;
             options.sync = true;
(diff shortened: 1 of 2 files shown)


phaazon:
I’m actually going to make a very small reproducible example to show the problem.

phaazon:
The following demonstrates the problem of our current code.

```cpp
// g++ -std=c++1y weak.cpp

#include <iostream>
#include <memory>
#include <string>
#include <unordered_map>

using namespace std;

// The scarce resource. No two instances can live at the same time in memory (i.e. lockfile).
struct Scarce {
  Scarce() {
    cout << "Scarce created" << endl;
  }

  ~Scarce() {
    cout << "Scarce dropped" << endl;
  }
};

// Preferences just holds a pointer to the scarce resource
struct Preferences {
  ~Preferences() {
    cout << "Preferences dropped" << endl;
  }

  // Simulate what happens when a thread dies.
  static void clear() {
    CACHE.clear();
  }

  // Get an instance of a scarce resource. If it’s already in memory, we don’t have to create it
  // but can use the cached version
  static shared_ptr<Scarce> obtain_instance(string const& name) {
    auto it = CACHE.find(name);

    if (it != CACHE.end()) {
      // cached resource, we just return it if it’s viable
      auto scarce = it->second.lock();
      if (scarce != nullptr) {
        return scarce;
      }
    }

    // the resource is not cached; create it
    auto scarce = make_shared<Scarce>();
    CACHE[name] = scarce;

    return scarce;
  }

private:
  shared_ptr<Scarce> _scarce;
  static unordered_map<string, weak_ptr<Scarce>> CACHE;
};

unordered_map<string, weak_ptr<Scarce>> Preferences::CACHE;

int main() {
  cout << "starting" << endl;

  {
    // get the first instance
    auto x = Preferences::obtain_instance("foo");
    auto y = Preferences::obtain_instance("foo"); // shouldn’t create
    auto z = Preferences::obtain_instance("foo"); // shouldn’t create
  }

  // here, the inst shared_ptr is not accessible and has been dropped; everything’s fine

  // now consider the following: a thread ask for an instance and keeps it around, but we reset
  // the Preferences
  {
    auto x = Preferences::obtain_instance("foo");
    Preferences::clear();
    auto y = Preferences::obtain_instance("foo"); // create while x is still living!
  }

  cout << "ending" << endl;
}
```

phaazon:
And this demonstrates the correct path:

```cpp
// g++ -std=c++1y weak.cpp

#include <iostream>
#include <memory>
#include <string>
#include <unordered_map>

using namespace std;

// The scarce resource. No two instances can live at the same time in memory (i.e. lockfile).
struct Scarce {
  Scarce() {
    cout << "Scarce created" << endl;
  }

  ~Scarce() {
    cout << "Scarce dropped" << endl;
  }
};

// Preferences just holds a pointer to the scarce resource
struct Preferences {
  ~Preferences() {
    cout << "Preferences dropped" << endl;
  }

  // Simulate what happens when a thread dies.
  static void clear() {
    CACHE.clear();
  }

  // Get an instance of a scarce resource. If it’s already in memory, we don’t have to create it
  // but can use the cached version
  static weak_ptr<Scarce> obtain_instance(string const& name) {
    auto it = CACHE.find(name);

    if (it != CACHE.end()) {
      // cached resource, we just return it if it’s viable
      auto scarce = it->second;
      if (scarce != nullptr) {
        return scarce;
      }
    }

    // the resource is not cached; create it
    auto scarce = make_shared<Scarce>();
    CACHE[name] = scarce;

    return scarce;
  }

private:
  shared_ptr<Scarce> _scarce;
  static unordered_map<string, shared_ptr<Scarce>> CACHE;
};

unordered_map<string, shared_ptr<Scarce>> Preferences::CACHE;

int main() {
  cout << "starting" << endl;

  {
    // get the first instance
    auto x = Preferences::obtain_instance("foo");
    auto y = Preferences::obtain_instance("foo"); // shouldn’t create
    auto z = Preferences::obtain_instance("foo"); // shouldn’t create
  }


  // here, the inst shared_ptr is not accessible and has been dropped; everything’s fine

  // now consider the following: a thread ask for an instance and keeps it around, but we reset
  // the Preferences
  {
    auto x = Preferences::obtain_instance("foo");
    cout << "resources will get dropped here" << endl;
    Preferences::clear();
    auto y = Preferences::obtain_instance("foo"); // create while x is still living!
  }

  cout << "ending" << endl;
}
```

phaazon:
Ping @pollastri-pierre

gre:
awesome. i hope this can be shipped soon.

phaazon:
@gre I set @KhalilBellakrid as approver too but we already have three approvals and it’s a one-commit PR. If we can have QA testing that specific situation, I think we’re safe merging it.

PR #280, opened 2019-08-20
Feature/bitcoin der signature

alekece:
Feature related to https://ledgerhq.atlassian.net/browse/LLC-247


core/idl/wallet/bitcoin/bitcoin_like_wallet.djinni
```diff
@@ -112,6 +112,12 @@ BitcoinLikeBlock = interface +c {
     getTime(): date;
 }
 
+BitcoinLikeSignature = record {
```
phaazon: Even though those fields make sense for DER / ECDSA, it’s a bit cryptic here when reading the IDL documentation. Can you provide some documentation for this type and its fields?

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
```diff
@@ -268,6 +268,54 @@ namespace ledger {
...
+                writer.writeByteArray(signatures[i].s);
+                writer.writeByte(0x01); // SIGHASH byte
+                _inputs[i]->setP2PKHSigScript(writer.toByteArray());
+                
+            }
+        }
+
+        void BitcoinLikeTransactionApi::setDERSignatures(const std::vector<std::vector<uint8_t>> & signatures) {
+            if (signatures.size() != _inputs.size()) {
+                throw make_exception(api::ErrorCode::ILLEGAL_ARGUMENT, "DER signature length differs to current input length");
```
phaazon: s/to/from

phaazon: Also, I’m not a huge fan of exceptions. An exception means weak-typed error handling and hence, there’s no way to quickly see whether a function will “handle errors.” In the case of Live / Vault, it means an Internal Error instead of explicitly handling the returned error — plus, cross-language exception handling is rarely something easy 😛

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.h
```diff
@@ -109,7 +111,10 @@ namespace ledger {
             std::vector<uint8_t> serializeOutputs() override;
 
             int32_t getVersion() override;
+            
+            void setSignatures(const std::vector<api::BitcoinLikeSignature> & signatures) override;
```
phaazon: Please provide documentation about what that function does. Especially, given the current implementation, is it idempotent?

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.h
```diff
@@ -109,7 +111,10 @@ namespace ledger {
...
 
+            void setDERSignatures(const std::vector<std::vector<uint8_t>> & signatures) override;
```
phaazon: Ditto.

alekece: Comments are present in generated code - Should I report them into class implementation ?

core/idl/wallet/bitcoin/bitcoin_like_wallet.djinni
phaazon: Thank you!

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.h
phaazon: Ah, no. My mistake.

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
alekece: Look better with specific error code ?

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
```diff
@@ -268,6 +268,61 @@ namespace ledger {
...
+            }
+            for (std::size_t i = 0; i < signatures.size(); ++i) {
+                if (_inputs[i]->getScriptSig().size() > 0) {
+                    return api::BitcoinLikeSignatureState::ALREADY_SIGNED;
+                }
+                BytesWriter writer;
+                //Var bytes Signature (prefix length)
+                //Get length of VarInt representing length of R
+                BytesWriter rLength;
+                rLength.writeVarInt(signatures[i].r.size());
```
teams2ua: Why do you use `writeVarInt`? In spec it should be just one byte, so the `writeByte` should be more appropriate.

alekece: My bad, this is something I missed from the BIP.

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
```diff
@@ -280,29 +280,29 @@ namespace ledger {
                 //Var bytes Signature (prefix length)
                 //Get length of VarInt representing length of R
                 BytesWriter rLength;
-                rLength.writeVarInt(signatures[i].r.size());
+                rLength.writeByte(signatures[i].r.size());
                 //Get length of VarInt representing length of R
                 BytesWriter sLength;
```
teams2ua: Now, as you know the size of length + size of data, you don't need this additional `BytesWriter`, the same applies for `sAndRLength`

core/src/wallet/bitcoin/networks.hpp
```diff
@@ -45,6 +45,15 @@ namespace ledger {
     namespace core {
 
         namespace networks {
+
+            enum sigHashType : uint8_t {
+                SIGHASH_ALL = 1,
+                SIGHASH_NONE = 2,
+                SIGHASH_SINGLE = 3,
+                SIGHASH_FORKID = 0x40,
```
phaazon: Can you change the way we write those literals? Mixing decimals with hexadecimals is _wat_.

PR #281, opened 2019-08-20
Add the Services object.

phaazon:
That object has the same scope as one scope of the current
WalletPool: share objects between coins without duplicating them. It
allows us to rapidly iterate on integrating coins without having to rely
on the Wallet / Account structures, which are highly convoluted.


phaazon:
I will change the name from `Meta` to `Services` according to a slack conversation with @teams2ua.

phaazon:
@teams2ua done!

phaazon:
I’m merging this because it’s just linked to a PoC right now, and since we’re two working on it, we need to be able to synchronize.

PR #282, opened 2019-08-24
Replay-By-Fee implementation

KhalilBellakrid:
A first implemetation of [RBF](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki).
**TODO**: this is a good incentive to change our generation of code to finally support unsigned integers :trollface:


phaazon:
> to finally support unsigned integers

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -708,5 +707,195 @@ namespace ledger {
...
+                                        out,
+                                        self->getWallet()->getCurrency()
+                                )
+                        );
+                        tx->removeOutput(outputChange->getOutputIndex());
+                    } else {
+                        // Create a change
+                        auto changeAddress = self->getKeychain()->getFreshAddress(BitcoinLikeKeychain::CHANGE)->toString();
+                        BitcoinLikeBlockchainExplorerOutput out;
+                        out.index = static_cast<uint64_t>(tx->getOutputs().size());
```
phaazon: I’m just realizing that all that “create a new change output” is a bit `unsafe` / _doing the same thing over and over_. I’m not asking any change for that PR but I’ll keep the idea in mind for our next iteration of libcore. We should have somekind of middleware / abstraction that does that work for us so that we don’t even have to manipulate the keychain logic.

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.cpp
```diff
@@ -262,6 +262,13 @@ namespace ledger {
             return *this;
         }
 
+        BitcoinLikeTransactionApi &BitcoinLikeTransactionApi::removeOutput(uint64_t index) {
```
phaazon: I think it’s okay regarding your code, but it’s easy to shoot ourselves in the foot with that kind of stuff, since it invalidates by one decrement everything right in the `_outputs` vector after the erasure.

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -708,5 +707,195 @@ namespace ledger {
             return _explorer->getFees().callback(getContext(), callback);;
         }
 
+        FuturePtr<api::BitcoinLikeTransaction> BitcoinLikeAccount::getReplayByFeeTransaction(const std::string & hash, const std::shared_ptr<api::Amount> & additionaFeesPerBytes, bool forceBuild) {
```
teams2ua: As I understand forceBuild is here only for testing purpose.
I am not a big fun of approach to include testing logic in "real" code, because it makes the "real" code harder to reason about, and open the door for new errors.
What we can do here is write a function that create RBF transaction having other transaction as input. So in tests we can get some transaction and directly call this new function and in "real" code first get transaction, then check the transaction and then call the `createRBFFromTransaction(checkedTransaction)`

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -708,5 +707,195 @@ namespace ledger {
...
+                );
+
+                // Get input amounts
+                BigInt inputsValue;
+                {
+                    soci::session sql(self->getWallet()->getDatabase()->getPool());
+                    for (auto &input : tx->getInputs()) {
+                        // Check if sequences are lower than 0xffffffff otherwise impossible to use RBF
+                        // TODO: support uints in Djinni
+                        if (!forceBuild && input->getSequence() == std::numeric_limits<uint32_t>::max()) {
```
teams2ua: Here you only check "explicit signaling". Are we going to support "Inherited signaling" as well?

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -708,5 +707,195 @@ namespace ledger {
...
+                        tx->addOutput(std::make_shared<BitcoinLikeOutputApi>(out, self->getWallet()->getCurrency()));
+                        tx->removeOutput(outputChange->getOutputIndex());
+                        incrementSequence(tx);
+                        return FuturePtr<api::BitcoinLikeTransaction>::successful(tx);
+                    }
+                }
+
+                // Construct new outputs taking into account fees
+                // Pick new UTXO
+                return self->getUTXO().mapPtr<api::BitcoinLikeTransaction>(self->getContext(), [=](const std::vector<std::shared_ptr<api::BitcoinLikeOutput>> &utxos) {
```
teams2ua: If we have a change in the transaction that we want to modify, this change will be in the list of UTXOs (right?), you don't take this into account when aggregating UTXOs to adjust the sum, and later you try to change the amount of the change.

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
phaazon: I think we have to if we want to be RBF compliant.

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
KhalilBellakrid: > Very good PR, thank you! Just a few questions:
> 
> * I am right to guess that currently, that PR doesn’t allow a user to confirm a RBF-signaled TX?
> * Currently, we do not check whether a transaction is confirmed when we retrieve them. So it means that we might spend unconfirmed transactions.
> * On the same leve, does it mean that we display unconfirmed amounts to the user on a `getBalance`?

What do you mean by confirming an RBF-signaled tx ? 
If you are talking about additional UTXOs that we might need, no we don't check but we take deeper first, so worst case all UTXOs are generated by unconfirmed txs and we can't do better than that in that case,
Yes, because of the optimistic update, but again if a tx got dropped the mechanism of sync is removing the tx from our db and the `getBalance` will return the new correct balance

KhalilBellakrid:
> Very good PR, thank you! Just a few questions:
> 
> * I am right to guess that currently, that PR doesn’t allow a user to confirm a RBF-signaled TX?
> * Currently, we do not check whether a transaction is confirmed when we retrieve them. So it means that we might spend unconfirmed transactions.
> * On the same leve, does it mean that we display unconfirmed amounts to the user on a `getBalance`?

@phaazon What do you mean by confirming an RBF-signaled tx ? 
If you are talking about additional UTXOs that we might need, no we don't check but we take deeper first, so worst case all UTXOs are generated by unconfirmed txs and we can't do better than that in that case,
Yes, because of the optimistic update, but again if a tx got dropped the mechanism of sync is removing the tx from our db and the `getBalance` will return the new correct balance

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
KhalilBellakrid: For the moment we are supporting just the `explicit` RBF but we should support the `implicit` as well later

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
KhalilBellakrid: Hmm if we do that we'll have to go through inputs twice, since we still need the inputs' value, unless I didn't really get what you want to say ? 🤔 
And here we are talking about a boolean which explicitly means what it does and used at only one condition so not adding that much complexity, if there were too much branching and additional parameters I would agree but it still to be discusses VS adding other methods with different signatures 🤔

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
KhalilBellakrid: So again since we are taking the deepest first we should be safe unless the only availale UTXO is the one generated by the previous tx we want tto use RBF on, in which case it will always fail ... 
But we can explicitly add a test and fail on creating RBF transaction in that case 🤔

phaazon:
`HODL` for stalling PR.

pollastri-pierre:
Closing this PR, we will probably handle RBF transaction through client instead of client (they have everything needed to make RBF transactions from libcore API)

PR #283, opened 2019-08-25
Fix Tezos

KhalilBellakrid:
- Related to [LLC-315](https://ledgerhq.atlassian.net/browse/LLC-315): Fix account creation info returned by `TezosLikeWallet`,
- Related to [LLC-312](https://ledgerhq.atlassian.net/browse/LLC-312): expose `getFees` method from `TezosLikeAccount`.

### DIFF
--- a/core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
@@ -69,8 +69,9 @@ namespace ledger {
 
         Future<std::shared_ptr<BigInt>>
         NodeTezosLikeBlockchainExplorer::getFees() {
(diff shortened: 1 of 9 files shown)


core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
```diff
@@ -69,8 +69,9 @@ namespace ledger {
 
         Future<std::shared_ptr<BigInt>>
         NodeTezosLikeBlockchainExplorer::getFees() {
+            bool parseNumbersAsString = true;
```
phaazon: +1 for putting that in a variable. `json(true)` would sound very weird to me, but `json(parseNumbersAsString)` is okay in that case.

PR #288, opened 2019-09-01
Fix XTZ

KhalilBellakrid:
Add final endpoints on Tzscan to get counter, storage and estimated gas limit.


core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.h
```diff
@@ -87,6 +87,11 @@ namespace ledger {
 
             std::string getExplorerVersion() const override;
 
+            Future<std::shared_ptr<BigInt>>
```
phaazon: This should be at least `protected` to me.

phaazon: Also, please provide some documentation about what it’s for.

core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
```diff
@@ -210,32 +210,37 @@ namespace ledger {
...
 
+        Future<std::shared_ptr<BigInt>> NodeTezosLikeBlockchainExplorer::getHelper(const std::string &url,
+                                                                                   const std::string &field,
+                                                                                   const std::unordered_map<std::string, std::string> &params) {
+            bool parseNumbersAsString = true;
+            auto networkId = getNetworkParameters().Identifier;
+
+            std::string p;
+            for (auto &param : params) {
+                std::string separator = p.empty() ? "?" : "&";
```
phaazon: This test is going to be done for all elements while, after the first iteration, we know it will always be `true`.

PR #293, opened 2019-09-06
Fix XRP accounts from being created with account index = 0.

phaazon:
Convenience for mister @onyb.

### DIFF
--- a/core/src/wallet/ripple/RippleLikeWallet.cpp
+++ b/core/src/wallet/ripple/RippleLikeWallet.cpp
@@ -121,6 +121,7 @@ namespace ledger {
 
         FuturePtr<ledger::core::api::Account>
         RippleLikeWallet::newAccountWithExtendedKeyInfo(const api::ExtendedKeyAccountCreationInfo &info) {
+            logger()->debug("Creating new XRP account (index = {}) with extended key info", info.index);
 
             if (info.extendedKeys.empty()) {
                 throw make_exception(api::ErrorCode::INVALID_ARGUMENT,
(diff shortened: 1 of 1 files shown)


phaazon:
Some context about this PR. @KhalilBellakrid made some changes in #273 that I don’t understand, especially _getting an account index from a derivation scheme_. A derivation scheme is, to me, a _placeholder-like_ object that **shouldn’t have any value linked to it**. Converting from a `DerivationScheme` to a `DerivationPath` shouldn’t be that implicit and should require an explicit _substitute_ function.

In this case, the “bug” is due to the fact the `DerivationScheme` is considered substituted while it’s not: we then get the default memoy value for the `ACCOUNT_INDEX` node, which is `0` for `uint32_t`.

onyb:
Tested the PR with the wallet daemon, and it fixes the issue I was facing. ✔️

KhalilBellakrid:
This was done to support potential exotic paths that Live has to deal with.
If user creates a wallet with derivation scheme where account level is hardened then it should read the account's index from that same derivation scheme, otherwise it will read it from the account creation infos provided when creating the account.
So this should not be a fix, it's a workaround, @onyb please check the derivation scheme that you're using in your configuration of the wallet pool ?

phaazon:
@KhalilBellakrid yes, we checked, the account derivation is hardened.

phaazon:
Ok, regarding @KhalilBellakrid comments, **we are not going to merge that yet.** I will try to find another way and pinpoint the acutal problem. Maybe the problem is linked to the setup of the account on @onyb’s side and we need a more explicit error message. Please provide the exact derivation path you used. I remember it was hardened but I don’t remember whether it had a value or not.

If it has a value, like `m/44'/144'/4'/0/<address>`, then there’s a bug in the derivation scheme parser code because the account node should be set to `4` here.

PR #296, opened 2019-09-10
Make SQLCipher to use memory to implement temp fs instead of files. 

phaazon:
This might be needed as none of the default SQLite temporary directories
(/usr/tmp, /tmp/, ., etc.) are available to be used as temporary
directories. I still don’t quite get why we haven’t bumped into that
situation yet but after analysis of LLC-292, I ended up by having to
lookup the `SQLITE_IOERR_GETTEPMPATH` error raised by SQLite3. That
error is set, on Unix systems, in the `unixGetTempname` function when it
fails to get a temporary file directory. In order to get it, it uses
internally the `unixTempFileDir` function, which is basicaly:

```c
static const char *unixTempFileDir(void){
  static const char *azDirs[] = {
     0,
     0,
     "/var/tmp",
     "/usr/tmp",
     "/tmp",
     "."
  };
  unsigned int i = 0;
  struct stat buf;
  const char *zDir = sqlite3_temp_directory;

  if( !azDirs[0] ) azDirs[0] = getenv("SQLITE_TMPDIR");
  if( !azDirs[1] ) azDirs[1] = getenv("TMPDIR");
  while(1){
    if( zDir!=0
     && osStat(zDir, &buf)==0
     && S_ISDIR(buf.st_mode)
     && osAccess(zDir, 03)==0
    ){
      return zDir;
    }
    if( i>=sizeof(azDirs)/sizeof(azDirs[0]) ) break;
    zDir = azDirs[i++];
  }
  return 0;
}
```

The `azDirs` array is traversed and exhausted because no environment
variable is passed and none of "/var/tmp", "/usr/tmp", "/tmp" and "."
are available as read-write directories.

One possible alternative to look at: pass en environment variable set to
"/data/local/tmp", but I’m not sure this location is really safe nor
even available on all devices.

Relates to LLC-292.

2143dbc


phaazon:
`HODL` because currently there’s a bug in Live Android with a `std::bad_cast` I’m trying to solve.

phaazon:
Removing `HODL` because we’ve found the bug and it’s going to be fixed. Basically, it was a bug in some _Bitcoin_ transaction inflate functions.

core/lib/sqlcipher/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
...
 
+message(STATUS "Configuring SQLCipher:")
+
 add_library(sqlcipher STATIC sqlite3.c sqlite3.h sqlite3ext.h)
 
+if (${CMAKE_SYSTEM_NAME} STREQUAL "Android")
+  # On Android, temporary directories are a bit special and are not located in the same place, such
+  # as /tmp or /usr/tmp; for this reason, we ask temporary files to be stored in memory instead
+  message("Compiling SQLCipher for Android without temporary directory disabled (all-on-memory).")
+  add_definitions(-DSQLITE_TEMP_STORE=3)
```
teams2ua: Better place for this line is core/src/CMakeLists.txt, and we already have ANDROID branch there.
It would be better to not touch this file as it is kind on 3rd-party

phaazon: Yeah, you are totally right. I’ll change that.

phaazon: `core/lib/CMakeLists.txt`, you meant.

PR #298, opened 2019-09-11
Add asTezosLikeAccount.

phaazon:
Relates to LLC-345.

Ping @gre


gre:
LGTM. we'll need to regen the bindings projects?

PR #299, opened 2019-09-12
Add configuration for RIPPLE_LAST_LEDGER_SEQUENCE_OFFSET.

phaazon:
# Content

This PR supplements #290 as some teams (i.e. Vault, ping @onyb) require
an higher value for the offset (default is 4 which is too short to get
all the signatures).

Possible alternative was to set maxLedgerVersion to `null` [but this is highly discouraged](https://xrpl.org/reliable-transaction-submission.html#lastledgersequence).

Relates to LLC-346.

# Weird picture of a dog


onyb:
Tested OK in my integration with the wallet daemon: https://github.com/LedgerHQ/ledger-wallet-daemon/commit/3bfc0797a75947cb066169ab7027259a126de666  ✔️

gre:
4 is very very fast; in Ledger Live we moved to 12 and it's still too fast for some users with just a nano S.
great you intro that config.

@juan-cortes we'll not forget to integrate this. we'll discuss a bit on our side

phaazon:
@gre glad to hear. Maybe we can change the default if we think it’s already too fast for our Ledger use case. 4 is the recommended default value but we can set whatever value we want. :)

gre:
4 means you have an average of 16 seconds to validate with the device (modulo the time it takes to reach the device). in context of Ledger, you need the time for user to effectively verify on the device the address so that time is blocking. 16 sec is way too fast for this.

4 makes sense for non-blocking systems where you don't wait user to validate.

phaazon:
@gre yep, that’s why I also mentioned for our Ledger case because Live or Vault, there is a manual, human interaction between the creation of the transaction and its validation and broadcasting. In that sense, it kinda makes sense to me to change the default value to a higher one. What do you suggest?

phaazon:
@onyb merged.

PR #300, opened 2019-09-13
Add support for XRP sequence when converting an Operation into a RippleLikeTransactionApi.

phaazon:
# Content

That should fix the `nullptr` exception @meriadec saw.

Relates to LLC-349.

# Weird picture of an animal

### DIFF
--- a/core/src/database/migrations.cpp
+++ b/core/src/database/migrations.cpp
@@ -667,5 +667,12 @@ namespace ledger {
 
         template <> void rollback<13>(soci::session& sql) {
         }
+
+        template <> void migrate<14>(soci::session& sql) {
+            sql << "ALTER TABLE ripple_transactions ADD COLUMN sequence BIGINT";
+        }
+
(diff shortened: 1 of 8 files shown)


core/src/database/migrations.cpp
```diff
@@ -408,7 +408,8 @@ namespace ledger {
                     "sender VARCHAR(255) NOT NULL,"
                     "receiver VARCHAR(255) NOT NULL,"
                     "fees VARCHAR(255) NOT NULL,"
-                    "confirmations BIGINT NOT NULL"
+                    "confirmations BIGINT NOT NULL",
+                    "sequence BIGINT NOT NULL",
```
phaazon: WHOOOOOOOOOPS, need to remove that.

phaazon:
> Maybe add a test to see if it parses right, getSequence returns a non null pointer

Oh yes I forgot. On my way.

PR #301, opened 2019-09-13
Throw exception if Ripple node does not return tesSUCCESS

onyb:
### What is this about?

On broadcasting a transaction, the Ripple node may respond with a transaction hash, even though the `engine_result` key has an error code. See the list of all engine result codes here: https://xrpl.org/transaction-results.html

We could also simply checked if `engine_result` startswith `tes`, but [according to the docs](https://xrpl.org/tes-success.html):
> The code tesSUCCESS is the only code that indicates a transaction succeeded.

Hence, this should be sufficient for now.

**Bonus:** Also fixed a typo that I discovered in the same file. Not sure if the function was used by anyone.

### DIFF
--- a/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
@@ -189,7 +189,7 @@ namespace ledger {
         Future<Bytes> NodeRippleLikeBlockchainExplorer::getRawTransaction(const String &transactionHash) {
             NodeRippleLikeBodyRequest bodyRequest;
             bodyRequest.setMethod("tx");
-            bodyRequest.pushParameter("trasnaction", transactionHash);
+            bodyRequest.pushParameter("transaction", transactionHash);
             bodyRequest.pushParameter("binary", "true");
             auto requestBody = bodyRequest.getString();
             std::unordered_map<std::string, std::string> headers{{"Content-Type", "application/json"}};
--- a/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
@@ -154,18 +154,27 @@ namespace ledger {
                             !json["result"].IsObject()) {
                             throw make_exception(api::ErrorCode::HTTP_ERROR, "Failed to broadcast transaction, no (or malformed) field \"result\" in response");
                         }
-                        //Is there an tx_json field ?
+
                         auto resultObj = json["result"].GetObject();
-                        if (!resultObj.HasMember("tx_json") || !resultObj["tx_json"].IsObject()) {
+
+                        if (resultObj.HasMember("engine_result") && resultObj["engine_result"] == "tesSUCCESS") {
+                          // Check presence of tx_json field
+                          if (!resultObj.HasMember("tx_json") || !resultObj["tx_json"].IsObject()) {
                             throw make_exception(api::ErrorCode::HTTP_ERROR, "Failed to broadcast transaction, no (or malformed) field \"tx_json\" in response");
+                          }
+                          auto txnObj = resultObj["tx_json"].GetObject();
+
+                          // Check presence of hash field
+                          if (!txnObj.HasMember("hash") || !txnObj["hash"].IsString()) {
+                            throw make_exception(api::ErrorCode::HTTP_ERROR, "Failed to broadcast transaction, no (or malformed) field \"hash\" in response");
+                          }
+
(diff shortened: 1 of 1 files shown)


phaazon:
I remember we discussed that live. LGTM.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -154,18 +154,17 @@ namespace ledger {
...
                         auto resultObj = json["result"].GetObject();
-                        if (!resultObj.HasMember("tx_json") || !resultObj["tx_json"].IsObject()) {
-                            throw make_exception(api::ErrorCode::HTTP_ERROR, "Failed to broadcast transaction, no (or malformed) field \"tx_json\" in response");
-                        }
 
-                        //Is there an hash field ?
-                        auto jsonObj = resultObj["tx_json"].GetObject();
-                        if (!jsonObj.HasMember("hash") || !jsonObj["hash"].IsString()) {
-                            throw make_exception(api::ErrorCode::HTTP_ERROR, "Failed to broadcast transaction, no (or malformed) field \"hash\" in response");
+                        if (resultObj.HasMember("engine_result") && resultObj["engine_result"] == "tesSUCCESS") {
```
phaazon: Do you know what happens if `resultObj["engine_result"]` is not there? Can it be compared to a string as you do? I guess yes but I’d like to be sure :D

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -154,18 +154,17 @@ namespace ledger {
...
+                          auto txnObj = resultObj["tx_json"].GetObject();
```
phaazon: You forgot to test whether `"tx_json"` is in `resultObj`. :wink:

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -154,18 +154,17 @@ namespace ledger {
...
+                          return txnObj["hash"].GetString();
```
phaazon: Same for `"hash"`.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
```diff
@@ -180,7 +179,7 @@ namespace ledger {
         Future<Bytes> NodeRippleLikeBlockchainExplorer::getRawTransaction(const String &transactionHash) {
             NodeRippleLikeBodyRequest bodyRequest;
             bodyRequest.setMethod("tx");
-            bodyRequest.pushParameter("trasnaction", transactionHash);
```
phaazon: :rofl:

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
onyb: If `resultObj["engine_result"]` is not there, I guess the second condition `resultObj["engine_result"] == "tesSUCCESS"` will be skipped. I could be wrong, but that's the behaviour in 🐍.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
onyb: I intentionally removed those checks, because if the status is `tesSUCCESS`, then Ripple's documentation provides us with the guarantee of a predictable JSON structure.

If you think it's not a good practice, then I'll add the checks.

phaazon: I get your point, but the JSON is broacasted on a network and could be tampered with. Or a node might have a bug and provide an ill-formed JSON data.

I always take the side of “everything that comes from the user, the network, or any other service” is not trustworthy and should be checked. The problem with such assumptions is that it’s very easy to break our lib by providing a JSON structure sligthly different. Instead of crashing the app, we should emit en error stating that a key is missing.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
phaazon: Oh yeah I had not seen the first part of the test. LGTM.

core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
onyb: Done. 👍

PR #303, opened 2019-09-16
Add generator of `AbstractWalletFactory`

alekece:
The goal here is to abstract the creation of abstract wallet - that means to remove that old fashion switch/case with coin related codes.
In that PR, I propose to handle that creation into a new class `AbstractFactoryGenerator`. This generator can learn at runtime new factories identified by templated-type key - in our case the string representing the coin name.


ledger-core/inc/core/utils/AbstractFactoryGenerator.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    AbstractFactoryGenerator() = default;
+    ~AbstractFactoryGenerator() = default;
+
+    value_type add(key_type const& key, value_type const& factory);
+    void remove(key_type const& key);
+    value_type make(key_type const& key) const;
+
+private:
+    std::map<key_type, value_type> factories_;
```
teams2ua: Why not just `unordered_map<string, shared_ptr<AbstractFactory>>`.
There is to many names and abstractions to remember by I already know what is `unordered_map` and how does it behave.

ledger-core/inc/core/utils/AbstractFactoryGenerator.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <map>
+#include <memory>
+#include <type_traits>
+
+namespace ledger
+{
+namespace core
+{
+
+template <typename Key, typename T>
```
phaazon: Please add documentation to what that class is for. By reading the code it’s very abstract.

ledger-core/inc/core/utils/AbstractFactoryGenerator.hpp
alekece: This is the current implementation in `WalletPool` - `std::vector` is used but anyway - and I'm ok with that. 
However add a layer of abstraction with a well-known class is always a best practice. The readability is increased and all the burden about the STL containers is hidden.
For example: 
```
auto fIt = std::find_if(_factories.begin(), _factories.end(), [currencyName] (const std::shared_ptr<AbstractWalletFactory>& f) {
                    return currencyName == f->getCurrency().name;
                });
...
                if (fIt != _factories.end())
                    _factories.erase(fIt);
```
compare to
```
factories_.forget(currencyName)
```

teams2ua: `However add a layer of abstraction with a well-known class is always a best practice` this is debatable, personally I think this add complexity for relatively small profit, I think we need someones else opinion on this @phaazon @pollastri-pierre 
Your example code is for `vector` for unordered_map it should be much simpler IMO:
```
map.erase(currencyName);
```

alekece: As I said before, I agree with the current implementation but with the usage of `map` instead  of `vector`. 
In our case, the generator is quite simple - then complexity keeps thin - for now and should probably evolve with the incoming cryptocurrencies.

phaazon: My opinion on this is that we should be using the right data structure to do a given job. If we don’t care about the order, we should be using `unordered_map`. The reason for this is both correctness and performance: maintaining a `std::map`’s internal structure (balancing) to preserve order has a cost. We shouldn’t pay for it if we don’t care.

To give more hindsight, C++ has the naming wrong. `std::map` is okay but `std::unordered_map` should be called `std::hash_map`. Here, a _hash map_ is the right tool so I agree with @teams2ua. For factories I see zero reason why we would need a binary tree map or red-black tree or whatever.

@teams2ua mentioned another point we haven’t talked and that I think we should talk to:

```cpp
     std::map<key_type, value_type> factories_; 
```

`key_type` and `value_type`, here, don’t conform to the naming convention we’ve been using in the codebase, which is PascalCase for types and type variables.

And about `std::vector`, I’m totally fine to use it _first_. We’re talking about factories, which we won’t have a lot of. I think the debate about `std::vector` is sterile as the performance between a linear search and a binary search with that few items will be very small.

alekece: We all agree to use `std::unordered_map`. The opinion conflict here is to use directly STL container in the WalletPool or use that class I made to abstract the generation of abstract factories.
@phaazon I ignore we have such naming convention about using aliased type - I don't see any in the code base

phaazon: We don’t use aliased types. But types follow the PascalCase convention. :)

PR #305, opened 2019-09-16
Fix/eth negative balances

phaazon:
# Content

This patch should fix the negative balances (and on more general level, wrong) balance histories on ETH. It takes into account the _internal transactions_ made on the Ethereum blockchain by looking them up in the core local DB when we compute the internal history.

The code is a bit hacky and messy around the usage of `Operation` but it’s already the case (i.e. the `Operation` used in the `getBalanceHistory` is partially initialized, which is very bad, so we will want to move from that scheme and use another type with less fields — basically we need only fees, values, dates and that’s pretty much all).

# Weird animal picture


core/src/wallet/ethereum/EthereumLikeAccount.cpp
```diff
@@ -317,9 +342,41 @@ namespace ledger {
...
+                      // a buffer for allocating internal operations
+                      std::vector<InternalTransaction> internalTransactions;
+                      std::vector<Operation> internalOperations;
+
+                      // this loop get all the internal transactions related to a given ETH
+                      // operation and will transform them into regular Operation
+                      for (auto const& operation : operations) {
+                        // get this operation’s internal tx
+                        internalTransactions.clear();
+                        getInternalTransactions(sql, operation, internalTransactions);
```
teams2ua: A little bit afraid about performance of this. For each operation we will do request to DB.
Maybe we need to redefine what is inside `Operation` so getOperation will return everything with one select with JOIN

phaazon: It’s a cold path so I’d say it’s okay.

phaazon: But I agree that we could do it at once. I could write such a SQL query actuall and bake it in `getOperationsWithInternal`.

phaazon:
@teams2ua I completely agree. The overall code is a bit hacky and I’m patching with band-aid here. One point I especially don’t like is that we use `Operation` as _temporary object_ (if you look at the `while` loops, they just don’t care about most fields of `Operation` but a few, like `date`, `amount`, `fees`).

phaazon:
@malikfassi do you think there’s a way for you to try that to see whether the balances look better before we merge? I think we have already synchronization tests so I would say I haven’t brought any regression but I’m not sure at all. Ping @pollastri-pierre and @KhalilBellakrid

core/src/wallet/ethereum/EthereumLikeAccount.cpp
```diff
@@ -249,6 +250,30 @@ namespace ledger {
             }
         }
 
+        void EthereumLikeAccount::getInternalTransactions(
+            soci::session &sql,
+            Operation const& parent,
+            std::vector<InternalTransaction>& internalTransactions
+        ) {
+          soci::rowset<soci::row> rows = (sql.prepare <<
+              "SELECT io.type, io.value, io.sender, io.receiver, io.gas_limit, io.gas_used"
```
KhalilBellakrid: Here you'll need a trailing space

core/src/wallet/ethereum/EthereumLikeAccount.cpp
```diff
@@ -249,6 +250,30 @@ namespace ledger {
...
+              "FROM internal_operations as io"
```
KhalilBellakrid: Ditto !

KhalilBellakrid:
Nice catch and thank you for the fix ! 👏 
This method is only used by Vault for the moment that's why we didn't catch it when adding internal transactions for Live !

core/src/wallet/ethereum/EthereumLikeAccount.cpp
phaazon: Whoopsie, nice catch my sir!

core/src/wallet/ethereum/EthereumLikeAccount.cpp
phaazon: :+1:

core/src/wallet/ethereum/EthereumLikeAccount.cpp
phaazon: Done.

core/src/wallet/ethereum/EthereumLikeAccount.cpp
phaazon: Done too.

core/test/integration/synchronization/ethereum_synchronization.cpp
```diff
@@ -82,6 +82,7 @@ TEST_F(EthereumLikeWalletSynchronization, MediumXpubSynchronization) {
                               api::EventCode::SYNCHRONIZATION_SUCCEED);
 
                     auto balance = wait(account->getBalance());
+                    std::cout << "Balance: " << balance->toString() << std::endl;
```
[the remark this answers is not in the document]
phaazon: Same as below (plus this was already there).

phaazon:
Currently, this is segfaulting in the integration tests…

core/src/wallet/common/database/OperationDatabaseHelper.cpp
```diff
@@ -160,17 +160,21 @@ namespace ledger {
                 auto recipients = strings::split(row.get<std::string>(5), ",");
                 if ((type == api::OperationType::SEND && row.get_indicator(4) != i_null && filterList(senders)) ||
                     (type == api::OperationType::RECEIVE && row.get_indicator(5) != i_null && filterList(recipients))) {
-                    operations.resize(operations.size() + 1);
-                    auto& operation = operations[operations.size() - 1];
+                    Operation operation;
+                    //operations.resize(operations.size() + 1);
```
teams2ua: remove commented lines

phaazon:
Done.

Screenshot 2019-10-16 at 11 09 07

core/test/integration/synchronization/ethereum_synchronization.cpp
```diff
@@ -113,12 +117,24 @@ TEST_F(EthereumLikeWalletSynchronization, MediumXpubSynchronization) {
                     auto tx = wait(txBuilder->build());
                     EXPECT_EQ(tx->getValue()->toLong(), balance->toLong() - (gasLimit * gasPrice));
 
+                    auto now = std::time(nullptr);
+                    char now_str[256];
+                    std::strftime(now_str, sizeof(now_str), "%Y-%m-%dT%H:%M:%SZ", std::localtime(&now));
+                    std::cout << now_str << std::endl;
```
teams2ua: remove cout

core/test/integration/synchronization/ethereum_synchronization.cpp
```diff
@@ -136,7 +152,7 @@ TEST_F(EthereumLikeWalletSynchronization, MediumXpubSynchronization) {
         auto account = std::dynamic_pointer_cast<EthereumLikeAccount>(wait(wallet->getAccount(0)));
         EXPECT_EQ(account->getERC20Accounts().size(), erc20Count);
 
-        wait(pool->freshResetAll());
+        //wait(pool->freshResetAll());
```
teams2ua: remove or uncomment

core/test/integration/synchronization/ethereum_synchronization.cpp
```diff
@@ -82,6 +84,8 @@ TEST_F(EthereumLikeWalletSynchronization, MediumXpubSynchronization) {
                               api::EventCode::SYNCHRONIZATION_SUCCEED);
 
                     auto balance = wait(account->getBalance());
+                    balanceStr = balance->toString();
+                    std::cout << "Balance: " << balanceStr << std::endl;
```
teams2ua: remove this, or add check instead of output

core/test/integration/synchronization/ethereum_synchronization.cpp
```diff
@@ -113,12 +117,24 @@ TEST_F(EthereumLikeWalletSynchronization, MediumXpubSynchronization) {
...
+                    std::strftime(now_str, sizeof(now_str), "%Y-%m-%dT%H:%M:%SZ", std::localtime(&now));
+                    std::cout << now_str << std::endl;
+
+                    auto history = wait(account->getBalanceHistory(
+                        "2019-09-20T00:00:00Z",
+                        now_str,
+                        api::TimePeriod::DAY
+                    ));
+
+                    EXPECT_EQ(history.back()->toString(), balanceStr);
```
teams2ua: [OPT] weak test :-p

phaazon: What do you mean weak test? I don’t really see what else we can do since we don’t know any of the previous balances but the last one. :)

core/test/integration/synchronization/ethereum_synchronization.cpp
phaazon: Yeah I need to uncomment, thanks! :)

core/test/integration/synchronization/ethereum_synchronization.cpp
phaazon: Sure thing.

core/src/wallet/common/database/OperationDatabaseHelper.cpp
phaazon: Ja.

core/test/integration/synchronization/ethereum_synchronization.cpp
teams2ua: Took the address which had a problem (ping @malikfassi ), and check each value that was wrong before and is correct now.

phaazon: I think it was a private-like xpub we cannot put in our integration tests IIRC.

phaazon:
> Awesome !
> Last thing to take into consideration: if an operation failed, added amount should be `0` !
> I've seen that you did something for internal transactions, but we miss it for ops themselves, because an operation with no internal txs could fail !
> Thanks in advance for the changes !

Sure thing, I’ll change it.

PR #306, opened 2019-09-17
Fix secp256k1 point construction.

phaazon:
We were not testing the return of secp256k1_ec_pubkey_parse correctly.

Relates to LLC-347.


phaazon:
Merging as it’s a really easy change and I’m 100% sure.

PR #307, opened 2019-09-17
Some SECP256k1Point enhancement

teams2ua:
Remove unused methods, save allocations,
provide configuration-free implementation of some functions

SECP256k1Point simplification


core/src/crypto/DeterministicPublicKey.cpp
```diff
@@ -113,9 +113,8 @@ namespace ledger {
                 throw Exception(api::ErrorCode::UNSUPPORTED_OPERATION, "Cannot derive key - IL >= N");
             }
 
-            SECP256k1Point K = SECP256k1Point(_key).generatorMultiply(IL.toByteArray());
             return DeterministicPublicKey(
-                    K.toByteArray(),
+                    SECP256k1Point(_key).generatorMultiply(IL.toByteArray()).toByteArray(),
```
phaazon: I don’t get why that change.

core/src/crypto/SECP256k1Point.cpp
```diff
@@ -36,76 +36,37 @@
 
 namespace ledger {
     namespace core {
+        SECP256k1Point::ContextUniqPtr SECP256k1Point::_context = ContextUniqPtr(secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY), secp256k1_context_destroy);
```
phaazon: Hm, I’m not sure about having that here. Have you considered implications with thread-safe code?

teams2ua: Initialization is safe, as this is `static` variable. From docs
```/** Opaque data structure that holds context information (precomputed tables etc.).
 *
 *  The purpose of context structures is to cache large precomputed data tables
 *  that are expensive to construct, and also to maintain the randomization data
 *  for blinding.
 *
 *  Do not create a new context object for each operation, as construction is
 *  far slower than all other API calls (~100 times slower than an ECDSA
 *  verification).
 *
 *  A constructed context can safely be used from multiple threads
 *  simultaneously, but API call that take a non-const pointer to a context
 *  need exclusive access to it. In particular this is the case for
 *  secp256k1_context_destroy and secp256k1_context_randomize.
 *
 *  Regarding randomization, either do it once at creation time (in which case
 *  you do not need any locking for the other calls), or use a read-write lock.
 */
typedef struct secp256k1_context_struct secp256k1_context;```

File: core/src/crypto/DeterministicPublicKey.cpp (hunk 1 above)
Comment (teams2ua) [reply core/src/crypto/DeterministicPublicKey.cpp]: To not have `SECP256k1Point K = blablabla`, and avoid having copies (now everything are moving), but actually now when I add move constructor it should be fine.

**Comment by phaazon** [comment]
LGTM. Your conflict is likely to be linked to #306 where I changed the check on the return value.

File: core/src/crypto/DeterministicPublicKey.cpp (hunk 1 above)
Comment (alekece) [reply core/src/crypto/DeterministicPublicKey.cpp]: You got a great point here but when I read that portion of code, the move mechanism isn't obvious.
I suggest you to keep the `K` variable and add an appropriate `std::move` in the constructor.

File: core/src/crypto/SECP256k1Point.hpp [hunk 3]
```diff
@@ -39,21 +39,16 @@ namespace ledger {
     namespace core {
         class SECP256k1Point {
         public:
-            SECP256k1Point(const std::vector<uint8_t>& p);
-            SECP256k1Point operator+(const SECP256k1Point& p) const;
+            explicit SECP256k1Point(const std::vector<uint8_t>& p);
             SECP256k1Point generatorMultiply(const std::vector<uint8_t>& n) const;
-            SECP256k1Point(const SECP256k1Point& p);
+            SECP256k1Point(SECP256k1Point&& p) = default;
```
alekece: You missed the move assignment operator here - [rule of thee](https://en.cppreference.com/w/cpp/language/rule_of_three)

teams2ua: Done. I just remove everything (except ctr) rule of zero :)

core/src/crypto/DeterministicPublicKey.cpp
teams2ua: Move it back

core/src/math/Base58.hpp
```diff
@@ -58,8 +60,18 @@ namespace ledger {
...
+        namespace Base58Utils {
+            template<typename HashAlgorithmFunction>
+            std::vector<uint8_t> computeChecksum(const std::vector<uint8_t>& bytes) {
+                auto doubleHash = HashAlgorithmFunction(HashAlgorithmFunction(bytes));
+                return std::vector<uint8_t>(doubleHash.begin(), doubleHash.begin() + 4);
+            }
 
-
+            template<typename HashAlgorithmFunction>
+            std::string encodeWithChecksum(const std::vector<uint8_t>& bytes, const std::string& digitis) {
```
gagbo: `digitDictionnary` or `digits` would be clearer than `digit_i_s` I think.

phaazon:
`HODL` for stalling PR.

PR #308, opened 2019-09-17
Adress Tezos pubkey curves issue.

phaazon:
Relates to LLC-347.


phaazon:
I’m also putting @gre because it will have implication on how they build Tezos wallet — i.e. you now **must** provide the type of curve you’re using. I could have used a default value but I hate default values as they always make things silent and hard to debug.

Normally, you should get the type of curve from the device and then you can just set the configuration with the `TEZOS_XPUB_CURVE` string. Bindings will have to be updated so that you can use things from `TezosConfigurationDefaults`.

gre:
i'm not sure we need to ever use SECP256K1 tho but i'm fine providing the default value.

it would be set as a string? 

config TEZOS_XPUB_CURVE -> "ED25519" string ?

phaazon:
@gre yes, but you should have a symbol available for that from the `TezosConfigurationDefaults` object.

PR #309, opened 2019-09-20
Fix *Wallet::type.

phaazon:


phaazon:
I’m not even sure whether your bugs @gre are related to that or not. 😆 I’m currently investigating.

PR #310, opened 2019-09-20
Fix double synchronization failure on Tezos

phaazon:
# Content

`std::bad_cast` due to a `soci` misusage of `row.get<_>` (`bool` is not accepted).

# Mandatory picture of mister animal


gre:
let's try to get this one merge today please. @KhalilBellakrid . so we can test many stuff after (like transactions) that and get back to you with potential feedbacks on what went well and what went not well. (val & I are off next week)

PR #311, opened 2019-09-23
Fix XRP database missing column when inserting transactions.

phaazon:
# Content

Bug found by @gre. May he have a cookie (thanks!).

# Animal portrait


gre:
🤛

PR #313, opened 2019-09-23
Normalize file names and header paths

alekece:
**DON'T BE A-FRAID !** 

I have been introduced a lot of changes but it's basically files renaming.

To make the naming in the code base consistent, I changed all file names that don't fit with **PascalCase** convention and replaced random ".h" extensions by ".hpp"
Moreover I allocated some times to remove all old header guards by `#pragma once` directive - this is supported by almost all compilers.
I also changed all relative header paths to the absolute ones as I precognized in the [draft](https://github.com/LedgerHQ/lib-ledger-core/blob/misc/coding-conventions/CONVENTIONS.md) convention.


teams2ua:
AFAIR, we use `*.hpp` for files that contains template function or classes as there are some implementation inside, for all other cases we use `*.h`.
[FAR] It would be nice to redo djinni as well to have `#pragma once` in generated code

phaazon:
For Djinni, I can take care of it.

phaazon:
By default, `djinni` outputs as `.h`.

alekece:
All capitalize renaming seems to be ignored... I'll fix that

phaazon:
Wait wait I have a fix incoming.

PR #321, opened 2019-09-30
Fix XRP: set Ledger Sequence for RippleLikeTransactionApi

KhalilBellakrid:
This is related to an issue faced by Vault during XRP integration.

### DIFF
--- a/core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
+++ b/core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
@@ -101,6 +101,12 @@ namespace ledger {
             _sender = RippleLikeAddress::fromBase58(tx.sender, _currency);
 
             _sequence = std::make_shared<api::BigIntImpl>(tx.sequence);
+
+            if (tx.block.hasValue()) {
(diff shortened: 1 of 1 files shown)


core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -101,6 +101,12 @@ namespace ledger {
             _sender = RippleLikeAddress::fromBase58(tx.sender, _currency);
 
             _sequence = std::make_shared<api::BigIntImpl>(tx.sequence);
+
+            if (tx.block.hasValue()) {
+                _ledgerSequence = std::make_shared<api::BigIntImpl>(BigInt(static_cast<unsigned long long>(tx.block.getValue().height)));
+            } else {
+                _ledgerSequence = std::make_shared<api::BigIntImpl>(BigInt::ZERO);
```
phaazon: I’m not sure what this is intended for. Can you add a comment to explain why we would like to have a transaction with the ledger sequence set to 0? Thanks! 🙂

onyb:
Tested **OK** with the wallet-daemon! ✅ 🚀 🎉

PR #325, opened 2019-10-01
Refactorize `OperationDatabaseHelper` class

alekece:
Before approve this PR, we have to have a talk about a comment I left in purpose in the source code.
There are one line I'm not proud of and I would like to change it.


ledger-core/inc/core/operation/OperationDatabaseHelper.inl
```diff
@@ -0,0 +1,184 @@
...
+#include <core/operation/Operation.hpp>
+#include <core/wallet/BlockDatabaseHelper.hpp>
+#include <core/wallet/TrustIndicator.hpp>
+
+namespace ledger {
+namespace core {
+
+namespace impl {
+
+    template <typename T, typename = void>
```
phaazon: Nice. Please add some comment about that trait so that people don’t have to read the whole series of declarations to get what it is.

ledger-core/inc/core/operation/OperationDatabaseHelper.inl
```diff
@@ -0,0 +1,193 @@
+/*
```
phaazon: I suggest you keep `.cpp` as file name. It contains implementation and I’m not sure we really care about naming files `.inl` just for when we have templated function implementations. Ping @LedgerHQ/core-library for other hindsight.

ledger-core/inc/core/operation/OperationDatabaseHelper.inl
```diff
@@ -0,0 +1,193 @@
...
+    const std::string& accountUid, const std::string& txId, const api::OperationType type)
+{
+    return SHA256::stringToHexHash(fmt::format("uid:{}+{}+{}", accountUid, txId, api::to_string(type)));
+}
+        
+template <typename Derived>
+void OperationDatabaseHelper<Derived>::queryOperations(
+    soci::session&, int32_t, int32_t, bool, bool, std::vector<Operation>&)
+{
+    // TODO: What we're supposed to do here ?
```
phaazon: That one is already empty so your `throw` is a good call.

ledger-core/inc/core/operation/OperationDatabaseHelper.inl
```diff
@@ -0,0 +1,193 @@
...
+#include <core/database/SociNumber.hpp>
+#include <core/database/SociDate.hpp>
+#include <core/database/SociOption.hpp>
+#include <core/collections/Strings.hpp>
+#include <core/operation/Operation.hpp>
+#include <core/wallet/BlockDatabaseHelper.hpp>
+#include <core/wallet/TrustIndicator.hpp>
+
+namespace ledger {
+namespace core {
```
phaazon: Please align / indent as we do everywhere else. :)

ledger-core/inc/core/operation/OperationDatabaseHelper.hpp
```diff
@@ -0,0 +1,200 @@
...
+                "OperationDatabaseHelper<T> requires 'T' to define 'updateCurrencyOperation' method");
+
+            using namespace soci;
+
+            auto count = 0;
+            std::string serializedTrust;
+            // TODO: I'm honestly not sure at all about this line below - since
+            // api::TrustIndicator has none `serialize` method, we obviously need the
+            // underlying implementation. Change `api::TrustIndicator` to
+            // `TrustIndicator` in the `Operation` class is for sure a better solution
```
phaazon: As we discussed live, it’d be better to move `TrustIndicator` in the modularization and change from `api::TrustIndicator` to `TrustIndicator` in `Operation` so that you don’t require that dynamic cast.

ledger-core/inc/core/operation/OperationDatabaseHelper.inl
```diff
@@ -0,0 +1,191 @@
...
+                return block.uid;
+            });
+            sql << "SELECT COUNT(*) FROM operations WHERE uid = :uid", use(operation.uid), into(count);
+            auto newOperation = count == 0;
+            if (!newOperation) {
+                sql << "UPDATE operations SET block_uid = :block_uid, trust = :trust WHERE uid = :uid"
+                        , use(blockUid)
+                        , use(serializedTrust)
+                        , use(operation.uid);
+                Derived::updateCurrencyOperation(sql, operation, newOperation);
```
teams2ua: Why should `OperationDatabaseHelper` depends on Derived?
Can the Derived do something like 
```
bool Derived::putOperation(soci::session& sql, const Operation& operation) {
  updateCurrencyOperation(sql, operation, OperationDatabaseHelper::putOperation(sql, operation));
}
```
?
This way there should be no dependencies for Derived in OperationDatabaseHelper, so it can be non-template

alekece: You make a point here ! In the current state, there are no problem with your solution because of the implementation of `putOperation` but there are some pitfalls I want to avoid.  

First I cannot guarantee than the `putOperation` definition will never change - we have to change all the derived classes if such a thing happens
Second I choose CRTP because all methods are statics and there are no way to tell to the derived class to implement/override this `putOperation` - this is error-proned and unclear. By taking a derived class as templated type, we mandatory the definition of a static method named `updateCurrencyOperation`.

teams2ua: > First I cannot guarantee than the putOperation definition will never change - we have to change all the derived classes if such a thing happens
If it changes, for some currencies it may not works correctly anymore.

> Second I choose CRTP because all methods are statics and there are no way to tell to the derived class to implement/override this putOperation - this is error-proned and unclear. By taking a derived class as templated type, we mandatory the definition of a static method named updateCurrencyOperation
CRTP is better then inheritance. But simple function composition is better then CRTP. Less dependencies, the better.

PR #326, opened 2019-10-02
Running integration tests and fixing them

KhalilBellakrid:
Fixed following tests: 
- `ledger-core-integration-BitcoinWalletDatabaseTests.PutTransaction`: not getting value of output because of account UID was not right,
- `ledger-core-integration-BitcoinWalletDatabaseTests.PutTransactionWithMultipleOutputs`: not getting value of output because of account UID was not right,
- `ledger-core-integration-EthereumLikeWalletSynchronization.XpubSynchronization`: not using the right account,
- `ledger-core-integration-EthereumMakeTransaction.ParseUnsignedRawTransaction`: additional character in raw tx,
- `ledger-core-integration-RippleMakeTransaction.ParseSignedRawTransactionWithDestinationTag`: Raw tx to lowerCase and comparing to LedgerSequence field instead of LastLedgerSequence field,
- `ledger-core-integration-TezosMakeTransaction.CreateTx`: Use the right curve and needs new explorer,
- `ledger-core-integration-TezosMakeTransaction.ParseSignedRawTransaction`: Use the right curve,
- `ledger-core-integration-TezosMakeTransaction.ParseSignedRawRevealTransaction`: Use the right curve,
- `ledger-core-integration-TezosMakeTransaction.ParseSignedRawOriginationTransaction`: Use the right curve,
- `ledger-core-integration-TezosLikeWalletSynchronization.MediumXpubSynchronization`: Parsing transactions,

For this one, I choose to have a workaround to fix ASAP tests: 
`ledger-core-integration-AccountCreationTest.ChangePassword`
The workaround consists in closing and reopening the db with only the new password`changePassword`, I enforced the test of `AccountCreationTest.ChangePassword` with a scenario `decryptedDB -> encryptedDB -> encryptedDBWithDifferentPassword -> decryptedDB`

This one is tricky, it's random crash after running the test, it seems to be related to `Qt` :
`ledger-core-integration-EthereumLikeWalletSynchronization.ReorgLastBlock (SEGFAULT)` 
It needs further investigation (SFOL ... or getting rid of `Qt`)


core/src/wallet/tezos/explorers/api/TezosLikeTransactionParser.cpp
```diff
@@ -183,6 +177,12 @@ namespace ledger {
                     }
                 } else if (_lastKey == "public_key") {
                     _transaction->publicKey = value;
+                } if (_lastKey == "amount") {
```
phaazon: That `if` will always be tested, even if a previous branch has succeeded. I think you need an `else if`.

core/test/integration/BaseFixture.cpp
```diff
@@ -128,7 +128,7 @@ void BaseFixture::TearDown() {
 std::shared_ptr<WalletPool> BaseFixture::newDefaultPool(const std::string &poolName, const std::string &password) {
     return WalletPool::newInstance(
             poolName,
-            "",
+            password,
```
phaazon: Nice catch.

core/src/wallet/tezos/explorers/api/TezosLikeTransactionParser.cpp
KhalilBellakrid: Clearly a typo ! Good catch ! 👍

phaazon:
I’m waiting for the CI and I’ll merge if it’s okay.

PR #328, opened 2019-10-02
Implement parsing of DestinationTag during Ripple account synchronization

onyb:
### What is this about?

We were not parsing `DestinationTag` from Ripple transactions during account sync, and so this information was not stored in the core DB. As a result, the destination tag in the wallet daemon was always `0`.

### DIFF
--- a/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
@@ -157,7 +157,9 @@ namespace ledger {
 
                         auto resultObj = json["result"].GetObject();
 
-                        if (resultObj.HasMember("engine_result") && resultObj["engine_result"] == "tesSUCCESS") {
+                        if (resultObj.HasMember("engine_result") &&
+                            (resultObj["engine_result"] == "tesSUCCESS") ||
+                             resultObj["engine_result"] == "terQUEUED")) {
                           // Check presence of tx_json field
                           if (!resultObj.HasMember("tx_json") || !resultObj["tx_json"].IsObject()) {
                             throw make_exception(api::ErrorCode::HTTP_ERROR, "Failed to broadcast transaction, no (or malformed) field \"tx_json\" in response");
@@ -171,6 +173,7 @@ namespace ledger {
 
                           return txnObj["hash"].GetString();
                         }
+
 
                         throw make_exception(api::ErrorCode::HTTP_ERROR,
                                              "Failed to broadcast transaction: {}",


core/src/wallet/ripple/explorers/RippleLikeBlockchainExplorer.h
```diff
@@ -62,25 +62,14 @@ namespace ledger {
...
             uint64_t confirmations;
+            Option<uint64_t> destinationTag;
+
             std::vector<api::RippleLikeMemo> memos;
 
             RippleLikeBlockchainExplorerTransaction() {
                 confirmations = 0;
             }
 
-            RippleLikeBlockchainExplorerTransaction(const RippleLikeBlockchainExplorerTransaction &cpy) {
```
phaazon: Why has this disappeared?

core/src/wallet/ripple/explorers/api/RippleLikeTransactionParser.cpp
```diff
@@ -131,10 +131,11 @@ namespace ledger {
                     _transaction->confirmations = value.toUint64();
                 } else if (_lastKey == "ledger_index") {
                     RippleLikeBlockchainExplorer::Block block;
-                    BigInt valueBigInt = BigInt::fromString(number);
-                    block.height = valueBigInt.toUint64();
+                    block.height = value.toUint64();
```
phaazon: That change is not linked to `DestinationTag`. Why is it needed?

core/src/wallet/ripple/explorers/RippleLikeBlockchainExplorer.h
onyb: I was pairing with @pollastri-pierre, and he thinks that we don't need this copy constructor.

core/src/wallet/ripple/explorers/api/RippleLikeTransactionParser.cpp
onyb: Just a small cleanup. We already had `BigInt value = BigInt::fromString(number);` declared outside the `if` block.

core/src/wallet/ripple/explorers/RippleLikeBlockchainExplorer.h
phaazon: Alright then. :)

core/src/wallet/ripple/api_impl/RippleLikeTransactionApi.cpp
```diff
@@ -102,6 +102,10 @@ namespace ledger {
 
             _sequence = std::make_shared<api::BigIntImpl>(tx.sequence);
 
+            if (tx.destinationTag.nonEmpty()) {
+              _destinationTag = (int64_t)tx.destinationTag.getValue();
```
alekece: I suggest here to use a proper cast (e.g `static_cast<int64_t>(tx.destinationTag.getValue())`)

onyb: Thanks for the suggestion. I didn't know this `static_cast` stuff. Updated my PR!

flocks:
\o/

PR #331, opened 2019-10-04
Add separate unit test for BigInt

teams2ua:


phaazon:
I think we need the unit tests from @teams2ua in that PR, might be useful! However, the copy ctor situation has been resolved in previous PRs — and there’re conflicting files, so this PR is unlikely to be merged in the current state.

What should we do @teams2ua?

phaazon:
`HODL` for stalling PR.

phaazon:
:crying_cat_face:

PR #332, opened 2019-10-07
Fix tezos

KhalilBellakrid:
This PR is related to : [LLC-360](https://ledgerhq.atlassian.net/browse/LLC-360), [LLC-313](https://ledgerhq.atlassian.net/browse/LLC-313), [LLC-361](https://ledgerhq.atlassian.net/browse/LLC-361), [LLC-357](https://ledgerhq.atlassian.net/browse/LLC-357) and [LLC-358](https://ledgerhq.atlassian.net/browse/LLC-358).

### DIFF
--- a/core/src/database/migrations.cpp
+++ b/core/src/database/migrations.cpp
@@ -586,7 +586,7 @@ namespace ledger {
                     "uid VARCHAR(255) NOT NULL PRIMARY KEY REFERENCES accounts(uid) ON DELETE CASCADE ON UPDATE CASCADE,"
                     "wallet_uid VARCHAR(255) NOT NULL REFERENCES wallets(uid) ON DELETE CASCADE ON UPDATE CASCADE,"
                     "idx INTEGER NOT NULL,"
-                    "address VARCHAR(255) NOT NULL"
+                    "public_key VARCHAR(255) NOT NULL"
                     ")";
 
             sql << "CREATE TABLE tezos_transactions("
--- a/core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
@@ -44,7 +44,8 @@ namespace ledger {
                 TezosLikeBlockchainExplorer(configuration, {api::Configuration::BLOCKCHAIN_EXPLORER_API_ENDPOINT}) {
             _http = http;
             _parameters = parameters;
-            _explorerVersion = configuration->getString(api::Configuration::BLOCKCHAIN_EXPLORER_VERSION).value_or("v3");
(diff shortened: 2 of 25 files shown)


core/src/database/migrations.cpp
```diff
@@ -586,7 +586,7 @@ namespace ledger {
                     "uid VARCHAR(255) NOT NULL PRIMARY KEY REFERENCES accounts(uid) ON DELETE CASCADE ON UPDATE CASCADE,"
                     "wallet_uid VARCHAR(255) NOT NULL REFERENCES wallets(uid) ON DELETE CASCADE ON UPDATE CASCADE,"
                     "idx INTEGER NOT NULL,"
-                    "address VARCHAR(255) NOT NULL"
+                    "public_key VARCHAR(255) NOT NULL"
```
phaazon: Wait, is it safe to modify an “old” transaction like that?

alekece: I'm interesting too. Should we add a new migration with that column renaming ?

KhalilBellakrid: I did it in purpose because nobody is using Tezos yet :trollface: so instead of adding a migration for that ...

core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
```diff
@@ -44,6 +44,7 @@ namespace ledger {
                 TezosLikeBlockchainExplorer(configuration, {api::Configuration::BLOCKCHAIN_EXPLORER_API_ENDPOINT}) {
             _http = http;
             _parameters = parameters;
+            _explorerVersion = configuration->getString(api::Configuration::BLOCKCHAIN_EXPLORER_VERSION).value_or("v3");
```
alekece: I suggest to store somewhere - configuration file, `constexpr` variable - this hardcoded value "v3"

core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
```diff
@@ -79,15 +80,23 @@ namespace ledger {
                             throw make_exception(api::ErrorCode::HTTP_ERROR,
                                                  "Failed to get fees from network, no (or malformed) field \"result\" in response");
                         }
-                        auto fees = json["fees"].GetString();
+                        std::string fees = json["fees"].GetString();
+                        // Sometimes network is sending 0 for fees
+                        if (fees == "0") {
+                            fees = "1420"; // Taken from some existing XTZ wallets
```
alekece: Same suggestion here

core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
KhalilBellakrid: Nice catch, I already did something for that 🤔 I will use it there

PR #333, opened 2019-10-07
Consider XRP txn submission with status terQUEUED as successful

onyb:
### What is this about?

We should not treat `terQUEUED` as a hard failure. The rippled server implements a local queue to hold transactions when the current ledger is on the verge of getting full. One way to bypass this queue is to pay an escalated fee called _open ledger cost_, but this escalation lasts hardly for a couple of seconds (according to my tests), therefore making the solution not useful for products like Vault where there is a significant delay between transaction creation and submission.

By considering a _queued_ transaction as successful, we optimistically update the libcore DB, and count on libcore to remove this transaction from the DB during a synchronization,  if it got rejected for some reason. Other services must make sure to mark these removed transactions as `DROPPED` (or something similar).

### DIFF
--- a/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
@@ -157,7 +157,9 @@ namespace ledger {
 
                         auto resultObj = json["result"].GetObject();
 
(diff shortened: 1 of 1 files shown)


phaazon:
So this changes the semantics of _pushing something on Core side_ as before we would say we failed if it’s queued. I don’t think it’s a problem (I think it’s actually okay to say we’ve successfully pushed if it’s queued since we don’t have to retry?).

onyb:
@phaazon Yes, that's the idea.

PR #340, opened 2019-10-09
XRP improvements

KhalilBellakrid:
- Use `server_state` instead of `server_info` mainly to get fields in drops instead of XRPs,
- Expose a method to check if an address has been activated or not.


onyb:
Is it possible to consider the load-factor while computing the fees? (see formula below)

```
Current Transaction Cost in Drops = (base_fee × load_factor) ÷ load_base
```

Unlike the _open ledger cost_ (a premium that you pay to get your txn in the current ledger sequence), we can have spikes in `load_factor` for extended periods of time.

onyb:
After discussing this directly with Khalil, we agreed to not consider the load factor, since it fluctuates a lot.

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -461,5 +461,28 @@ namespace ledger {
             });
         }
 
+        void RippleLikeAccount::isAddressActivated(const std::string &address,
+                                                   const std::shared_ptr<api::BoolCallback> &isActivated) {
+            isAddressActivated(address).callback(getContext(), isActivated);
+        }
+
+        Future<bool> RippleLikeAccount::isAddressActivated(const std::string &address) {
```
alekece: First thing first GJ here 
I have some suggestions to improve the readability of this method

core/test/integration/synchronization/ripple_synchronization.cpp
```diff
@@ -108,6 +108,8 @@ TEST_F(RippleLikeWalletSynchronization, MediumXpubSynchronization) {
             auto block = wait(account->getLastBlock());
             auto blockHash = block.blockHash;
 
+            EXPECT_EQ(wait(account->isAddressActivated("rageXHB6Q4VbvvWdTzKANwjeCT4HXFCKX7")), true);
+            EXPECT_EQ(wait(account->isAddressActivated("rageXHB6Q4VbvvWdTzKANwjeCT4HXFCK")), false);
```
teams2ua: This test only checks that address is valid, because it will fails the `parse` operation https://github.com/LedgerHQ/lib-ledger-core/pull/340/files#diff-212cd97f714ef22a58bc4f440cadb639R470
Could you please add the test when the address is valid but really not activated.

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -461,5 +461,28 @@ namespace ledger {
...
+            auto xrpAddress = RippleLikeAddress::parse(address, getWallet()->getCurrency());
+            if (!xrpAddress) {
+                return Future<bool>::successful(false);
+            }
+            auto self = getSelf();
+            return _explorer->getBaseReserve().flatMap<bool>(getContext(), [self, xrpAddress](const std::shared_ptr<BigInt> &reserve) -> Future<bool> {
+                return self->_explorer->getBalance(
+                        std::vector<RippleLikeKeychain::Address>{
+                                std::dynamic_pointer_cast<RippleLikeAddress>(xrpAddress)
+                        }).map<bool>(self->getContext(), [reserve](const std::shared_ptr<BigInt> &balance) -> bool {
```
alekece: Can you create some variables with proper name ?

```cpp
auto const isBalanceSufficient = [reserve](const std::shared_ptr<BigInt> &balance) -> bool {
   if (!reserve || !balance) {
      return false;
   }
    return reserve->compare(*balance) < 0;
}
auto const xrpAddresses = std::vector<RippleLikeKeychain::Address>{std::dynamic_pointer_cast<RippleLikeAddress(xrpAddress)}

return self->_explorer->getBalance(xrpAddresses).map<bool>(self->getContext(), isBalanceSufficient);
```

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -461,5 +461,28 @@ namespace ledger {
...
+                    if (!reserve || !balance) {
+                        return false;
+                    }
+                    return reserve->compare(*balance) < 0;
```
alekece: Is it possible to add the global `operator<` to compare `BigInt` ? Like that we don't manage anymore the result value.

core/test/integration/synchronization/ripple_synchronization.cpp
KhalilBellakrid: I skipped that case because I didn't have any address that is valid but nor activated 😢

core/src/wallet/ripple/RippleLikeAccount.cpp
KhalilBellakrid: Thanks ! Are we talking about your comment about variable creation ? or something else ?

core/test/integration/synchronization/ripple_synchronization.cpp
onyb: Your wish is my command: `rf1pjatD8LyyevP1BqQJtHoz5edC5vE77Q `

core/src/wallet/ripple/RippleLikeAccount.cpp
alekece: Exactly and the rework about `BigInt`

core/src/wallet/ripple/RippleLikeAccount.cpp
KhalilBellakrid: Are we talking about readability ? If so, then we should comply to that in every call with maps ... but if we do so it is going to be less readable when you chain `map` or `flatMap` so ...

phaazon: I think the code from @KhalilBellakrid is fine and what you suggested makes everything more verbose for nothing really interesting here. Chaining futures is a very common thing to do and especially, I think your version is weird because you declare the functions in reverse order of how they will be used while the inline version from @KhalilBellakrid feels more natural to me. And it’s more functional programming-like (because of the lambda), so it’s better because giving names to things that don’t require one is always something I want to stay away.

In Haskell, we have an adage that I think can be used in _any_ language: naming is hard, so if you don’t have to name something, you’re removing a problem, especially for temporaries.

core/src/wallet/ripple/RippleLikeAccount.cpp
phaazon: Yes, please create a Jira ticket; this has nothing to do with the current PR and it would make it’s merge slower.

core/test/integration/synchronization/ripple_synchronization.cpp
KhalilBellakrid: Thanks @onyb ! But never activate it ! Otherwise you break our tests ! :trollface:

core/src/wallet/ripple/RippleLikeAccount.cpp
```diff
@@ -461,5 +461,28 @@ namespace ledger {
...
+            if (!xrpAddress) {
+                return Future<bool>::successful(false);
+            }
+            auto self = getSelf();
+            return _explorer->getBaseReserve().flatMap<bool>(getContext(), [self, xrpAddress](const std::shared_ptr<BigInt> &reserve) -> Future<bool> {
+                return self->_explorer->getBalance(
+                        std::vector<RippleLikeKeychain::Address>{
+                                std::dynamic_pointer_cast<RippleLikeAddress>(xrpAddress)
+                        }).map<bool>(self->getContext(), [reserve](const std::shared_ptr<BigInt> &balance) -> bool {
+                    if (!reserve || !balance) {
```
phaazon: You can move the `!reserve` test before calling that future to prevent an I/O since you will not use the result.

core/test/integration/synchronization/ripple_synchronization.cpp
teams2ua: one from XRP documentation - rDTXLQ7ZKZVKz33zJbHjgVShjsBnqMBhmN (with private key generation, so I hope nobody will ever activate it, and if you see the test failing, quickly call me)

core/src/wallet/ripple/RippleLikeAccount.cpp
alekece: I disagree but not completely for our case. My previous comment should not be apply for all our lambdas obviously - I only think about several layers of nested lambdas. As @phaazon said, naming is hard but not only naming. Read some codes with nested lambdas containing some advanced logics are not something I enjoy a lot. Good naming and const lambda variables increase a lot the readability of piece of codes and expose clearly what the developer -  who has written the code - want to do. 
@KhalilBellakrid This is ok the way to do and the comment you left so I'll approve this PR

PR #342, opened 2019-10-09
Add migration for changing colum name of tezos_account column

KhalilBellakrid:
Thought that Live was not using Tezos version yet ... but it seems that it was the case ! 😢 😞


core/src/database/migrations.cpp
```diff
@@ -674,5 +674,13 @@ namespace ledger {
 
         template <> void rollback<14>(soci::session& sql) {
         }
+
+        template <> void migrate<15>(soci::session& sql) {
+            sql << "ALTER TABLE tezos_accounts RENAME COLUMN address TO public_key";
```
phaazon: Wait. If you don’t change the `public_key -> address` line 589… that PR is not even needed, right?

KhalilBellakrid: Initially it was address and it seems that Live is using already a version with XTZ in prod ... so ...

PR #344, opened 2019-10-09
added getter to Services

alekece:


ledger-core/src/core/wallet/AbstractWallet.cpp
```diff
@@ -71,6 +71,14 @@ namespace ledger {
             _logger = services->logger();
         }
 
+        std::shared_ptr<Services> AbstractWallet::getServices() const {
+            auto services = _services.lock();
+
+            if (!services)
+                throw Exception(api::ErrorCode::RUNTIME_ERROR, "Services is freed.");
```
phaazon: I’d say it’s wrong message (maybe it was never set) but I’ll merge because we shouldn’t face that error and if we do we know what’s wrong.

PR #345, opened 2019-10-09
Remove coin-related static method 

alekece:
These static methods should be implemented in coin library now

### DIFF
--- a/ledger-core/idl/address.djinni
+++ b/ledger-core/idl/address.djinni
@@ -11,17 +11,5 @@ Address = interface +c {
 
     # Get the currency associated with this address.
     getCurrency(): Currency;
-
-    # Attempt to parse a string address.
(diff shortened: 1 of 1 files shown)


ledger-core/idl/address.djinni
```diff
@@ -11,17 +11,5 @@ Address = interface +c {
 
     # Get the currency associated with this address.
     getCurrency(): Currency;
-
-    # Attempt to parse a string address.
```
phaazon: What does all this become?

alekece: As I said in the description, the static `parse` method - as well as `isValid` - calls underlying coin parse so all this become part of `ledger-core-xxx`.
If these two static methods are required in `ledger-core` library - so far I don't see any usage - we should rethink the way we manage its.

PR #348, opened 2019-10-12
Tezos: transaction serialization and parsing, signature and broadcast

KhalilBellakrid:
This is a PR that fixes last encountered issues while trying to broadcast a signed XTZ tx from `ledger-live` CLI, to sum up:

- Broadcasted raw signed transaction should not prefixed with `OPERATION_TAG_GENERIC`, only raw unsigned transaction should be (because why not :trollface:), there is a reference in the code where I figure out that,
- It seems that self-transactions are not allowed on XTZ (... hmm why not),
- During build of transaction we were setting counter of destination address not account's address (meh!),
- Finally, when broadcast is successful we were expecting a response of type `{"hash":"txHash"}`, but actually we got response of type `"txHash"`. 

So this changes were tested with `ledger-live-common@8.0.0-alpha.3` and I successfully managed to broadcast transactions from account `tz1iJuhk6v4A6wPbi1uLNy33ZuKu5JgHN1ye` 🚀 🚀 :
[opJchf5w9L12DThYBxv4NzidrLpSBaGsJqjM43hAuQ12R7ym258](https://tzscan.io/opJchf5w9L12DThYBxv4NzidrLpSBaGsJqjM43hAuQ12R7ym258?block_hash=BM1Ybxe7uSQaDUuFLey62aF6mmaC8c3EPTf2gRq7ChR84En5FJT),
[oouyitRd7puaFb3qCFB8QYTQ4JwbuqwxQc2v9Pq44xhXJbyo52N](https://tzscan.io/oouyitRd7puaFb3qCFB8QYTQ4JwbuqwxQc2v9Pq44xhXJbyo52N?block_hash=BL7iG9EHZJ6GRFB5LGZdN7YiNPFJ7rh95sUbetx5tqKcwAJhT5C),
[ooU8DwYymRr9VxhjaMLUPhFb7BzwX5G8RNvD4mfecQU2crYMsmn](https://tzscan.io/ooU8DwYymRr9VxhjaMLUPhFb7BzwX5G8RNvD4mfecQU2crYMsmn?block_hash=BMQ6ryFXjGRcUang9u2Gyp74AX58z54u7Zh5ah4KKWqjz8Gkb1A)


core/src/wallet/tezos/transaction_builders/TezosLikeTransactionBuilder.cpp
```diff
@@ -129,15 +134,18 @@ namespace ledger {
...
                                                          const std::vector<uint8_t> &rawTransaction,
                                                          bool isSigned) {
             auto params = currency.tezosLikeNetworkParameters.value();
             auto tx = std::make_shared<TezosLikeTransactionApi>(currency);
             BytesReader reader(rawTransaction);
-            // Watermark: Generic-Operation
-            reader.readNextByte();
+            if (!isSigned) {
+                // Watermark: Generic-Operation
+                reader.readNextByte();
```
phaazon: I think we’re going to need to make a huge pass of cleaning of our parsers because such a code can easily break libcore as we’re not ensuring that the read byte has the expected value.

PR #350, opened 2019-10-14
`Block` to `api::block`

alekece:
This PR contains also a small fix concerning a bad assignment.

4RZB2KR_700w_0


ledger-core/inc/core/explorers/LedgerApiParser.hpp
```diff
@@ -171,7 +171,7 @@ namespace ledger {
 
             void attach(const std::string& statusText, uint32_t statusCode) {
                 _statusCode = statusCode;
-                _statusText = _statusText;
+                _statusText = statusText;
```
phaazon: Yeah I get warnings about that one very time and I had planned to fix it. Thanks for doing it.

PR #355, opened 2019-10-14
Bitcoin Feature

KhalilBellakrid:
Support choosing change address path.

### DIFF
--- a/core/src/wallet/bitcoin/transaction_builders/BitcoinLikeUtxoPicker.cpp
+++ b/core/src/wallet/bitcoin/transaction_builders/BitcoinLikeUtxoPicker.cpp
@@ -119,8 +119,21 @@ namespace ledger {
 
             if (buddy->changeAmount > BigInt(_currency.bitcoinLikeNetworkParameters.value().DustAmount)) {
                 // TODO implement multi change
-                // TODO implement use specific change address
-                auto changeAddress = buddy->keychain->getFreshAddress(BitcoinLikeKeychain::CHANGE)->toString();
+                // If multiple change paths provided we take the last one
+                auto changeAddressIndex = Try<uint32_t>::from([&](){
+                    if (buddy->request.changePaths.empty()) {
+                        throw make_exception(api::ErrorCode::INVALID_ARGUMENT, "No change path provided by user.");
+                    }
+                    auto it = buddy->request.changePaths.back();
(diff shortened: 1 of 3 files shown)


core/src/wallet/bitcoin/transaction_builders/BitcoinLikeUtxoPicker.cpp
```diff
@@ -119,8 +119,21 @@ namespace ledger {
...
                 // TODO implement multi change
-                // TODO implement use specific change address
-                auto changeAddress = buddy->keychain->getFreshAddress(BitcoinLikeKeychain::CHANGE)->toString();
+                // If multiple change paths provided we take the last one
+                auto changeAddressIndex = Try<uint32_t>::from([&](){
+                    if (buddy->request.changePaths.empty()) {
+                        throw make_exception(api::ErrorCode::INVALID_ARGUMENT, "No change path provided by user.");
+                    }
+                    auto it = buddy->request.changePaths.back();
+                    return std::stoul(it.substr(it.find_last_of('/') + 1), nullptr, 0);
```
alekece: Which version of c++ we are using here ? Because you can remove the substring construction by using the new `std::string_view`

phaazon: It’s a C++17 feature. We can’t use it.

alekece: @phaazon Hence my question to @KhalilBellakrid ;)

KhalilBellakrid: We use C++14 so .. SFYL ! :trollface: 
Concerning the behaviour, yes this is exactly what I'm looking for, we want support all ways of setting paths, user can input `0/1/8`, `1/8` or `8` and we will always retrieve the right index, if someone inputs `8/` or some other invalid path then we would throw an `exception` on the `stoul` hence the encapsulation in the `Try` and in that case we would fall back on previous behaviour (getting a fresh change address)

phaazon:
@KhalilBellakrid you got two approvals, so you can merge — unless you wanna investigate those failing builds.

phaazon:
`HODL` for stalling PR.

PR #357, opened 2019-10-21
Tezos: use sql transactions during originated accounts sync

KhalilBellakrid:


core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -127,27 +127,32 @@ namespace ledger {
...
-                                }
-
-                                return getTxs(id + 1);
-                            }).recover(self->getContext(), [self] (const Exception& ex) -> Unit {
-                                throw ex;
-                            });
-                        };
-                        return getTxs(0);
+                        static std::function<Future<Unit> (std::shared_ptr<TezosLikeAccount>, size_t)> getTxs =
+                                [] (const std::shared_ptr<TezosLikeAccount> &account, size_t id) -> Future<Unit> {
```
alekece: The `id` - given as parameter - is really useful ? Because all I see is a "opaque" check in line 146 and you can retrieve it from `_originatedAccounts` before create a new account. May I wrong but the logic around the `id` parameter isn't clear.

PR #358, opened 2019-10-21
Tezos: add support for Tzstats explorer

KhalilBellakrid:
This is in case our Tzscan is down for a moment, by the time we fix potential issues caused by Babylon update.


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,257 @@
...
+            std::string addressesStr = addresses[0]->toString();
+            return getHelper(fmt::format("account/{}", addressesStr),
+                             "total_balance",
+                             std::unordered_map<std::string, std::string>{}
+            );
+        }
+
+        Future<std::shared_ptr<BigInt>>
+        ExternalTezosLikeBlockchainExplorer::getFees() {
+            bool parseNumbersAsString = true;
```
alekece: Can you set the variable as `const` here ?

core/idl/wallet/tezos/tezos_like_wallet.djinni
```diff
@@ -158,7 +158,7 @@ TezosConfiguration = interface +c {
 TezosConfigurationDefaults = interface +c {
 	const TEZOS_DEFAULT_API_ENDPOINT: string = "https://explorers.api.live.ledger.com";
 	const TEZOS_DEFAULT_API_VERSION: string = "v3";
-	const TEZOS_OBSERVER_NODE_ENDPOINT_S2: string = "https://s2.tezos.com";
+	const TZSTATS_API_ENDPOINT: string = "https://api.tzstats.com/explorer";
```
gre: does https://explorers.api.live.ledger.com points to this? because in current live logic, we will **always** set the endpoint to https://explorers.api.live.ledger.com

KhalilBellakrid: What ?! 
No it won't, when you set `BLOCKCHAIN_EXPLORER_ENGINE` to `TZSTATS_API` we will set the endpoint as we should, so you don't need to set anything else from your side !

phaazon:
In theory, that fix is not even needed if Live has the URL too.

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+        Future<String>
+        ExternalTezosLikeBlockchainExplorer::pushLedgerApiTransaction(const std::vector<uint8_t> &transaction) {
+            std::stringstream body;
+            body << '"' << hex::toString(transaction) << '"';
+            auto bodyString = body.str();
+            return _http->POST("/injection/operation?chain=main",
+                               std::vector<uint8_t>(bodyString.begin(), bodyString.end()),
+                               std::unordered_map<std::string, std::string>{},
+                               "https://mainnet.tezrpc.me")
```
alekece: Why the url isn't store the same way than the others ?

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,257 @@
...
+            if (offset > 0) {
+                params += fmt::format("?offset={}",offset);
+            }
+            using EitherTransactionsBulk = Either<Exception, std::shared_ptr<TransactionsBulk>>;
+            return _http->GET(fmt::format("account/{}/op", addresses[0]))
+                    .template json<TransactionsBulk, Exception>(
+                            LedgerApiParser<TransactionsBulk,
+                            TezosLikeTransactionsBulkParser>())
+                    .template mapPtr<TransactionsBulk>(getExplorerContext(),
+                                                           [=](const EitherTransactionsBulk &result) {
```
alekece: The copy-capture isn't necessary

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,257 @@
...
+            return pushLedgerApiTransaction(transaction);
+        }
+
+        FuturePtr<TezosLikeBlockchainExplorer::TransactionsBulk>
+        ExternalTezosLikeBlockchainExplorer::getTransactions(const std::vector<std::string> &addresses,
+                                                             Option<std::string> fromBlockHash,
+                                                             Option<void *> session) {
+            uint64_t offset = 0, limit = 100;
+            if (session.hasValue()) {
+                auto s = _sessions[*((std::string *)session.getValue())];
```
alekece: I suggest to use `reinterpret_cast` and store the value in a const variable

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,257 @@
...
+        std::string ExternalTezosLikeBlockchainExplorer::getExplorerVersion() const {
+            return "";
+        }
+
+        Future<std::shared_ptr<BigInt>>
+        ExternalTezosLikeBlockchainExplorer::getHelper(const std::string &url,
+                                                       const std::string &field,
+                                                       const std::unordered_map<std::string, std::string> &params,
+                                                       const std::string &fallbackValue) {
+            bool parseNumbersAsString = true;
```
alekece: DITTO

KhalilBellakrid:
> In theory, that fix is not even needed if Live has the URL too.

Nope it does not work because URL suffixes are not the same, nor the returned data so parsing is different

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
KhalilBellakrid: It could, it was just to avoid creating dozens of urls specially that it is the only place we use it !

PR #359, opened 2019-10-21
Tezos: patch code to support Babylon update

KhalilBellakrid:
This PR is meant to support Babylon update, thanks to the configuration we can switch between former protocol and the one with Babylon update. 
It is meant also to be easy to review so we can later on get rid of support of old protocol.

Removed HODL tag because I tested with `ledger-live` CLI and it's working 🚀 :
[opGvkEh34g7i3fEbErgdfePgy3F3grZHwZJoT9zuwf6xfHKEkfv](https://tzstats.com/operation/opGvkEh34g7i3fEbErgdfePgy3F3grZHwZJoT9zuwf6xfHKEkfv)

### DIFF
--- a/core/src/api/TezosConfiguration.cpp
+++ b/core/src/api/TezosConfiguration.cpp
@@ -7,4 +7,6 @@ namespace ledger { namespace core { namespace api {
 
 std::string const TezosConfiguration::TEZOS_XPUB_CURVE = {"TEZOS_XPUB_CURVE"};
 
+std::string const TezosConfiguration::TEZOS_PROTOCOL_UPDATE = {"TEZOS_PROTOCOL_UPDATE"};
+
 } } }  // namespace ledger::core::api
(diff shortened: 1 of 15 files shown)


core/src/api/TezosConfiguration.cpp
```diff
@@ -7,4 +7,6 @@ namespace ledger { namespace core { namespace api {
 
 std::string const TezosConfiguration::TEZOS_XPUB_CURVE = {"TEZOS_XPUB_CURVE"};
 
+std::string const TezosConfiguration::TEZOS_PROTOCOL_UPDATE = {"TEZOS_PROTOCOL_UPDATE"};
```
gre: do we need to send this constants or it's set as default?

KhalilBellakrid: For the moment, the default is the old protocol, so you need to inject that to the wallet's configuration, set `TEZOS_PROTOCOL_UPDATE ` to `TEZOS_PROTOCOL_UPDATE_BABYLON`

gre: cc @valpinkman . derivation.js can set custom param per derivation. maybe need same concept somewhere **per crypto**

PR #363, opened 2019-10-22
Segregate the database system across coins (ledger-core + XRP + ETH)

phaazon:
# Content

This PR migrates the split database migration-based system. Basically, each coin codebase (e.g. `ledger-core-ethereum`) doesn’t have to _register_ for a DB initialization, but it must call the _forward migration_ mechanism at some place. The idea is to do it automagically so that users don’t see it, and we don’t it when creating the first wallet for a given coin.

Sub-coin family wallet won’t do anything then as they share the same DB structure.

# Mandatory picture of animal


ledger-core-ethereum/src/ethereum/database/Migrations.cpp
```diff
@@ -2,9 +2,9 @@
 
 namespace ledger {
     namespace core {
-        int const EthereumMigration::coinID = 60;
```
alekece: Maybe we can put here some preprocessor macros here to check the C++ version - in case we upgrade our version later ? In C++17 we don't need this declaration anymore and more than that it's deprecated.

phaazon: We’re not using C++17 so why bother now? When we move to C++17 (if we ever do) that line will make a compile error and we’ll fix it.

PR #371, opened 2019-10-29
Tezos: fix operations on originated accounts

KhalilBellakrid:
This a patch to allow actions from originated account.
We rely on RPC call not to make serialization more complex than it is and also because those accounts are considered as legacy.
This PR was tested with Live and it's working : https://tzstats.com/KT1PTC8w6Mk6MSnx56TzMRx4Z8weLLq3oGgP

### DIFF
--- a/core/src/wallet/common/api_impl/OperationApi.cpp
+++ b/core/src/wallet/common/api_impl/OperationApi.cpp
@@ -30,14 +30,12 @@
  */
 #include "OperationApi.h"
 #include <wallet/common/Amount.h>
-#include "../AbstractAccount.hpp"
-#include <wallet/common/Amount.h>
+#include <wallet/common/AbstractAccount.hpp>
 #include <wallet/bitcoin/api_impl/BitcoinLikeOperation.h>
 #include <wallet/ethereum/api_impl/EthereumLikeOperation.h>
 #include <wallet/ripple/api_impl/RippleLikeOperation.h>
 #include <wallet/tezos/api_impl/TezosLikeOperation.h>
 #include <api/WalletType.hpp>
-
 
 namespace ledger {
     namespace core {
@@ -70,7 +68,7 @@ namespace ledger {
             return _backend.recipients;
         }
 
-        Operation &OperationApi::getBackend() {
+        ledger::core::Operation &OperationApi::getBackend() {
             return _backend;
         }
 
--- a/core/src/wallet/tezos/TezosLikeAccount2.cpp
+++ b/core/src/wallet/tezos/TezosLikeAccount2.cpp
@@ -208,7 +208,7 @@ namespace ledger {
                                       })
                     .recoverWith(getContext(), [] (const Exception &e) {
                         // Avoid fmt error because of JSON string
-                        auto error = "Failed to broadcast raw transaction, reason " + e.getMessage();
+                        const auto error = "Failed to broadcast raw transaction, reason " + e.getMessage();
                         return Future<std::string>::failure(
                                 Exception(api::ErrorCode::INCOMPLETE_TRANSACTION, error)
                         );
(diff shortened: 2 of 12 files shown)


core/src/wallet/common/api_impl/OperationApi.cpp
```diff
@@ -70,7 +68,7 @@ namespace ledger {
             return _backend.recipients;
         }
 
-        Operation &OperationApi::getBackend() {
+        ledger::core::Operation &OperationApi::getBackend() {
```
phaazon: We’re in the same namespace; why is that needed?

core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -206,11 +207,10 @@ namespace ledger {
                                           return seq.str();
                                       })
                     .recoverWith(getContext(), [] (const Exception &e) {
+                        // Avoid fmt error because of JSON string
+                        auto error = "Failed to broadcast raw transaction, reason " + e.getMessage();
```
alekece: Don't forget `const` qualifier

core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -206,11 +207,10 @@ namespace ledger {
...
                         return Future<std::string>::failure(
-                                make_exception(
-                                        api::ErrorCode::INCOMPLETE_TRANSACTION,
-                                        fmt::format("Failed to broadcast raw transaction, reason {}", e.getMessage())
-                                )
+                                Exception(api::ErrorCode::INCOMPLETE_TRANSACTION, error)
```
alekece: What is the advantage to use `make_exception` over `Exception` ?

core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -243,22 +244,31 @@ namespace ledger {
                 tx->setSender(accountAddress);
                 tx->setReceiver(TezosLikeAddress::fromBase58(request.toAddress, currency));
                 tx->setSigningPubKey(self->getKeychain()->getPublicKey().getValue());
+                tx->setManagerAddress(managerAddress);
                 tx->setType(request.type);
-                return explorer->getCounter(senderAddress).flatMapPtr<Block>(self->getContext(), [self, tx, explorer] (const std::shared_ptr<BigInt> &counter) {
+                auto counterAddress = protocolUpdate == api::TezosConfigurationDefaults::TEZOS_PROTOCOL_UPDATE_BABYLON ?
```
alekece: Ditto

core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.h
```diff
@@ -103,6 +102,10 @@ namespace ledger {
 
             TezosLikeTransactionApi & setBalance(const BigInt &balance);
 
+            TezosLikeTransactionApi & setManagerAddress(const std::string &managerAddress);
+            std::string getManagerAddress();
```
alekece: Ditto

core/src/wallet/tezos/TezosLikeAccount2.cpp
KhalilBellakrid: Not going through `fmt` which is crashing if you don't escape braces ! :trollface:

alekece: You cannot use here `make_exception(api::ErrorCode::INCOMPLETE_TRANSACTION, error)` ?

KhalilBellakrid: Please read my comment above (in the code too) and `make_exception` code ! 👌

core/src/wallet/common/api_impl/OperationApi.cpp
KhalilBellakrid: It was just a 'clean up' to comply with declaration ! (and we can't get rid of the one in declaration because of some namespaces clash 😞)

PR #373, opened 2019-10-29
Fixes on Tezos integration

KhalilBellakrid:
- Originated accounts: balance and transaction history were messed up after Babylon update, now it is possible to have an operation with same `hash`, `OperationType` and `OperationTag` between Implicit and Originated accounts,
- Send max: was never taken into account


core/src/wallet/tezos/explorers/api/TezosLikeTransactionParser.cpp
```diff
@@ -139,6 +139,8 @@ namespace ledger {
                     _transaction->gas_limit = toValue(number, false);
                 } else if (_lastKey == "storage_limit") {
```
alekece: I suggest to put all these raw strings to `constexpr` variables somewhere. Even if these values are used just one time, I think any potential future changes will be faster with that kind of constant values.

phaazon: You would have to change each variable’s name. I think it’s fine here.

PR #376, opened 2019-10-31
Fix tezos: fees calculation in some edge cases

KhalilBellakrid:
- Fix fees when performing a send max and a reveal at same time,
- Add burned XTZs to fees when sending to unrevealed account (storage limit is considered as mXTZ to burn),
- Fix parser for Tzstats to catch delegators addresses.

### DIFF
--- a/core/src/wallet/tezos/TezosLikeAccount2.cpp
+++ b/core/src/wallet/tezos/TezosLikeAccount2.cpp
@@ -249,14 +249,26 @@ namespace ledger {
                                     ->getString(api::TezosConfiguration::TEZOS_PROTOCOL_UPDATE)
                                     .value_or("");
 
+                            auto tx = std::make_shared<TezosLikeTransactionApi>(currency, protocolUpdate);
+
+                            // Check whether we need a reveal operation
+                            // We assume that accounts are synced
+                            soci::session sql(self->getWallet()->getDatabase()->getPool());
+                            // Check if there is a send operation, otherwise we need a reveal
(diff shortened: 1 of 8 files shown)


core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -243,65 +243,82 @@ namespace ledger {
...
-                                    return explorer->getBalance(listAddresses).mapPtr<api::TezosLikeTransaction>(self->getContext(), [tx] (const std::shared_ptr<BigInt> &balance) {
+                            // Check if recepient is revealed or not
+                            // because if not we have to add additional fees equal to storage_limit (in mXTZ)
+                            auto getRevealFee = [self, explorer, request]() -> Future<BigInt> {
+                               if (request.type != api::TezosOperationTag::OPERATION_TAG_TRANSACTION) {
+                                   return Future<BigInt>::successful(BigInt::ZERO);
+                               }
+                                return explorer->getManagerKey(request.toAddress).map<BigInt>(self->getContext(), [request] (const std::string &managerKey) -> BigInt {
+                                    // Base unit is uXTZ
+                                    return !managerKey.empty() ? BigInt::ZERO : *request.storageLimit * BigInt("1000");
```
teams2ua: [OPT] BigInt(1000);? Why string?

phaazon: That was actually a good question I would have preferred to see resolved before merging. :( @pollastri-pierre

PR #381, opened 2019-11-08
Tezos Fix: revelation of sending account

KhalilBellakrid:
- Check `manager_key` instead of relying on number of outgoing ops from an address to check revelation, because it seems that an account could be revealed, send transactions but still unrevealed (... black magic),
- Introduce concept of `unallocated` accounts to know if we should burn XTZs or not when sending tokens,
- Avoid check of revelation on `KT` accounts, since no revelation is required for KT accounts since Babylon update.

### DIFF
--- a/core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.h
+++ b/core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.h
@@ -144,6 +144,12 @@ namespace ledger {
                                                      const std::shared_ptr<HttpClient> &http,
                                                      const std::string &rpcNode);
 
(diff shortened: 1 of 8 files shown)


core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.h
```diff
@@ -144,6 +144,12 @@ namespace ledger {
                                                      const std::shared_ptr<HttpClient> &http,
                                                      const std::string &rpcNode);
 
+            virtual Future<bool> isAllocated(const std::string &address) = 0;
+            static Future<bool> isAllocated(const std::string &address,
```
phaazon: Oh I don’t like that overloaded method… :D

PR #391, opened 2019-11-19
Fix ETH negative balances.

phaazon:
# Content

This fix basicaly removes fees when converting internal transactions ->
operations. Taking them into account will make duplicates, as they are
already paid by the ETH transactions that generated them, or by the
contract (case of NONE type), and in this case, we don’t care about
fees.

# Mandatory picture of dog maneers


phaazon:
So after some discussion internally, the negative balances _might_ be already fixed, but I think this fix is needed whatsoever because I think we got lucky. Lucky because the fees, I guess, were set to 0, so it didn’t change much. I need investigate tomorrow to get what’s going on but there’s something fishy here.

PR #398, opened 2019-11-22
Review execution context usage

KhalilBellakrid:
Better usage of execution contexts.
**TODO:** 
- Additional Tests: I've already tested and now no task is blocked, but we should perform more QA in case we missed some blocking tasks,

### DIFF
--- a/core/src/wallet/common/AbstractAccount.cpp
+++ b/core/src/wallet/common/AbstractAccount.cpp
@@ -42,12 +42,7 @@ namespace ledger {
     namespace core {
 
         AbstractAccount::AbstractAccount(const std::shared_ptr<AbstractWallet> &wallet, int32_t index)
-                :
-                DedicatedContext(
-                        wallet->getPool()
-                                ->getDispatcher()
-                                ->getThreadPoolExecutionContext(fmt::format("account_{}_{}", wallet->getName(), index))
-                )
+                : DedicatedContext(wallet->getMainExecutionContext())
         {
             _uid = AccountDatabaseHelper::createAccountUid(wallet->getWalletUid(), index);
             _logger = wallet->logger();
@@ -150,7 +145,7 @@ namespace ledger {
             return std::make_shared<OperationQuery>(
                     api::QueryFilter::accountEq(getAccountUid()),
                     getWallet()->getDatabase(),
-                    getContext(),
+                    getWallet()->getPool()->getThreadPoolExecutionContext(),
                     getMainExecutionContext()
             );
         }
@@ -233,6 +228,5 @@ namespace ledger {
         void AbstractAccount::eraseDataSince(const std::chrono::system_clock::time_point & date, const std::shared_ptr<api::ErrorCodeCallback> & callback) {
             eraseDataSince(date).callback(getMainExecutionContext(), callback);
         }
-
     }
 }
--- a/core/src/wallet/common/AbstractAccount.cpp
+++ b/core/src/wallet/common/AbstractAccount.cpp
@@ -36,12 +36,18 @@
 #include <api/ErrorCode.hpp>
 #include <events/Event.hpp>
 #include <wallet/common/database/BlockDatabaseHelper.h>
-
+#include <wallet/pool/WalletPool.hpp>
 namespace ledger {
     namespace core {
 
         AbstractAccount::AbstractAccount(const std::shared_ptr<AbstractWallet> &wallet, int32_t index)
-                : DedicatedContext(wallet->getMainExecutionContext()) {
+                :
+                DedicatedContext(
+                        wallet->getPool()
+                                ->getDispatcher()
+                                ->getThreadPoolExecutionContext(fmt::format("account_{}_{}", wallet->getName(), index))
+                )
+        {
             _uid = AccountDatabaseHelper::createAccountUid(wallet->getWalletUid(), index);
             _logger = wallet->logger();
             _index = index;
--- a/core/src/wallet/common/AbstractAccount.cpp
+++ b/core/src/wallet/common/AbstractAccount.cpp
@@ -37,6 +37,7 @@
 #include <events/Event.hpp>
 #include <wallet/common/database/BlockDatabaseHelper.h>
 #include <wallet/pool/WalletPool.hpp>
+
 namespace ledger {
     namespace core {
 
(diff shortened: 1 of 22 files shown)


phaazon:
As @RomanWlm needs the `getWallet` fix today, I think I’m just going to create a smaller PR for him to quickly be able to use it (that PR does more things and as you stated some work is still left to do, among testing). I hope we can merge that soon though!

core/src/wallet/pool/WalletPool.cpp
```diff
@@ -294,6 +294,12 @@ namespace ledger {
         FuturePtr<AbstractWallet> WalletPool::getWallet(const std::string &name) {
             auto self = shared_from_this();
             return async<std::shared_ptr<AbstractWallet>>([=] () {
+                auto it = self->_wallets.find(WalletDatabaseEntry::createWalletUid(self->getName(), name));
```
phaazon: This is going to conflict with the fix I wrote yesterday and landed this morning.

core/src/wallet/ethereum/api_impl/EthereumLikeOperation.h
```diff
@@ -47,6 +47,8 @@ namespace ledger {
         private:
             std::shared_ptr<api::EthereumLikeTransaction> _transaction;
             std::vector<std::shared_ptr<api::InternalTransaction>> _internalTxs;
+            std::shared_ptr<OperationApi> _backend;
+            bool _initialized;
```
phaazon: That’s a convoluted name. What’s that for?

core/src/wallet/common/AbstractAccount.cpp
```diff
@@ -36,12 +36,18 @@
 #include <api/ErrorCode.hpp>
 #include <events/Event.hpp>
 #include <wallet/common/database/BlockDatabaseHelper.h>
-
+#include <wallet/pool/WalletPool.hpp>
```
phaazon: Please put a whitespace after this include. Compacted code is harder to read.

core/src/wallet/ethereum/api_impl/EthereumLikeOperation.cpp
```diff
@@ -68,6 +54,25 @@ namespace ledger {
 
         std::vector<std::shared_ptr<api::InternalTransaction>>
         EthereumLikeOperation::getInternalTransactions() {
+            if (_initialized) {
```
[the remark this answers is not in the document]
KhalilBellakrid: Haha true, I completely forgot to set it to true and should be tested as false there.

core/src/wallet/ethereum/api_impl/EthereumLikeOperation.h
KhalilBellakrid: For lazy loading of internal transactions

phaazon: LGTM, even though I think that comment should be in the code.

core/src/wallet/ethereum/EthereumLikeAccount.cpp
```diff
@@ -480,43 +481,42 @@ namespace ledger {
...
+                if (result.isSuccess()) {
+                    code = api::EventCode::SYNCHRONIZATION_SUCCEED;
+                } else {
+                    code = api::EventCode::SYNCHRONIZATION_FAILED;
+                    payload->putString(api::Account::EV_SYNC_ERROR_CODE, api::to_string(result.getFailure().getErrorCode()));
+                    payload->putInt(api::Account::EV_SYNC_ERROR_CODE_INT, (int32_t)result.getFailure().getErrorCode());
+                    payload->putString(api::Account::EV_SYNC_ERROR_MESSAGE, result.getFailure().getMessage());
+                }
+                eventPublisher->postSticky(std::make_shared<Event>(code, payload), 0);
+                std::lock_guard<std::mutex> lock(self->_synchronizationLock);
```
alekece: As I said in the PR about TTL, there is probably a dead lock here. Please use a recursive mutex at least or find a solution to avoid the clashing

core/src/wallet/ethereum/api_impl/EthereumLikeOperation.h
alekece: Agreed. Such a variable name could be clear at a time and completely obscur later

core/src/wallet/common/AbstractWallet.cpp
```diff
@@ -48,11 +48,10 @@ namespace ledger {
                                        const api::Currency &currency,
                                        const std::shared_ptr<WalletPool> &pool,
                                        const std::shared_ptr<DynamicObject> &configuration,
-                                       const DerivationScheme &derivationScheme
-        )
-                : DedicatedContext(
-                pool->getDispatcher()->getSerialExecutionContext(fmt::format("wallet_{}", walletName))),
-                  _scheme(derivationScheme) {
+                                       const DerivationScheme &derivationScheme)
+                : DedicatedContext(pool->getDispatcher()->getThreadPoolExecutionContext(fmt::format("wallet_{}", walletName))),
```
phaazon: Do we really need a new EC? Why not using the main thread pool EC and let it dispatch parallelism for us?

core/src/wallet/common/AbstractWallet.cpp
```diff
@@ -241,15 +240,14 @@ namespace ledger {
         }
 
         FuturePtr<api::Account> AbstractWallet::getAccount(int32_t index) {
+            auto it = _accounts.find(index);
```
phaazon: Ah, this is a caching fix! Can you explain in the PR that you also fix that, pretty please? :)

core/src/wallet/common/AbstractWallet.cpp
KhalilBellakrid: Nice catch, I forgot to remove this one !

core/src/wallet/ethereum/EthereumLikeAccount.cpp
KhalilBellakrid: So first, the only thing I changed about that code is the execution context in which we complete future, then I reformatted the code.
And by the way, I don't think there will be an issue since we call `onComplete` on a different thread.

core/src/wallet/ethereum/api_impl/EthereumLikeOperation.h
KhalilBellakrid: I renamed it so it will be clearer !

core/src/wallet/ethereum/EthereumLikeAccount.cpp
KhalilBellakrid: @alekece so `synchronize` will be called in the "global" execution context of the host, then when the future is completed, the rest of the code will be called on a thread of the execution context of the account.

PR #400, opened 2019-11-24
Basic caching

KhalilBellakrid:
Implementation of a simple cache to keep some values that are queried multiple times, here balance and last block.

### DIFF
--- a/core/src/utils/TTLCache.h
+++ b/core/src/utils/TTLCache.h
@@ -73,6 +73,11 @@ namespace ledger {
                         });
             };
 
+            void erase(const K &key) {
+                std::lock_guard<std::mutex> lock(_lock);
+                _cache.erase(key);
+            };
+
         private:
             Duration _ttl;
             std::unordered_map<K, std::pair<V, Duration>> _cache;
--- a/core/src/utils/TTLCache.h
+++ b/core/src/utils/TTLCache.h
@@ -29,8 +29,7 @@
  */
 
 
-#ifndef LEDGER_CORE_TTLCACHE_H
-#define LEDGER_CORE_TTLCACHE_H
+#pragma once
 
 #include <chrono>
 #include <unordered_map>
@@ -50,12 +49,11 @@ namespace ledger {
             Option<V> get(const K &key) {
                 std::lock_guard<std::mutex> lock(_lock);
                 auto it = _cache.find(key);
-                if (it == _cache.end() ||
-                        std::chrono::duration_cast<Duration>(std::chrono::steady_clock::now().time_since_epoch()) - (*it).second.second > _ttl) {
-                    if (it != _cache.end()) {
-                        // Clean the cache
-                        _cache.erase(key);
-                    }
+                if (it == _cache.end()) {
+                    return Option<V>();
+                } else if (getNowDurationSinceEpoch() - (*it).second.second > _ttl) {
+                    // Clean the cache
+                    _cache.erase(key);
                     return Option<V>();
                 }
                 return Option<V>((*it).second.first);
@@ -66,9 +64,7 @@ namespace ledger {
                 _cache.insert(
                         { key,
                           { value,
-                            std::chrono::duration_cast<Duration>(
-                                    std::chrono::steady_clock::now().time_since_epoch()
-                            )
+                            getNowDurationSinceEpoch()
                           }
                         });
             };
@@ -79,6 +75,9 @@ namespace ledger {
             };
 
         private:
+            Duration getNowDurationSinceEpoch() {
+                return std::chrono::duration_cast<Duration>(std::chrono::steady_clock::now().time_since_epoch());
+            }
             Duration _ttl;
             std::unordered_map<K, std::pair<V, Duration>> _cache;
             std::mutex _lock;
@@ -86,6 +85,3 @@ namespace ledger {
 
     }
 }
-
-
-#endif //LEDGER_CORE_TTLCACHE_H
--- a/core/src/utils/TTLCache.h
+++ b/core/src/utils/TTLCache.h
@@ -1,0 +1,86 @@
+/*
+ *
+ * TTLCache
+ *
+ * Created by El Khalil Bellakrid on 22/11/2019.
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019 Ledger
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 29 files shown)


core/src/wallet/common/AbstractWallet.cpp
```diff
@@ -48,11 +48,11 @@ namespace ledger {
...
                                        const std::shared_ptr<DynamicObject> &configuration,
-                                       const DerivationScheme &derivationScheme
-        )
-                : DedicatedContext(
-                pool->getDispatcher()->getSerialExecutionContext(fmt::format("wallet_{}", walletName))),
-                  _scheme(derivationScheme) {
+                                       const DerivationScheme &derivationScheme)
+                : DedicatedContext(pool->getDispatcher()->getThreadPoolExecutionContext(fmt::format("wallet_{}", walletName))),
+                  _scheme(derivationScheme),
+                  _balanceCache(std::chrono::seconds(configuration->getInt(api::Configuration::TTL_BLOCK_CACHE).value_or(30)))
```
alekece: I'm wandering if it's a good idea to put a fallback value to something which must be defined :
1. the default value `30` can be far away from the configuration value then introduce weird/unwanted/tricky behaviour
2. missing configuration value should be considered as technical bug and fix as soon as possible

core/src/wallet/ethereum/EthereumLikeAccount.cpp
```diff
@@ -480,43 +488,42 @@ namespace ledger {
...
+                if (result.isSuccess()) {
+                    code = api::EventCode::SYNCHRONIZATION_SUCCEED;
+                } else {
+                    code = api::EventCode::SYNCHRONIZATION_FAILED;
+                    payload->putString(api::Account::EV_SYNC_ERROR_CODE, api::to_string(result.getFailure().getErrorCode()));
+                    payload->putInt(api::Account::EV_SYNC_ERROR_CODE_INT, (int32_t)result.getFailure().getErrorCode());
+                    payload->putString(api::Account::EV_SYNC_ERROR_MESSAGE, result.getFailure().getMessage());
+                }
+                eventPublisher->postSticky(std::make_shared<Event>(code, payload), 0);
+                std::lock_guard<std::mutex> lock(self->_synchronizationLock);
```
alekece: **IMPORTANT**
I found a potential dead lock in that portion of code that lead to never assign `_currentSyncEventBus` to `nullptr`. 
A mutex is locked at the very beginning of `synchronize` method then locked in the future so in the future has been completed before the end of the synchronisation, the deadlock occurs. 
To fix that I suggest to use `std::recursive_mutex` to avoid to modify the current code and allow the correct behavior

core/src/utils/TTLCache.h
```diff
@@ -0,0 +1,86 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+
+#ifndef LEDGER_CORE_TTLCACHE_H
```
alekece: Since the segregation use now `#pragma once` it could be great to now use this convention in new header file

core/src/utils/TTLCache.h
```diff
@@ -0,0 +1,86 @@
...
+        template <typename K, typename V, typename Duration = std::chrono::seconds>
+        class TTLCache {
+        public:
+            TTLCache(const Duration &ttl) : _ttl(ttl) {};
+
+            Option<V> get(const K &key) {
+                std::lock_guard<std::mutex> lock(_lock);
+                auto it = _cache.find(key);
+                if (it == _cache.end() ||
+                        std::chrono::duration_cast<Duration>(std::chrono::steady_clock::now().time_since_epoch()) - (*it).second.second > _ttl) {
```
alekece: Maybe you can create a protected/private method to provide the current duration since epoch ?
`std::chrono::duration_cast<Duration>(std::chrono::steady_clock::now().time_since_epoch())`

The code will be more readable and BTW you will remove redundant code (i.e line 69)

core/src/utils/TTLCache.h
```diff
@@ -0,0 +1,86 @@
...
+                    if (it != _cache.end()) {
```
alekece: This is an open suggestion so I let you decide if you will apply it or not.

I am not a big fan to that double check on `it` so I suggest to to replace it by something like :
```cpp
if (it == _cache.end()) {
   return Option<V>();
}
else if (std::chrono::duration_cast<Duration>(std::chrono::steady_clock::now().time_since_epoch()) - (*it).second.second > _ttl) {
   _cache.erase(key);
  return Option<V>();
}
```

Even if the two `return` look redundant we can now return different values depending of the user case  (imagine an extra parameter telling us to valid the outdated cache for a new period of time).

core/src/wallet/common/AbstractWallet.cpp
KhalilBellakrid: We should always provide default values when there is a rationale behind it, otherwise I agree we should not set any and make it a mandatory config, in that case if it's missing we throw an exception for that ! 
Here 30sec seems reasonable for both balance and blocks since it is inferior to the frequency of blocks mining

core/src/wallet/ethereum/EthereumLikeAccount.cpp
KhalilBellakrid: Already discussed in previous PR

core/src/utils/TTLCache.h
KhalilBellakrid: Good catch I will do so !

core/src/utils/TTLCache.h
KhalilBellakrid: Hmm this is a good thought indeed, I was hesitating between redundant `return` or the double check on `it` but you are right, this can let us more freedom about how to make this code easily evolve

core/src/wallet/common/AbstractWallet.hpp
```diff
@@ -162,7 +167,7 @@ namespace ledger {
             DerivationScheme _scheme;
             std::weak_ptr<WalletPool> _pool;
             std::unordered_map<int32_t, std::shared_ptr<AbstractAccount>> _accounts;
-
+            TTLCache<std::string, Amount> _balanceCache;
```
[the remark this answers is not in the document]
KhalilBellakrid: It is overkilled to have a whole cache for one account, but it makes sense to have wallet handling multiple accounts and keeping a cache of their balances !

PR #401, opened 2019-11-26
Fix remaning issues with Tezos 

KhalilBellakrid:
- Fix send from KT when tz parent account in unrevealed: The issue was that we should have prepend the revelation instead of appending it to the send operation.
This was tested with latest CLI of Live: https://tzstats.com/oora48f1jbqPsR6AMe7vM8SopgR9gWwby4aeBQoDGt6zLfY5MFB 🚀 
- Fix check of unallocated accounts: we use to rely on `/chains/main/blocks/head/context/contracts/<contract_id>` which is the right way to do it, but since XTZ nodes are buggy on that same endpoint, I implement a workaround, which is checking if a the account has a tx history,
- Set storage to `0` for reveal operations

### DIFF
--- a/core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
+++ b/core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
@@ -256,7 +256,9 @@ namespace ledger {
             writer.writeByteArray(zarith::zSerializeNumber(bigIntGasLimit.toByteArray()));
 
             // Storage Limit
-            writer.writeByteArray(zarith::zSerializeNumber(_storage->toByteArray()));
+            // No sotrage for reveal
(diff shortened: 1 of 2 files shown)


core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
```diff
@@ -250,7 +256,9 @@ namespace ledger {
             writer.writeByteArray(zarith::zSerializeNumber(bigIntGasLimit.toByteArray()));
 
             // Storage Limit
-            writer.writeByteArray(zarith::zSerializeNumber(_storage->toByteArray()));
+            // No sotrage for reveal
```
alekece: I found a typo here: "sotrage" -> "storage"

phaazon: storage :d

PR #406, opened 2019-11-28
Tezos fees fixes

KhalilBellakrid:
So this PR is fixing tweaking fees:
- `KT` accounts and send max: remove fees from max spendable amount when sending from KT, since fees are paid by the parent `tz` account (interaction with smart contract),
- After some feedbacks from Live:

1. it seems that `2500` was low as default fees, so I bumped it to `5000` (from what I observed from some other XTZ wallets),
2. The nodes are really fluctuating between fees of values that are crazy high, so to avoid that we fix a threshold of `6*default_fees` i.e. `30000` (again from other XTZ wallets).

### DIFF
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -88,9 +88,7 @@ namespace ledger {
                         }
                         // Since nodes are giving some awkward values, we set a threshold to avoid having really fees
                         // Factor for threshold is inspired from other XTZ wallets
-                        return std::stoi(fees) > 6 * std::stoi(api::TezosConfigurationDefaults::TEZOS_DEFAULT_FEES) ?
-                        std::make_shared<BigInt>(6 * std::stoi(api::TezosConfigurationDefaults::TEZOS_DEFAULT_FEES)) :
-                        std::make_shared<BigInt>(fees);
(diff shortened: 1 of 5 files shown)


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -86,7 +86,11 @@ namespace ledger {
                         } else if (fees.find('.') != std::string::npos) {
                             fees = api::BigInt::fromDecimalString(fees, 6, ".")->toString(10);
                         }
-                        return std::make_shared<BigInt>(fees);
+                        // Since nodes are giving some awkward values, we set a threshold to avoid having really fees
+                        // Factor for threshold is inspired from other XTZ wallets
+                        return std::stoi(fees) > 6 * std::stoi(api::TezosConfigurationDefaults::TEZOS_DEFAULT_FEES) ?
```
alekece: Can you stored the calculation `6 * std::atoi(api::TezosConfigurationDefaults::TEZOS_DEFAULT_FEES)` to avoid unwanted computation and to give a name to that magic value ?

gre: i kinda agree, to me this should be a constant as well. because the DEFAULT_FEES is already arbitrary and it would be unwanted that a change of default fees affect the "max cap". because the minimum cap can vary independently of the max cap

gre:
hope this get merged soon 🙏  kudos to the fix

PR #414, opened 2019-12-06
Add support for PostgreSQL through soci

KhalilBellakrid:
**TODO later:** 
- Activate PostgreSQL encryption.


core/src/database/PostgreSQLBackend.cpp
```diff
@@ -0,0 +1,80 @@
...
+            session.open(*soci::factory_postgresql(), _dbName);
+#else
+            throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Libcore should be compiled with PG_SUPPORT flag.");
+#endif
+        }
+
+        void PostgreSQLBackend::changePassword(const std::string & oldPassword,
+                                            const std::string & newPassword,
+                                            soci::session &session) {
+            throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Database should be initiated before changing password.");
```
alekece: In my opinion the error message isn't correct since the exception can be raised even if the database is initiated

core/test/integration/synchronization/synchronization_tests.cpp
```diff
@@ -34,14 +34,23 @@
...
 #include <api/KeychainEngines.hpp>
+#include <api/PoolConfiguration.hpp>
 class BitcoinLikeWalletSynchronization : public BaseFixture {
 
 };
 
 TEST_F(BitcoinLikeWalletSynchronization, MediumXpubSynchronization) {
+    auto configuration = DynamicObject::newInstance();
+#ifdef PG_SUPPORT
+    bool usePostgreSQL = true;
```
alekece: add const qualifier here

core/src/database/PostgreSQLBackend.cpp
KhalilBellakrid: Yes I forgot to set more accurate message, nice catch !

.circleci/setup_macos.sh
```diff
@@ -30,7 +30,7 @@ brew install cmake
 
 if [ "$BUILD_CONFIG" == "Debug" ]; then
 	echo "========> Install Qt5"
```
gagbo: Install Qt5 and Postgres

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-## 3.2.1 (on-going)
+## 3.3.0 (on-going)
+
+- Add support for PostgreSQL (tested with PostgreSQLv12.1), for more details about how to
```
gagbo: Seems also tested with pg9.6 on Debian at least 
https://github.com/LedgerHQ/lib-ledger-core/blob/1611c04baba0fb5760c2cb0bdf3cc7457bbe0584/.circleci/setup_debian.sh#L34

core/src/database/DatabaseBackend.cpp
```diff
@@ -40,6 +41,10 @@ namespace ledger {
             return std::make_shared<SQLite3Backend>();
         }
 
+        std::shared_ptr<api::DatabaseBackend> api::DatabaseBackend::getPostgreSQLBackend(int32_t connectionPoolSize) {
```
phaazon: I think we need a way to get the `PostgreSQLBackend` directly. The problem with this code is that even though it’s fine for interface code (they will use the `api::DatabaseBackend` API) because upcast are free in C++, it’s not the case whenever we need the concrete type, resulting in a downcast and hence a dynamic cast.

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -39,7 +39,8 @@ namespace ledger {
             const std::shared_ptr<api::PathResolver> &resolver,
             const std::shared_ptr<spdlog::logger>& logger,
             const std::string &dbName,
-            const std::string &password
+            const std::string &password,
+            bool usingPostgreSQLDatabase
```
phaazon: The backend is already provided as argument… that `bool` is suspicious to me in terms of API.

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -69,10 +72,11 @@ namespace ledger {
                                             const std::shared_ptr<api::PathResolver> &resolver,
                                             const std::shared_ptr<spdlog::logger> &logger,
                                             const std::string &dbName,
-                                            const std::string &password) {
-            return FuturePtr<DatabaseSessionPool>::async(context, [backend, resolver, dbName, logger, password] () {
+                                            const std::string &password,
+                                            bool usingPostgreSQL) {
```
phaazon: Ditto.

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -88,7 +92,7 @@ namespace ledger {
             int version = getDatabaseMigrationVersion(sql);
 
             soci::transaction tr(sql);
-            migrate<CURRENT_DATABASE_SCHEME_VERSION>(sql, version);
+            migrate<CURRENT_DATABASE_SCHEME_VERSION>(sql, version, _usingPostgreSQL);
```
phaazon: As @alekece said, I also think we need a better representation of backends. Booleans don’t fit well here (booleans actually never fit well anything most of the type; it’s better to use typed enums).

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -97,7 +101,7 @@ namespace ledger {
             int version = getDatabaseMigrationVersion(sql);
 
             soci::transaction tr(sql);
-            rollback<CURRENT_DATABASE_SCHEME_VERSION>(sql, version);
+            rollback<CURRENT_DATABASE_SCHEME_VERSION>(sql, version, _usingPostgreSQL);
```
phaazon: Ditto.

core/src/database/PostgreSQLBackend.cpp
```diff
@@ -0,0 +1,80 @@
...
+                                  const std::string &password,
+                                  soci::session &session) {
+            _dbName = dbName;
+            setPassword(password, session);
+        }
+
+        void PostgreSQLBackend::setPassword(const std::string &password,
+                                         soci::session &session) {
+            if (_dbName.empty()) {
+                throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Database should be initiated before setting password.");
```
phaazon: I think an exception here is very violent. `setPassword` should return something stating an error occurred.

core/src/database/PostgreSQLBackend.cpp
```diff
@@ -0,0 +1,80 @@
...
+            }
+            session.close();
+#ifdef PG_SUPPORT
```
phaazon: I don’t get why you need this. That file should be part of a compilation unit that is conditionally checked against `PG_SUPPORT` so that you don’t have to make that kind of test inside your code.

core/src/database/PostgreSQLBackend.cpp
```diff
@@ -0,0 +1,80 @@
...
+            session.open(*soci::factory_postgresql(), _dbName);
+#else
+            throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Libcore should be compiled with PG_SUPPORT flag.");
+#endif
+        }
+
+        void PostgreSQLBackend::changePassword(const std::string & oldPassword,
+                                            const std::string & newPassword,
+                                            soci::session &session) {
+            throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Change of password is not handled with PostgreSQL backend.");
```
phaazon: Maybe you can use the `UNIMPLEMENTED` type error instead.

.circleci/build_lib.sh
```diff
@@ -146,6 +146,28 @@ if [ "$1" == "ios" ]; then
     sudo xcode-select --reset
 fi
 
+if [ "$BUILD_CONFIG" == "Debug" ]; then
+	echo "======> Create PostgreSQL with name test_db, host localhost and port 5432"
+	export POSTGRES_USER=postgres
+	export POSTGRES_DB=test_db
+	if [ "$unamestr" == "Linux" ]; then
+		echo "======> Modify PostgreSQL configuration file ..."
```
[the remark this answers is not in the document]
KhalilBellakrid: It's just allowing to connect to PostgreSQL without any authentication ! I will add a comment on that

.circleci/build_node_module.sh
```diff
@@ -5,6 +5,11 @@ export NVM_DIR="${XDG_CONFIG_HOME/:-$HOME/.}nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 source ~/.bashrc
 
+# Issue for MacOS: dyld: Library not loaded: /usr/local/opt/icu4c/lib/libicui18n.62.dylib
+if [ $(uname) == "Darwin" ]; then
+	brew upgrade
```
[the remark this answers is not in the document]
KhalilBellakrid: I agree on limiting the upgrade to node only, but unfortunately the issue seems to be a mismatch in dependencies installed by `brew` and we don't know the whole list so that's why we upgrade everything. For the answer about why it's working, I could not find the exact reason and solution was suggested here and worked for me: https://stackoverflow.com/questions/53828891/dyld-library-not-loaded-usr-local-opt-icu4c-lib-libicui18n-62-dylib-error-run
Since it's really specific to `node` I'm not spending more time on it.

.circleci/build_node_module.sh
```diff
@@ -14,6 +19,8 @@ node --version
 echo "=====>Build node module"
 
 cd ledger-core-samples/nodejs
+# Speed up module's build
+export JOBS=8
```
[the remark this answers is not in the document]
KhalilBellakrid: It's for `node-gyp` itself !

.circleci/setup_debian.sh
```diff
@@ -31,7 +31,7 @@ export PATH=$HOME/cmake-3.12.3/bin:$PATH
 
 if [ "$BUILD_CONFIG" == "Debug" ]; then
     echo "========> Install Qt5"
-    apt-get install -y qt5-default libqt5websockets5 libqt5websockets5-dev
+    apt-get install -y qt5-default libqt5websockets5 libqt5websockets5-dev postgresql-9.6 libpq-dev postgresql-server-dev-all
```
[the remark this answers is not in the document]
KhalilBellakrid: I will move it to another line ! 👌

CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ set(CMAKE_MACOSX_RPATH 1)
 
 add_definitions("-DSQLITE_HAS_CODEC")
 
+# To compile with PostgreSQL support
+if (PG_SUPPORT)
+    add_definitions("-DPG_SUPPORT")
```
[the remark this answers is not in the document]
KhalilBellakrid: That's the purpose of using a specific name for this macro, since it's affecting how the whole project is configured and compiling it should appear at top level (same as option `BUILD_TESTS`)

core/idl/database/database.djinni
```diff
@@ -239,6 +239,10 @@ DatabaseBackend = interface +c {
     # @return DatabaseBackend object
     static getSqlite3Backend(): DatabaseBackend;
 
+    # Create an instance of PostgreSQL database.
+    # @return DatabaseBackend object
+    static getPostgreSQLBackend(connectionPoolSize: i32): DatabaseBackend;
```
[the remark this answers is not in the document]
KhalilBellakrid: This is made only for PostgreSQL, there is nothing to be future-proof about, other backends are created through `createBackendFromEngine`

core/lib/soci_postgresql/CMakeLists.txt
```diff
@@ -9,43 +9,53 @@
...
-if(SOCI_POSTGRESQL_NOBINDBYNAME)
-message("X")
-  add_definitions(-DSOCI_POSTGRESQL_NOBINDBYNAME=1)
-endif()
-
-if(SOCI_POSTGRESQL_NOPREPARE)
-message("Y")
-  add_definitions(-DSOCI_POSTGRESQL_NOPREPARE=1)
-endif()
+#include(CMakeDependentOption)
```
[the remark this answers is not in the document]
KhalilBellakrid: It was there before and I forgot to clean, nice catch ! 👍

core/src/database/PostgreSQLBackend.cpp
KhalilBellakrid: I kept same behaviour as before, if we want to change the whole method then it should be part of a dedicated PR.

core/src/database/PostgreSQLBackend.cpp
KhalilBellakrid: This file is always part of compilation, even if I exclude it from compilation on the `CMakeLists.txt` level, I'll have to do same thing with `PG_SUPPORT` on `DatabaseBackend` side, so in the end it introduces more complications than it solves ! 
WDYT ?

core/src/database/DatabaseSessionPool.cpp
KhalilBellakrid: I wanted to be part of a more general conversation, but I introduced a new enum `DatabaseBackendType` to handle that !

.circleci/build_lib.sh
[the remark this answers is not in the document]
phaazon: Perfect, thanks!

core/idl/database/database.djinni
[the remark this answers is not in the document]
phaazon: Okay, fair enough.

core/src/database/PostgreSQLBackend.cpp
phaazon: Ah, okay, right. I think I will open a Jira ticket about that because C++ is not really friendly with error handling the _right_ way (i.e. not exception), and I’m very versed into it because of my Haskell and Rust background. Maybe that’s something that should be done when migrating to Rust. I don’t really know.

core/src/database/PostgreSQLBackend.cpp
phaazon: I mean, you’re in a file `PostgreSQLBackend`, so that file should always be compiled with `PG_SUPPORT`. Compiling that file without `PG_SUPPORT` doesn’t make sense (or maybe it does but then the filename should be something else).

My point is:

- Instead of always compiling that file and check for `PG_SUPPORT` inside the file…
- Compile the file only when `PG_SUPORT` is set and then you have zero assumption to make inside the implementation of the file.

phaazon: However, maybe that is a bit boring to do or you don’t want to (it would mean removing GLOB patterns in a `CMakeFiles.txt` and add the source conditionally). I’ll let you decide what you prefer to do, even though I think what I describe is a bit cleaner.

core/src/database/migrations.cpp
```diff
@@ -369,20 +369,27 @@ namespace ledger {
...
 
-            // ETH operations
-            sql << "DROP TABLE ethereum_operations";
-
-            // ETH transactions
-            sql << "DROP TABLE ethereum_transactions";
-
+            // ETH transactions and operations
+            switch (type) {
+                case api::DatabaseBackendType ::POSTGRESQL: {
```
phaazon: Extra space between `DatabaseBackendType` and `::POSTGRESQL`. 🙂

core/src/database/migrations.cpp
```diff
@@ -534,7 +552,18 @@ namespace ledger {
              sql << "INSERT INTO eth_swap "
                 "SELECT transaction_uid, hash, nonce, value, block_uid, time, sender, receiver, input_data, gas_price, gas_limit, gas_used, confirmations, status "
                 "FROM ethereum_transactions";
-            sql << "DROP TABLE ethereum_transactions";
+
+            switch (type) {
+                case api::DatabaseBackendType ::POSTGRESQL: {
```
phaazon: Ditto.

core/src/database/PostgreSQLBackend.cpp
KhalilBellakrid: I got your previous point, but even by compiling that only with `PG_SUPPORT` you'll have to do the check on `DatabaseBackend` because we include `PostgreSQLBackend` (that was my previous comment)

PR #421, opened 2019-12-11
Fix memory leak in SECP256k1Point::generatorMultiply.

phaazon:
# Content

Relates to LLC-434.

# Mandatory picture of WHAT THE F- IS THAT?!

### DIFF
--- a/core/src/crypto/SECP256k1Point.cpp
+++ b/core/src/crypto/SECP256k1Point.cpp
@@ -87,6 +87,8 @@ namespace ledger {
             auto flag = secp256k1_ec_pubkey_tweak_add(_context, pubKey, num.data());
             if (flag == 0) throw Exception(api::ErrorCode::RUNTIME_ERROR, "SECP256k1Point SECP256k1Point::generatorMultiply(const std::vector<uint8_t> &n) failed");
             secp256k1_ec_pubkey_serialize(_context, serializedKey.data(), &len, pubKey, SECP256K1_EC_COMPRESSED);
+            delete pubKey;
+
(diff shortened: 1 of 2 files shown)


core/src/crypto/SECP256k1Point.cpp
```diff
@@ -87,6 +87,8 @@ namespace ledger {
             auto flag = secp256k1_ec_pubkey_tweak_add(_context, pubKey, num.data());
             if (flag == 0) throw Exception(api::ErrorCode::RUNTIME_ERROR, "SECP256k1Point SECP256k1Point::generatorMultiply(const std::vector<uint8_t> &n) failed");
```
teams2ua: We also need to delete this key before throwing.

gagbo: Why not using a unique_ptr to get free cleanup ?

teams2ua: I think we even can have the object on stack.

phaazon: `unique_ptr` has issues here because of opaque types; we would need to provide custom deleters and I think the scope of that function is small enough.

@teams2ua we cannot put them on the stack because they’re opaque types.

phaazon:
> I found `destroy()` function as a not very clean concept.
> The objects of SECP256k1Point class should be in valid state or destructed. With destroy function the set of possible states is bigger, so it is more error prone.
> I didn't expect you are going to change the context behavior in this PR. But if you want to change this I prefer to merge #307 instead.

PR #426, opened 2019-12-17
Prevent BitcoinLike UTXO picker from using tx<Dust

gagbo:
Prevent picker from using UTXO when the amount is less than dustAmount

Small issue with local tests, so using CI for now.


gagbo:
HODL, I finally built my tests and it fails :(

gagbo:
It's fine now. I had to make a few changes to BitcoinMakeBaseTransaction test class to make it work though.

phaazon:
Hello,

Can we merge this? :) ping @gagbo @alekece

gagbo:
> Hello,
> 
> Can we merge this? :) ping @gagbo @alekece

HODL status was because I didn't have a good test, I fixed it since so I'm confident about merging this (porting to design/modularization might take a little time depending on how test fixtures need to be migrated)

phaazon:
I think we need to merge this. @YBadiss has already found a case that would have used this! :)

PR #436, opened 2019-12-27
Tezos: fix recursive sync on originated accounts

KhalilBellakrid:
Basically we were not passing the session from a call to next recursion.
This is related to : https://ledgerhq.atlassian.net/browse/LLC-472

### DIFF
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -137,12 +137,17 @@ namespace ledger {
 
         FuturePtr<TezosLikeBlockchainExplorer::TransactionsBulk>
         ExternalTezosLikeBlockchainExplorer::getTransactions(const std::vector<std::string> &addresses,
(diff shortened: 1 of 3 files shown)


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -137,12 +137,17 @@ namespace ledger {
 
         FuturePtr<TezosLikeBlockchainExplorer::TransactionsBulk>
         ExternalTezosLikeBlockchainExplorer::getTransactions(const std::vector<std::string> &addresses,
-                                                             Option<std::string> fromBlockHash,
+                                                             Option<std::string> offset,
```
phaazon: `offset` is not really a good name here. What offset?

KhalilBellakrid: It's the `offset` related to pagination, our explorers paginate according to block hashes and tzstats do it according to an `limit` and `offset` that you append as params to your query.
And BTW this is pretty straightforward when you look at the code of `getTransactions`.

PR #437, opened 2019-12-27
Tezos: avoid creation of originated accounts based on failing origination ops

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-437

### DIFF
--- a/core/src/wallet/tezos/TezosLikeAccount.cpp
+++ b/core/src/wallet/tezos/TezosLikeAccount.cpp
@@ -143,7 +143,7 @@ namespace ledger {
                 if (OperationDatabaseHelper::putOperation(sql, operation)) {
                     emitNewOperationEvent(operation);
                 }
-                if (transaction.type == api::TezosOperationTag::OPERATION_TAG_ORIGINATION) {
+                if (transaction.type == api::TezosOperationTag::OPERATION_TAG_ORIGINATION && transaction.status == 1) {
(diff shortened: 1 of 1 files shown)


core/src/wallet/tezos/TezosLikeAccount.cpp
```diff
@@ -143,7 +143,7 @@ namespace ledger {
                 if (OperationDatabaseHelper::putOperation(sql, operation)) {
                     emitNewOperationEvent(operation);
                 }
-                if (transaction.type == api::TezosOperationTag::OPERATION_TAG_ORIGINATION) {
+                if (transaction.type == api::TezosOperationTag::OPERATION_TAG_ORIGINATION && transaction.status == 1) {
```
teams2ua: Do we have enum for this magic number?

phaazon: Yeah I raised that issue in some previous PRs, but we don’t currently. That’s on Tezos fault: they don’t provide semantics associated with the values if I recall correctly. :(

We could type it though.

KhalilBellakrid: Same, but this would be impacting all the currencies if we want to do it properly, so I created a ticket for that : https://ledgerhq.atlassian.net/browse/LLC-488

PR #443, opened 2020-01-07
Ripple test segregation

phaazon:
# Content

This PR provides both unit testing and integration testing for the `ledger-core-ripple` segregated project. Some code had to be modified to pass all tests — most of it is backported code from `develop` / `master`.

# Mandatory picture of…


ledger-core-ripple/src/ripple/keychains/RippleLikeKeychain.cpp
```diff
@@ -168,29 +168,39 @@ namespace ledger {
...
                 if (_address.empty()) {
-                    auto p = getDerivationScheme().getSchemeFrom(DerivationSchemeLevel::NODE).shift(1)
-                            .setAccountIndex(getAccountIndex())
-                            .setCoinType(getCurrency().bip44CoinType)
-                            .setNode(0)
-                            .setAddressIndex(0).getPath().toString();
-                    auto xpub = _xpub;
+                    auto nodeScheme = getDerivationScheme()
+                        .getSchemeFrom(DerivationSchemeLevel::NODE);
+                    auto p = nodeScheme.getPath().getDepth() > 0 ? nodeScheme
```
alekece: maybe a comment will be appropriate here

ledger-core-ripple/test/Common.cpp
```diff
@@ -0,0 +1,6 @@
+#include "Common.hpp"
+
+namespace ledger {
+    namespace core {
+    }
+}
```
alekece: What is it for ?

phaazon: The .hpp is used by several integration tests, but good call, that .cpp can be removed.

ledger-core-ripple/src/ripple/keychains/RippleLikeKeychain.cpp
phaazon: I agree 100%, but since I wasn’t the one writing the code, I have no idea what to write as comment. 😕

PR #455, opened 2020-01-15
Support BTC BEv3

KhalilBellakrid:
2 changes compared to BEv2:
1. value of outputs in a string,
2. transactions returned by `/transactions` endpoint is returning 1 object instead of 1 array of 1 object.

### DIFF
--- a/core/src/wallet/bitcoin/explorers/api/TransactionsParser.hpp
+++ b/core/src/wallet/bitcoin/explorers/api/TransactionsParser.hpp
@@ -46,7 +46,8 @@ namespace ledger {
 
             bool StartObject() {
                 _objectDepth += 1;
-
+                // In v2 /transactions/${hash} endpoint returns an array of tx object => _arrayDepth == 0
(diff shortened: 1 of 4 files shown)


core/src/wallet/bitcoin/explorers/api/TransactionsParser.hpp
```diff
@@ -44,6 +44,18 @@ namespace ledger {
                 _objectDepth = 0;
             }
 
+            bool StartObject() {
+                _objectDepth += 1;
+
+                if ((_arrayDepth == 1 || _arrayDepth == 0) && _objectDepth == 1) {
+                    BitcoinLikeBlockchainExplorerTransaction transaction;
+                    _transactions->push_back(transaction);
+                    getTransactionParser().init(&_transactions->back());
```
alekece: This line seems really opaque to me.

 Can you explain it or comment it ?

KhalilBellakrid: Done !!

PR #457, opened 2020-01-16
Paginate properly UTXOs

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-486

### DIFF
--- a/core/src/wallet/bitcoin/database/BitcoinLikeUTXODatabaseHelper.cpp
+++ b/core/src/wallet/bitcoin/database/BitcoinLikeUTXODatabaseHelper.cpp
@@ -69,8 +69,7 @@ namespace ledger {
 
             for (auto& row : rows) {
                 if (row.get_indicator(0) != i_null && filter(row.get<std::string>(0))) {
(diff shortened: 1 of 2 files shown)


core/src/wallet/bitcoin/database/BitcoinLikeUTXODatabaseHelper.cpp
```diff
@@ -63,31 +63,25 @@ namespace ledger {
...
-                        output.index = get_number<uint64_t>(row, 1);
-                        output.transactionHash = row.get<std::string>(2);
-                        output.value = row.get<BigInt>(3);
-                        output.script = row.get<std::string>(4);
-                        if (row.get_indicator(5) != i_null) {
-                            output.blockHeight = row.get<BigInt>(5).toUint64();
-                        }
-                        c += 1;
+                    out.resize(out.size() + 1);
+                    auto& output = out[out.size() - 1];
```
alekece: Well we already talked about that weird way to add a new element in a vector during the segregation.

Can just create the element in place and push back to that vector ?

PR #458, opened 2020-01-16
LTC: support of bech32 addresses

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-324

**NOTE:** HODL label will be removed once : https://github.com/LedgerHQ/lib-ledger-core/pull/451 is merged !


core/src/bitcoin/bech32/Bech32Factory.cpp
```diff
@@ -35,7 +35,8 @@
 namespace ledger {
     namespace core {
         Option<std::shared_ptr<Bech32>> Bech32Factory::newBech32Instance(const std::string &networkIdentifier) {
-            if (networkIdentifier == "btc" || networkIdentifier == "btc_testnet") {
+            auto btcBech32Identifiers = std::vector<std::string>{"btc", "btc_testnet", "dgb", "ltc"};
+            if (std::find(btcBech32Identifiers.begin(), btcBech32Identifiers.end(), networkIdentifier) != btcBech32Identifiers.end()) {
```
alekece: Please add `const` qualifier.

I have a other suggestion here it's to replace `std::vector` by `std::string[]` then use `std::cbegin(btcBech32Identifiers)` and `std::cend(btcBech32Identifiers`. 
With that modification we remove the runtime overhead to construct the `std::vector`.

core/src/bitcoin/bech32/Bech32Parameters.cpp
```diff
@@ -73,6 +73,30 @@ namespace ledger {
...
+                    static const Bech32Struct LITECOIN = {
+                            "litecoin",
+                            "ltc",
+                            "1",
+                            6,
+                            {0x3b6a57b2ULL, 0x26508e6dULL, 0x1ea119faULL, 0x3d4233ddULL, 0x2a1462b3ULL},
+                            {0x00},
+                            {0x00}
+                    };
+                    return LITECOIN;
```
alekece: Since the `if/else` statements grow up day after day, we can consider to put all these `Bech32Struct` to a `constexpr std::unordored_map<Bech32Struct>` and then compute a find operation on it.

core/src/bitcoin/bech32/Bech32Factory.cpp
KhalilBellakrid: ACK for the const qualifier, for the second part this is definitely to be part of the enhancements your working it, because we have many places where we could have to same optim (better to isolate in another PR)

core/src/bitcoin/bech32/Bech32Parameters.cpp
KhalilBellakrid: Same, this should be in a dedicated PR because we have multiple places where we do that, so let's open a ticket for this !

PR #459, opened 2020-01-17
Inherit DerivationPath from its API interface

alekece:
In our legacy codebase we can found several definition and implementation of `DerivationPath`.

The purpose of this PR is to uniform that class and to expose a proper interface.  
We use signed integer type to represent index because of djinni limitation.

## Mandatory picture of a majestic cat
majestic_cat

### DIFF
--- a/ledger-core/inc/core/utils/DerivationPath.hpp
+++ b/ledger-core/inc/core/utils/DerivationPath.hpp
@@ -34,35 +34,42 @@
 #include <string>
 #include <vector>
 
+#include <core/api/DerivationPath.hpp>
 #include <core/utils/Exception.hpp>
 
 namespace ledger {
     namespace core {
-        class DerivationPath {
+        class DerivationPath : public api::DerivationPath {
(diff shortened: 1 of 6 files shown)


ledger-core/inc/core/utils/DerivationPath.hpp
```diff
@@ -34,35 +34,42 @@
...
 #include <core/utils/Exception.hpp>
 
 namespace ledger {
     namespace core {
-        class DerivationPath {
+        class DerivationPath : public api::DerivationPath {
         public:
             explicit DerivationPath(const std::string& path);
-            explicit DerivationPath(const std::vector<uint32_t>& path);
+            explicit DerivationPath(const std::vector<int32_t>& path);
```
phaazon: Even though I get why you need to change the type, it’s a regression here, because we would authorize `m/44'/-1'/0` for instance.

The fix you need prior to this is to add the support of unsigned integrals to Djinni.

PR #460, opened 2020-01-17
Fix Ripple transaction dates

phaazon:
Relates to LLC-496, BACK-423.


phaazon:
# About dates…

Before:

Screenshot 2020-01-17 at 17 15 52

After:

Screenshot 2020-01-17 at 18 24 26

phaazon:
I’m switching the PR scope to dates only as the remaining fix is much larger.

PR #464, opened 2020-01-22
Segregate ETH tests

phaazon:
# Content

This PR brings the ETH tests to the modularized branch. It also contains some required fixes on `ledger-core` (isolated in separate commits).

# Mandatory picture of winky winky


ledger-core-ethereum/inc/ethereum/operations/EthereumLikeOperationQuery.hpp
```diff
@@ -28,6 +28,8 @@
  *
  */
 
+#pragma once
```
phaazon: That one was tricky to find in the first place! :)

ledger-core-ethereum/src/ethereum/ERC20/ERC20LikeOperationQuery.cpp
```diff
@@ -53,5 +52,18 @@ namespace ledger {
                 .outerJoin("erc20_operations AS e", "o.uid = e.ethereum_operation_uid")
                 .execute(sql);
         }
+
+        void ERC20LikeOperationQuery::inflateCompleteTransaction(
+            soci::session &sql,
+            const std::string &accountUid,
+            ERC20LikeOperation& operation
+        ) {
```
phaazon: So, why is this left empty, you might ask. It’s left empty because `performExecute` is overriden to return the right data. It’s a special case of tokens and we ought to change the whole design of the `OperationQuery` interface later.

ledger-core-ethereum/src/ethereum/EthereumLikeTransaction.cpp
```diff
@@ -66,7 +66,7 @@ namespace ledger {
 
             _gasPrice = std::make_shared<Amount>(_currency, 0, tx.gasPrice);
             _gasLimit = std::make_shared<Amount>(_currency, 0, tx.gasLimit);
-            _gasUsed = std::make_shared<Amount>(_currency, 0, tx.gasUsed.getValue());
+            _gasUsed = std::make_shared<Amount>(_currency, 0, tx.gasUsed.getValueOr(BigInt::ZERO));
```
phaazon: So this fix is required because now we use constructors instead of dummy initializing objects (like with `Foo{}`).

ledger-core-ethereum/inc/ethereum/ERC20/ERC20LikeOperation.hpp
```diff
@@ -139,27 +150,26 @@ namespace ledger {
...
             std::shared_ptr<api::BigInt> _gasLimit;
             std::shared_ptr<api::BigInt> _gasUsed;
-            std::string _sender;
-            std::string _receiver;
-            std::shared_ptr<api::BigInt> _value;
             std::vector<uint8_t> _data;
-            std::chrono::system_clock::time_point _time;
-            api::OperationType _operationType;
             int32_t _status;
-            int64_t _blockHeight;
```
alekece: Why all these members are deleted ?

ledger-core-ethereum/test/SynchronizationTests.cpp
```diff
@@ -0,0 +1,464 @@
...
+
+        auto wallet = wait(walletStore->getWallet(walletName));
+        auto account = std::dynamic_pointer_cast<EthereumLikeAccount>(wait(wallet->getAccount(0)));
+        EXPECT_EQ(account->getERC20Accounts().size(), erc20Count);
+
+        wait(services->freshResetAll());
+    }
+}
+
+//TEST_F(EthereumLikeWalletSynchronization, BalanceHistory) {
```
alekece: Why all the below tests are commented ?

ledger-core-ethereum/inc/ethereum/ERC20/ERC20LikeOperation.hpp
phaazon: As discussed out loud, those come from `Operation` (the common type), since we have to inherit from it! :)

ledger-core-ethereum/test/SynchronizationTests.cpp
phaazon: Because I’ve forgot about them. Dammit. Good catch :D :D :D

phaazon: _c’était pour voir si tu suivais_ 😆

ledger-core-ethereum/inc/ethereum/operations/EthereumLikeOperationQuery.hpp
alekece: Meh. Dunno what happen in my life during that segregation.. But obviously bad things ^^

phaazon:
I’m putting this on `HODL` as I have sevral commits needed to completely implement the rest of the sync tests.

PR #465, opened 2020-01-22
Segregate BTC fixes

alekece:
This goal of this PR is to backport all the BTC fixes from develop to our modularization branch.

youpi


alekece:
After a quick glance the fix [06aac3a](https://github.com/LedgerHQ/lib-ledger-core/commit/06aac3a57cd8289b2fc82e71bd4866eac4f34c22) was already in that PR.

phaazon:
Do you plan to do the UXTO pagination here too? I would suggest to do that in another PR (but we need tracking on Jira), so that the PR remains smol enough.

PR #466, opened 2020-01-23
extract Bech32 from BTC branch into ledger-core

gagbo:
Only Bech32.cpp is included in core, all factories and parameters and
possible additional Bech32 types (BTCBech32, BCHBech32, or CosmosBech32)
will have to be implemented in the coin-specific projects.

Factory and Parameters are therefore declared in `ledger-core` and defined in `ledger-core-${coin}`. I need to wrap my head around this to check if this is fine


phaazon:
I’ll let @alekece review this as it might collide with his work on BTC. Thanks guys!

ledger-core-bitcoin/inc/bitcoin/bech32/BitcoinLikeBech32ParametersHelper.hpp
```diff
@@ -0,0 +1,53 @@
...
+    #endif
+#endif
+
+#include <core/math/bech32/Bech32ParametersHelper.hpp>
+
+namespace ledger {
+    namespace core {
+            class BitcoinLikeBech32ParametersHelper : public Bech32ParametersHelper<BitcoinLikeBech32ParametersHelper> {
+                public:
+                typedef Bech32ParametersHelper<BitcoinLikeBech32ParametersHelper> tBase;
```
alekece: Please use `using` instead of `typedef` to be consistent with our codebase.

BTW we have some documentation about good practice [here](https://github.com/LedgerHQ/lib-ledger-core/blob/misc/coding-conventions/CONVENTIONS.md) and [here](https://github.com/phaazon/lib-ledger-core/blob/docs/coin-migration/doc/improvements.md).
It's not completely finished so feel free to improve it or at least discuss with us for amelioration or new ideas.

gagbo:
> As I said to you just before, we need to abstract `BCHBech32` class and let the coin library implement the correct and needed behaviour.

It's a bit messy, but BCHBech32 is already an implementation of the abstract `Bech32` class which contains all the interface BCHBech32 implements. There are a few differences between BCH and BTC regarding Bech32, and the goal (for Cosmos) is to have a CosmosBech32 class in lib-core-cosmos with all the coin specific implementation

ledger-core-bitcoin/inc/bitcoin/bech32/BitcoinLikeBech32ParametersHelper.hpp
gagbo: :+1:

gagbo:
> Other topic about CRTP, we have to take a technical decision if we just inherit a plain-old abstract class and directly use the coin implementation in coin-specific code or use your current proposal with CRTP.

As discussed earlier, for the time being we'll leave the CRTP : all users in bitcoin (and cosmos) code of `getBech32Params`, `insertParameters` and `vector<Bech32Struct>::ALL` are using static methods, so this is the less intrusive way of introducing the change. Another refactoring might occur at later parts of migration

ledger-core-bitcoin/inc/bitcoin/bech32/BitcoinLikeBech32ParametersHelper.hpp
```diff
@@ -0,0 +1,53 @@
...
+    #else
+        #define LIBCORE_EXPORT
+    #endif
+#endif
+
+#include <core/math/bech32/Bech32ParametersHelper.hpp>
+
+namespace ledger {
+    namespace core {
+            class BitcoinLikeBech32ParametersHelper : public Bech32ParametersHelper<BitcoinLikeBech32ParametersHelper> {
```
phaazon: Small note: if you use both `public` inheritance and `public` fields only, you might want to use `struct` directly.

phaazon: That’s an optional comment though, feel free to ignore if you don’t like the `struct` keyword. :)

gagbo: Fair enough, I totally missed that :D

ledger-core/inc/core/math/bech32/Bech32ParametersHelper.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <core/math/bech32/Bech32Parameters.hpp>
+
+namespace ledger {
+    namespace core {
+        // CRTP used to get static polymorphism on the interface
+        template <class CoinLikeBech32ParametersHelper>
+        class Bech32ParametersHelper {
+            public:
+                static const Bech32Parameters::Bech32Struct getBech32Params(const std::string &networkIdentifier) {
+                    return CoinLikeBech32ParametersHelper::getCoinLikeBech32Params(networkIdentifier);
```
phaazon: Thats method is just a forward that doesn’t add any value. I struggle a bit to understand the goal of that CRTP class. Can you explain a bit why you need it and what it tries to solve? Thanks! 🙂

gagbo: Bech32 interface (the virtual methods [here](https://github.com/LedgerHQ/lib-ledger-core/blob/1bd439d5d9b05025259f788f0ff9d99c8571821f/ledger-core/inc/core/math/bech32/Bech32.hpp#L55-L74) ) is implemented by coins using coin-specific parameters ( defined [here](https://github.com/LedgerHQ/lib-ledger-core/blob/1bd439d5d9b05025259f788f0ff9d99c8571821f/ledger-core-bitcoin/src/bitcoin/bech32/BitcoinLikeBech32ParametersHelper.cpp#L44) for bitcoin )

The implementors of Bech32 interface (BTCBech32 example below, it's the same of BCHBech32) are using a call to a static `getBech32Params` method to get the correct private parameters attribute.
https://github.com/LedgerHQ/lib-ledger-core/blob/1bd439d5d9b05025259f788f0ff9d99c8571821f/ledger-core-bitcoin/inc/bitcoin/bech32/BTCBech32.hpp#L42

(I also noticed that I forgot another usage in `BitcoinLikeAddress` for [getVersionFromKeychainEngine code](https://github.com/LedgerHQ/lib-ledger-core/blob/1bd439d5d9b05025259f788f0ff9d99c8571821f/ledger-core-bitcoin/src/bitcoin/BitcoinLikeAddress.cpp#L77), I'm having issues compiling DerivationPath recently so I wasn't able to properly produce working tests for now :( )

From this point :
- All usages of `Bech32Parameters` come from static methods getBech32Params(string&) and insertParameters(..) ; and static member `ALL`
- Those static members depend on static `Bech32Parameters::Bech32Struct` which should stay coin-specific and only instantiated in coin-specific libraries (bitcoin specific params are defined [here](https://github.com/LedgerHQ/lib-ledger-core/blob/1bd439d5d9b05025259f788f0ff9d99c8571821f/ledger-core-bitcoin/src/bitcoin/bech32/BitcoinLikeBech32ParametersHelper.cpp#L44) - for example cosmos specific ones look like [this](https://github.com/gagbo/lib-ledger-core/blob/9820248bd409e2b8039f16c999d43af30abf731a/ledger-core-cosmos/inc/cosmos/bech32/CosmosBech32.hpp#L48) )

So the goal with CRTP was to provide a way to specify a `static getBech32Params(string&)` interface to use in coins.

Using classic inheritance here means that client code has to change to provide an instance of `BitcoinLikeBech32Parameters` to override the non-static `getBech32Params(string&)` method

gagbo:
> I would like to understand the CRTP stuff, especially in the context of BTC vs. Cosmos.


> The goal with CRTP was to provide a way to specify a `static getBech32Params(string&)` interface to use in coins.
> 
> Using classic inheritance here means that client code has to change to provide an instance of `BitcoinLikeBech32Parameters` to override the non-static `getBech32Params(string&)` method

The main question with using classic inheritance is "where should we instantiate the owner of called `getBech32Params(string&)` in each coin library ?"

alekece:
What is the status of this PR ? I think we should merge it as soon as possible and keep it mind some improvement have to be done later.
@gagbo @phaazon

gagbo:
After discussion, CRTP has been removed, and namespaced functions are used instead.
Pasting 8006c28 commit message below as a broad overview of changes in that last commit :

#### Core Bech32 changes :
- No default constructor of Bech32
- Added Bech32(Bech32Parameters::Bech32Struct) constructor for  easier inheritance

#### Core Bech32Parameters changes :
- Bech32Parameters / Bech32ParametersHelper aren't classes anymore

#### Bitcoin Bech32Parameters changes
- the 3 static values those classes exposed are now functions under  ledger::core::${coin} namespace. used as  `ledger::core::bitcoin::getBech32Params(string&)`
- The functions are defined in the coin-specific  Bech32ParametersHelpers header file
- Renaming to give names bech32 context within ledger::core::$coin  namespace

gagbo:
With those changes, adding a new bitcoin like bech32 implementation needs at most to do :
- register the new networkIdentifier in `ledger::core::BitcoinLikeBech32Factory::newBech32Instance`
- add the new `ledger::core::Bech32Parameters::Bech32Struct` in the dispatch of `ledger::core::bitcoin::getBech32Params`
- add this value to the `ledger::core::bitcoin::ALL_BECH32PARAMS` vector

ping @alekece @phaazon (moving factories out of `ledger::core` and into `ledger::core::bitcoin` has been considered, but will be done later when naming conventions have been discussed with more people)

PR #467, opened 2020-01-23
Fix XTZ: creation of same account (with KT accounts) under two different wallets

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-507

### DIFF
--- a/core/test/integration/synchronization/tezos_synchronization.cpp
+++ b/core/test/integration/synchronization/tezos_synchronization.cpp
@@ -32,6 +32,7 @@
 #include <gtest/gtest.h>
 #include "../BaseFixture.h"
 #include <set>
+#include <functional>
 #include <api/KeychainEngines.hpp>
 #include <utils/DateUtils.hpp>
 #include <wallet/tezos/database/TezosLikeAccountDatabaseHelper.h>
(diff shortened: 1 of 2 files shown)


core/test/integration/synchronization/tezos_synchronization.cpp
```diff
@@ -51,12 +52,20 @@ class TezosLikeWalletSynchronization : public BaseFixture {
...
+                            const std::string &nextWalletName) {
+        if (walletName.empty()) {
+            return;
+        }
         auto configuration = DynamicObject::newInstance();
         configuration->putString(api::Configuration::KEYCHAIN_DERIVATION_SCHEME,"44'/<coin_type>'/<account>'/<node>'/<address>");
         configuration->putString(api::TezosConfiguration::TEZOS_XPUB_CURVE, api::TezosConfigurationDefaults::TEZOS_XPUB_CURVE_ED25519);
         configuration->putString(api::Configuration::BLOCKCHAIN_EXPLORER_ENGINE, api::BlockchainExplorerEngines::TZSTATS_API);
-        auto wallet = wait(pool->createWallet("e847815f-488a-4301-b67c-378a5e9c8a61", "tezos", configuration));
+        configuration->putString(api::Configuration::BLOCKCHAIN_EXPLORER_API_ENDPOINT, "https://xtz-explorer.api.live.ledger.com/explorer");
```
phaazon: This was added and doesn’t related to the fix. Is this really needed?

KhalilBellakrid: Actually, the test was failing from time to time because of `https://api.tzstats.com/explorer` instability, which is the value by default.
Plus it is the one used in prod by the Live.

phaazon: ACK, LGTM.

PR #468, opened 2020-01-23
PostgreSQL support: build libcore with PG support only for JNI builds

KhalilBellakrid:
We activate PG support only for JNI builds (for the Vault) not to introduce any issues on Live's side.
Activating it for Live would require building PG in libcore and link against it: https://ledgerhq.atlassian.net/browse/LLC-517

### DIFF
--- a/.circleci/build_lib.sh
+++ b/.circleci/build_lib.sh
@@ -16,12 +16,12 @@ export PATH=$PATH:~/cmake_folder/bin
 ###
 
 function command_target_jni {
-  add_to_cmake_params -DTARGET_JNI=ON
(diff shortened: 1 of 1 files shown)


core/lib/soci_postgresql/CMakeLists.txt
```diff
@@ -13,9 +13,8 @@ add_library(soci_postgresql STATIC
         blob.cpp common.cpp common.h error.cpp factory.cpp row-id.cpp session.cpp soci-postgresql.h standard-into-type.cpp
         standard-use-type.cpp statement.cpp vector-into-type.cpp vector-use-type.cpp)
 
-set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+target_link_libraries(soci_postgresql PUBLIC pqxx_static)
```
phaazon: Aaaaah. Good. That will be easy to move that to the modularization then!

PR #469, opened 2020-01-23
Fix Ethereum ReorgLastBlock test segfault.

phaazon:
# Content

The problem was due to the fact most of the time, we use newDefaultPool,
which wraps the backend construction. In the case of ReorgLastBlock, the
WalletPool is hand-crafted and backend is not initialized (it’s a null
shared pointer).

Fixes LLC-514.

# Mandatory picture of animal

### DIFF
--- a/core/test/integration/synchronization/ethereum_synchronization.cpp
+++ b/core/test/integration/synchronization/ethereum_synchronization.cpp
@@ -31,6 +31,7 @@
 #include <gtest/gtest.h>
 #include "../BaseFixture.h"
 #include <set>
+#include <api/ConfigurationDefaults.hpp>
 #include <api/KeychainEngines.hpp>
 #include <api/EthereumLikeTransaction.hpp>
 #include <api/OperationOrderKey.hpp>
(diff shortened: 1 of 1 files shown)


core/test/integration/synchronization/ethereum_synchronization.cpp
```diff
@@ -347,6 +348,9 @@ TEST_F(EthereumLikeWalletSynchronization, ReorgLastBlock) {
     auto walletName = "e847815f-488a-4301-b67c-378a5e9c8a61";
     {
         auto fakeHttp = std::make_shared<test::FakeHttpClient>();
+
+        backend = std::static_pointer_cast<DatabaseBackend>(DatabaseBackend::getSqlite3Backend());
```
phaazon: Note to reviewer: this has to be done this way to be iso with the rest of the tests (they don’t pass the `usePostgreSQL` argument to `newDefaultPool`, so they use SQLite3).

phaazon: Extra note to reviewers: maybe the `static_pointer_cast` is not needed?

PR #470, opened 2020-01-23
Fix integer overflow on DerivationPath

alekece:
Since we modified `DerivationPath` to inherit from its api, we use `int32_t` (because of djinni limitation) instead of `uint32_t`. 
The problem is hardened derivation value is between 2^31 and 2^32 so the value cannot fit into its type anymore.

I found a "solution" : USE `int64_t` !

troll

One more thing I refactorized a bit the `DerivationPath::parse` method

### DIFF
--- a/ledger-core/src/core/utils/DerivationPath.cpp
+++ b/ledger-core/src/core/utils/DerivationPath.cpp
@@ -47,20 +47,16 @@ namespace ledger {
 
         }
 
-        DerivationPath::DerivationPath(const std::vector<int32_t> &path) : _path(path) {
-            // we cannot use unsigned integer because of some unsupported type in binding 
-            // so we have to check the validity of all the indexes in the path.
-            for (auto index : path)  {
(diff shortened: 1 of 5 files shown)


ledger-core/src/core/utils/DerivationPath.cpp
```diff
@@ -47,20 +47,16 @@ namespace ledger {
 
         }
 
-        DerivationPath::DerivationPath(const std::vector<int32_t> &path) : _path(path) {
-            // we cannot use unsigned integer because of some unsupported type in binding 
-            // so we have to check the validity of all the indexes in the path.
-            for (auto index : path)  {
-                assertIndexIsValid(index, "ledger::core::DerivationPath::DerivationPath");
-            }
```
alekece: I removed those lines because the check is on the index itself and not the value behind that index (I created a JIRA ticket to manage negative derivation value)

phaazon:
Have you tried the Ripple tests? They are all green on `design/modularization`. That should give you a good idea about the impacts.

PR #471, opened 2020-01-28
Fix includes for btc files

gagbo:
Optional and Logger had small typos

DjinniHelpers file does not exist anymore


gagbo:
I don't have permission to merge on this repo, I don't know if I'm the one expected to do this ?

We didn't have compilation errors before because `ledger-core-bitcoin` needs `ledger-core` to be *installed* in order to compile, so all old header files were copied in the install location and not cleaned up after the header file changed name, so it was still found on local "dirty" builds

phaazon:
Fair enough, thanks for the explanation! I can merge! :)

PR #474, opened 2020-01-29
Implement Services and WalletStore API interfaces

phaazon:
# Content

This PR provides support for `api::Services` and `api::WalletStore`, effectively making those objects callable from foreign interfaces.

# Mandatory picture of Satan’s Pet™ being discovered

### DIFF
--- a/ledger-core/idl/wallet_store.djinni
+++ b/ledger-core/idl/wallet_store.djinni
@@ -5,7 +5,7 @@
 
 # Class representing a cache of wallets.
 WalletStore = interface +c {
-    # Return number of wallets instanciated under wallet pool.
+    # Return number of wallets instantiated under the store.
     # @param callback, Callback object returns a 32 bits integer, count of wallets
     getWalletCount(callback: callback2<void, optional<i32>, optional<Error>>);
 
@@ -28,14 +28,14 @@ WalletStore = interface +c {
     # > that might have been created before remain intact
     updateWalletConfig(name: string, configuration: DynamicObject, callback: callback2<void, optional<ErrorCode>, optional<Error>>);
 
-    # Instanciate a new wallet under wallet pool.
+    # Instantiate a new wallet under the store.
     # @param name, string, name of newly created wallet
     # @param currency, Currency object, currency of the wallet
     # @param configuration, DynamicObject object, configuration of wallet (preferences)
     # @param callback, Callback object returning a Wallet object
     createWallet(name: string, currency: Currency, configuration: DynamicObject, callback: callback2<void, optional<Wallet>, optional<Error>>);
 
-    # Return all supported currencies by wallet pool, at least one wallet support one of returned currencies.
+    # Return all supported currencies by the wallet store, at least one wallet support one of returned currencies.
     # @param callback, ListCallback object, returns a list of Currency objects
     getCurrencies(callback: callback2<void, optional<list<Currency>>, optional<Error>>);
 
--- a/ledger-core/idl/wallet_store.djinni
+++ b/ledger-core/idl/wallet_store.djinni
@@ -1,0 +1,50 @@
+@import "configuration.djinni"
+@import "currency.djinni"
+@import "error.djinni"
+@import "wallet.djinni"
+
+# Class representing a cache of wallets.
+WalletStore = interface +c {
+    # Return number of wallets instanciated under wallet pool.
+    # @param callback, Callback object returns a 32 bits integer, count of wallets
+    getWalletCount(callback: callback2<void, optional<i32>, optional<Error>>);
+
+    # Get instanciated wallets having index in a given range.
+    # @param from, 32 bits integer, lower bound of indices to pick
+    # @param to, 32 bits integer, upper bound of indices to pick
+    # @param callback, ListCallback object returns a list of Wallet objects
+    getWallets(from: i32, size: i32, callback: callback2<void, optional<list<Wallet>>, optional<Error>>);
+
+    # Get wallet with a giver name.
+    # @param name, string, name of wallet to look for
+    # @param callback, Callback object returns a Wallet object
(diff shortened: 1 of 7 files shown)


ledger-core/idl/wallet_store.djinni
```diff
@@ -0,0 +1,50 @@
...
+
+    # Update wallet configuration
+    # @param name, string, name of wallet to update
+    # @param configuration, DynamicObject object, configuration object with fields to update
+    # @param callback, Callback object returns the error code, returns ErrorCode::FUTURE_WAS_SUCCESSFULL if everything is fine
+    # > Note: other fields that are not passed in 'configuration' parameter
+    # > that might have been created before remain intact
+    updateWalletConfig(name: string, configuration: DynamicObject, callback: callback2<void, optional<ErrorCode>, optional<Error>>);
+
+    # Instanciate a new wallet under wallet pool.
```
alekece: Do we have always that notion of wallet pool ? Is it wallet store instead ?

ledger-core/idl/wallet_store.djinni
```diff
@@ -0,0 +1,50 @@
...
+    # @param name, string, name of newly created wallet
+    # @param currency, Currency object, currency of the wallet
+    # @param configuration, DynamicObject object, configuration of wallet (preferences)
+    # @param callback, Callback object returning a Wallet object
+    createWallet(name: string, currency: Currency, configuration: DynamicObject, callback: callback2<void, optional<Wallet>, optional<Error>>);
+
+    # Return all supported currencies by wallet pool, at least one wallet support one of returned currencies.
```
alekece: Ditto

ledger-core/src/core/Services.cpp
```diff
@@ -133,11 +134,11 @@ namespace ledger {
...
         }
 
-        const std::string &Services::getTenant() const {
+        std::string Services::getTenant() const {
             return _tenant;
         }
 
-        const std::string Services::getPassword() const {
+        std::string Services::getPassword() const {
             return _password;
```
alekece: What is the purpose of creating a copy of tenant and password strings instead of returning a reference to its ?

ledger-core/src/core/wallet/WalletStore.cpp
```diff
@@ -19,10 +19,34 @@ namespace ledger {
             return Option<api::Currency>();
         }
 
+        void WalletStore::getCurrency(
+            const std::string & name,
+            const std::function<void(std::experimental::optional<api::Currency>, std::experimental::optional<api::Error>)> & callback
+        ) {
+            auto self = shared_from_this();
+            Future<api::Currency>::async(_services->getDispatcher()->getMainExecutionContext(), [self, name] () {
```
alekece: Maybe we can use in-place capture for `self` variable to avoid a copy and "under the hood" count references involved in `shared_ptr` ?

ledger-core/src/core/wallet/WalletStore.cpp
```diff
@@ -19,10 +19,34 @@ namespace ledger {
...
+
         std::vector<api::Currency> const& WalletStore::getCurrencies() const {
             return _currencies;
         }
 
+        void WalletStore::getCurrencies(
+            const std::function<void(std::experimental::optional<std::vector<api::Currency>>, std::experimental::optional<api::Error>)> & callback
+        ) {
+            auto self = shared_from_this();
+            Future<std::vector<api::Currency>>::async(_services->getDispatcher()->getMainExecutionContext(), [self] () {
```
alekece: Ditto

ledger-core/src/core/wallet/WalletStore.cpp
```diff
@@ -73,6 +97,14 @@ namespace ledger {
             });
         }
 
+        void WalletStore::getWalletCount(
+            const std::function<void(std::experimental::optional<int32_t>, std::experimental::optional<api::Error>)> & callback
+        ) {
+            getWalletCount().map<int32_t>(_services->getContext(), [] (const int64_t count) {
+                return (int32_t) count;
```
alekece: Please replace C cast by the proper C++ one.

ledger-core/idl/wallet_store.djinni
phaazon: Good call, I need to update those comments.

ledger-core/src/core/Services.cpp
phaazon: Implementing the interface. Mostly. :D

ledger-core/src/core/wallet/WalletStore.cpp
phaazon: You mean a _move_? Can lambda support that?

ledger-core/src/core/wallet/WalletStore.cpp
phaazon: Good catch.

ledger-core/src/core/wallet/WalletStore.cpp
alekece: I mean in-place capture variable declaration (waouh such a long term) such as :
`[self = shared_from_this(), ...](...) {}`

phaazon: Hm. I didn’t know you could do that. Are you sure it’s not going to make an extra copy as well?

alekece: It's quite new in lambda capture; the `self` variable will be directly declared as lambda variable instead of captured variable from the stack context.

phaazon: Actually, not here, because the value is used below in the same scope. I’ll apply the idea to the rest of the file whenever possible though.

ledger-core/src/core/wallet/WalletStore.cpp
phaazon: Same remark as with line 46.

PR #477, opened 2020-01-30
Improve integration BaseFixture

alekece:
The purpose of this PR is to improve our current `BaseFixture` class to ease the readability of our current tests and create a common scope for new coin tests.

I also removed unwanted log message and start to remove coin-specific global variable such as `P2PKH_MEDIUM_XPUB_INFO`.

## Mandatory picture of a cat 
cat3

### DIFF
--- a/ledger-core/test/integration/BaseFixture.cpp
+++ b/ledger-core/test/integration/BaseFixture.cpp
@@ -29,75 +29,15 @@
  *
  */
 
-#include <core/utils/Hex.hpp>
-
 #include <utils/FilesystemUtils.hpp>
+
+#include <core/utils/Hex.hpp>
+#include <core/wallet/WalletDatabaseEntry.hpp>
+#include <core/wallet/WalletDatabaseHelper.hpp>
+#include <core/wallet/AccountDatabaseHelper.hpp>
+
 #include "BaseFixture.hpp"
 #include "IntegrationEnvironment.hpp"
-
-api::ExtendedKeyAccountCreationInfo P2WPKH_MEDIUM_XPUB_INFO(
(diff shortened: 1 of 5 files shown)


ledger-core/test/integration/BaseFixture.cpp
```diff
@@ -147,7 +142,7 @@ std::shared_ptr<WalletStore> BaseFixture::newWalletStore(
     return std::make_shared<WalletStore>(services);
 }
 
-void BaseFixture::createWallet(
+void BaseFixture::createWalletInDatabase(
```
phaazon: Good idea!

ledger-core/test/integration/BaseFixture.hpp
```diff
@@ -94,19 +93,41 @@ class BaseFixture : public ::testing::Test {
...
     );
 
-    void createAccount(
+    void createAccountInDatabase(
         const std::shared_ptr<Services>& services,
         const std::string &walletName,
         int32_t index
     );
 
+    template<class Account>
```
phaazon: I think we should use `typename` everywhere as it encompasses more usages, even though it’s unlikely we use anything else here. I think _one syntax to rule them all_ is better.

ledger-core/test/integration/BaseFixture.hpp
```diff
@@ -94,19 +93,41 @@ class BaseFixture : public ::testing::Test {
...
+        std::shared_ptr<AbstractWallet> const &wallet,
+        int32_t index,
+        api::AccountCreationInfo info) 
+    {
+        info.index = index;
+
+        return std::dynamic_pointer_cast<Account>(wait(wallet->newAccountWithInfo(info)));
+    }
+
+    template <class Account>
```
phaazon: Ditto.

ledger-core/test/integration/BaseFixture.hpp
alekece: I agree about __one syntax to rule them all__ is better. 

I used to write `class` in template parameter because is shorter but the main reason is `typename` keyword are a second meaning - specify to the compiler that the symbol in the class/struct is a type. 
So to resume is all about cognitive complexity and using the right keyword (even if `class` keyword has two meaning as well).

PR #478, opened 2020-01-30
Add WalletFixture class

alekece:
This PR allow use to test wallet (new and existing one) really quickly and with less code than actual one. 

### What happens "under the hood"
The `testWallet` method will create `Services`, `WalletStore` and coin-specific wallet factory instances and register the `currency` given as input parameter before calling the user-specified callback (this function should contains the test code about wallet).

## Mandatory picture of a cat
cat4

### DIFF
--- a/ledger-core/test/integration/WalletFixture.hpp
+++ b/ledger-core/test/integration/WalletFixture.hpp
@@ -34,6 +34,7 @@
 #include <core/wallet/AbstractWallet.hpp>
 #include <core/wallet/WalletStore.hpp>
 #include <core/collections/DynamicObject.hpp>
+#include <core/api/ErrorCode.hpp>
 
 #include "BaseFixture.hpp"
 
@@ -51,6 +52,7 @@ public:
         std::string const &name,
         api::Currency const &currency,
         TestStrategy strategy,
+        std::shared_ptr<api::DynamicObject> configuration,
         FunctionType const &f) {
 
         auto services = newDefaultServices();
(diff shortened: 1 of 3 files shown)


ledger-core/test/integration/WalletFixture.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+        auto wallet = [&]() {
+            switch (strategy) {
+                case TestStrategy::NEW_WALLET:
+                    return wait(walletStore->createWallet(name, currency.name, api::DynamicObject::newInstance()));
+                case TestStrategy::EXISTING_WALLET:
+                default:
+                    return wait(walletStore->getWallet(name));
+            }
+        }();
```
alekece: This is called IIFE (for immediately invoked function expression). 

It allows complex variable initialization without the burden to declare an uninitialized variable (this type have a default constructor ? If yes, this default constructor may introduce some undesired side-effects).

phaazon: So as discussed out loud, this kind of trick is necessary because the language lacks some feature (the feature it lacks here is pattern-matching and statement-as-expressions).

However, you don’t need it either case. You could replace the whole thing with:

```cpp
auto wallet = strategy == TestTrategy::NEW_WALLET ?
  wait(walletStore->createWallet(name, currency.name, api::DynamicObject::newInstance())) :
  return wait(walletStore->getWallet(name));
```

Removing the visual noise.

phaazon: And just for the cultural note, in Rust, undefined variables are _safe_ until they’re used. So you could write:

```rust
let x;

match something {
  Foo => x = 1,
  Bar => x = 2,
}
```

Which compiles and is completly safe — but it’s so trivial we don’t do that. Instead, and about the feature C++ is missing, in Rust:

```rust
let x = match something {
  Foo => 1,
  Bar => 2,
};
```

alekece: Thanks for the Rush example, really instructive. 

The technical choice to use IIFE here is to let the possibly to add new enumeration in future.
Since the need is not really fixed and probably will evolve with new coin integration; I prefer to avoid to use ternary statement here.

alekece:
Finally I took the decision to remove a lot of logic in `WalletFixture` to be more flexible in underlying tests.

ledger-core/test/integration/WalletFixture.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <core/wallet/WalletStore.hpp>
+#include <core/collections/DynamicObject.hpp>
+#include <core/api/ErrorCode.hpp>
+
+#include "BaseFixture.hpp"
+
+template <class WalletFactory>
+class WalletFixture : public BaseFixture {
+public:
+    static constexpr auto DEFAULT_PASSWORD = "";
```
alekece: These two constant variables can be needed for underlying tests like change password (we need the old password to set a new one).

PR #480, opened 2020-01-31
Fix segfault in Linux + node with OpenSSL 1.1. (node > 8).

phaazon:
Relates to LLC-521.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,8 @@ dependencies:
     - ~/cmake-3.12.3
 jobs:
   build_linux_release:
+    environment:
+        CORE_SSL_1_1: arch_ssl_1_1
     docker:
       - image: debian:stretch
     working_directory: ~/lib-ledger-core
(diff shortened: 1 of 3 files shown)


.circleci/copy_artifacts.sh
```diff
@@ -12,6 +12,8 @@ if [ "$unamestr" == "Linux" ]; then
 		export BUILD_TYPE=linux/jni
 	elif [ "$TARGET" == "android" ]; then
 		export BUILD_TYPE=android/${ARCH}
+    elif [ "$TARGET" == "$CORE_SSL_1_1" ]; then
```
phaazon: Indentation is a bit meh.

.circleci/config.yml
```diff
@@ -38,6 +40,14 @@ jobs:
...
           command: |
             . .circleci/copy_artifacts.sh
+      - run:
+          name: Build_Library (with OpenSSL 1.1)
+          command: |
+            . .circleci/build_lib.sh Release $CORE_SSL_1_1
+      - run:
+          name: Copy_Artifacts (with OpenSSL 1.1)
+          command: |
+            . .circleci/copy_artifacts.sh Release $CORE_SSL_1_1
```
KhalilBellakrid: Here you should give the `copy_artifacts.sh` the `CORE_SSL_1_1` as first argument so `copy_artifacts.sh` will set `TARGET` properly.

phaazon: Good catch.

PR #482, opened 2020-02-06
Add SSL_SUPPORT configuration to build Core library with SSL support.

KhalilBellakrid:
This relates to : https://ledgerhq.atlassian.net/browse/INFRA-1572.

Before we didn't need SSL support because we were using only Crypto libraries on OpenSSL.
Eventually we were using SSL for tests because we were implementing our own Http client with SSL support.
We didn't face that issue before with any client because Http clients are provided by the host of the Core library.

We are not only building with support of SSL but in this commit we initialize of course the SSL libraries when using PG backend.

### DIFF
--- a/core/src/database/PostgreSQLBackend.cpp
+++ b/core/src/database/PostgreSQLBackend.cpp
@@ -34,6 +34,7 @@
 #include <soci-postgresql.h>
 
 #ifdef SSL_SUPPORT
+    #include <thread>
     #include <openssl/ssl.h>
 #endif
 
@@ -80,13 +81,16 @@ namespace ledger {
             throw make_exception(api::ErrorCode::IMPLEMENTATION_IS_MISSING, "Change of password is not handled with PostgreSQL backend.");
         }
 
+        std::once_flag PostgreSQLBackend::sslFlag;
         void PostgreSQLBackend::initSSLLibraries() {
 #ifdef SSL_SUPPORT
-    #if OPENSSL_VERSION_NUMBER < 0x10100000L
-            SSL_library_init();
-    #else
-            OPENSSL_init_ssl(0, NULL);
-    #endif
+            std::call_once(PostgreSQLBackend::sslFlag, [] () {
+            #if OPENSSL_VERSION_NUMBER < 0x10100000L
+                SSL_library_init();
+            #else
+                OPENSSL_init_ssl(0, NULL);
+            #endif
+            });
 #endif
         }
     }
--- a/core/src/database/PostgreSQLBackend.cpp
+++ b/core/src/database/PostgreSQLBackend.cpp
@@ -32,17 +32,26 @@
 #include <utils/Exception.hpp>
 #include <api/ConfigurationDefaults.hpp>
 #include <soci-postgresql.h>
+
+#ifdef SSL_SUPPORT
+    #include <openssl/ssl.h>
+#endif
+
 using namespace soci;
 
 namespace ledger {
     namespace core {
         PostgreSQLBackend::PostgreSQLBackend() : DatabaseBackend(),
                                                  _connectionPoolSize(api::ConfigurationDefaults::DEFAULT_PG_CONNECTION_POOL_SIZE)
-        {}
+        {
+            initSSLLibraries();
+        }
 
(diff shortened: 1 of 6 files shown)


core/src/database/PostgreSQLBackend.cpp
```diff
@@ -32,17 +32,26 @@
...
+
 using namespace soci;
 
 namespace ledger {
     namespace core {
         PostgreSQLBackend::PostgreSQLBackend() : DatabaseBackend(),
                                                  _connectionPoolSize(api::ConfigurationDefaults::DEFAULT_PG_CONNECTION_POOL_SIZE)
-        {}
+        {
+            initSSLLibraries();
```
phaazon: The main problem with going this way is that you’re doing something non-trivial — initializing OpenSSL — in something that is not immediately related (i.e. a backend database).

I’m not sure this is the right place to do that.

core/src/database/PostgreSQLBackend.cpp
```diff
@@ -32,17 +32,26 @@
...
+        }
 
         PostgreSQLBackend::PostgreSQLBackend(int32_t connectionPoolSize) : DatabaseBackend(),
                                                                            _connectionPoolSize(connectionPoolSize)
-        {}
+        {
+            initSSLLibraries();
```
phaazon: Ditto.

core/src/database/PostgreSQLBackend.cpp
```diff
@@ -70,5 +79,15 @@ namespace ledger {
                                             soci::session &session) {
             throw make_exception(api::ErrorCode::IMPLEMENTATION_IS_MISSING, "Change of password is not handled with PostgreSQL backend.");
         }
+
+        void PostgreSQLBackend::initSSLLibraries() {
```
phaazon: You should wrap that function in another function adding the static initializer safety (or do it inside that function, I don’t really care). And we must ensure there’s a TLS on the actual initialization, otherwise it’s going to be madness when using two backends in two different threads.

core/src/database/PostgreSQLBackend.cpp
phaazon: I think the _saner_ approach would be to be that initializing in the `WalletPool` (which will translate to `Services` in the modularized core) and ask for its initialization there. That means that we cannot create a `PostgreSQLBackend` with the default ctor — that seems perfectly fine to me, as it has an external dependency.

What do you guys think? It could be something like:

```cpp
WalletPool::initSSL() {
  static done = false;
  
  if (!done) {
#ifdef SSL_SUPPORT
    #if OPENSSL_VERSION_NUMBER < 0x10100000L
            SSL_library_init();
    #else
            OPENSSL_init_ssl(0, NULL);
    #endif
#endif

    done = true;
  }
}
```

KhalilBellakrid: Thought about it ! But potentially we could use PG without WalletPool in a future, then it would take us time time to think about that, we should bind PG with SSL support.

KhalilBellakrid: Second point, I don't like the fact that client should call that explicitly ! They should not worry about that

teams2ua: https://en.cppreference.com/w/cpp/thread/call_once is thread safe variant of this.

phaazon: `call_once` is a really good point! @KhalilBellakrid that’s I split `WalletPool` into `Services + WalletStore` in the modularization. :)

KhalilBellakrid: @teams2ua as discussed, agreed ! 👌 
@phaazon that's exactly kind of things to enhance in new Core, so I let you port that fix on it taking into consideration all what have been discussed above ! Thank you sir !

core/src/database/PostgreSQLBackend.cpp
```diff
@@ -70,5 +81,18 @@ namespace ledger {
...
+#ifdef SSL_SUPPORT
+            std::once_flag sslFlag;
```
teams2ua: shouldn't sslFlag be static, otherwise it is new value each call, right?

phaazon: The link you pasted even puts that as a global 😨

PR #483, opened 2020-02-10
Make fixes and changes to prepare the new WD release

KhalilBellakrid:
**Fix:** now that we are using a "real" database :trollface: we could notice some issues that we were not taking care of on DB side, as an example the input data is a `VARCHAR(255)` which is not always the case : https://explorers.api.live.ledger.com/blockchain/v3/eth/transactions/0xd1811f0c7b1dcb11bc9b28a3be194ec3d1bb273318d23e5353ee65ad6754a8e0
So now we check that and if not we drop input data. This information is used only to display details about transactions, so if an end user wants these details about their tx they have links to official explorers in the app to do so.

**Resilience:**
- Sync token deletion: issues raised by Live & Vault about failing to drop this token, now we wrap that deletion in a `Try` and if we fail to drop it we display a warning.
- `UNIQUE_CONSTRAINT` failure on `bitcoin_inputs`:issues raised by Live & Vault about failing sync when we have some issues with explorers, now we wrap the `putTransaction` in a `Try` to handle better failures caused by this DB update, as a consequence, he will miss some txs on his history and worst case his balance won't be updated but he will still be able to make transactions with previous state.
- Sync failure displaying a `Reorganization` error message: even if we have test covering that use case it seems that still some situations are not well handled. To avoid that we don't throw when `recoverWith` is not working, we allow to continue synchronization. This plus the above changes should not impact the clients (cf other impacts of above changes).

### DIFF
--- a/core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
+++ b/core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
@@ -244,6 +244,8 @@ namespace ledger {
                         buddy->logger->warn("Failed to delete synchronization token {} for account#{} of wallet {}",
                                             static_cast<char *>(buddy->token.getValue()), buddy->account->getIndex(),
                                             buddy->account->getWallet()->getName());
+                        // We return a successful Unit because deleting the sync token should not be a "failure"
+                        // Note: in a near future we'll try to get rid of sync token mechanism
                         return Future<Unit>::successful(unit);
                     }
                     return tryKillSession.getValue();
--- a/core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
+++ b/core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
@@ -237,14 +237,14 @@ namespace ledger {
                 }).template flatMap<Unit>(account->getContext(), [buddy, self] (const Unit&) {
                     return self->synchronizeBatches(0, buddy);
                 }).template flatMap<Unit>(account->getContext(), [self, buddy] (const Unit&) {
-                    auto tryKillSession = Try<Unit>::from([=](){
+                    auto tryKillSession = Try<Future<Unit>>::from([=](){
                         return self->_explorer->killSession(buddy->token.getValue());
                     });
                     if (tryKillSession.isFailure()) {
-
-                        buddy->logger->warn("Failed to delete synchronization token {} for account#{} of wallet {}", * buddy->token.getValue(), buddy->account->getIndex(),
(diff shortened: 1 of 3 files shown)


core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -236,7 +237,16 @@ namespace ledger {
...
                 }).template flatMap<Unit>(account->getContext(), [self, buddy] (const Unit&) {
-                    return self->_explorer->killSession(buddy->token.getValue());
+                    auto tryKillSession = Try<Future<Unit>>::from([=](){
+                        return self->_explorer->killSession(buddy->token.getValue());
+                    });
+                    if (tryKillSession.isFailure()) {
+                        buddy->logger->warn("Failed to delete synchronization token {} for account#{} of wallet {}",
+                                            static_cast<char *>(buddy->token.getValue()), buddy->account->getIndex(),
+                                            buddy->account->getWallet()->getName());
+                        return Future<Unit>::successful(unit);
```
alekece: Why do you return a successful `Future<Unit>` when something bad occurs ? Is it any functional choice behind that stuff ?

core/src/wallet/ethereum/explorers/api/EthereumLikeTransactionParser.cpp
```diff
@@ -216,10 +217,13 @@ namespace ledger {
                     std::istringstream issNonce(value);
                     issNonce >> _transaction->nonce;
                 } else if (_lastKey == "input") {
-                    if (isInternalTx) {
-                        _transaction->internalTransactions.back().inputData = fromStringToBytes(value);
-                    } else {
-                        _transaction->inputData = fromStringToBytes(value);
+                    auto inputData = fromStringToBytes(value);
```
alekece: The constness guy has something to say : "Put const qualifier in front of that variable" :trollface:

Another suggestion is you can avoid to copy the `inputData` by using move semantic.

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
KhalilBellakrid: Yes, failing on sync token deletion should not be a "functional failure" that's why we return a successful `Unit` and raise a warning on that.
We are trying to figure out how to get rid of that sync token from Backend side e.g. include it in header or better have a solution from infra side.

alekece: I suggest to add a comment when the code doesn't reflect the "normal intention" such as returning false positive case.

RomanWlm:
🔥

PR #486, opened 2020-02-11
Migrate ETH account test

alekece:
This PR contains the UNIQUE account test for ETH.

BTW, the test fails on the latest expectation and I can't figure out if this is a regression or only a bad test ?
@KhalilBellakrid can you check it (see the `TODO` comment) ?

## Mandatory picture of a animal
hotdog_dog

### DIFF
--- a/ledger-core-ethereum/test/AccountTests.cpp
+++ b/ledger-core-ethereum/test/AccountTests.cpp
@@ -50,6 +50,20 @@ TEST_F(EthereumAccounts, FirstEthAccountInfo) {
     auto info = wait(wallet->getNextAccountCreationInfo());
     EXPECT_EQ(info.index, 0);
     EXPECT_EQ(info.owners[0], "main");
-    // TODO: info.derivations[0] is equal to "44'/61'/0'". Is it a regression ?
+    EXPECT_EQ(info.derivations[0], "44'/60'/0'");
+}
+
+TEST_F(EthereumAccounts, FirstEthCustomDerivationAccountInfo) {
+    auto const currency = currencies::ethereum();
+    auto configuration = api::DynamicObject::newInstance();
+
+    registerCurrency(currency);
+
+    configuration->putString(api::Configuration::KEYCHAIN_DERIVATION_SCHEME, "44'/<coin_type>'/<account>'/<node>/<address>");
+
+    auto wallet = wait(walletStore->createWallet("my_wallet", currency.name, configuration));
+    auto info = wait(wallet->getNextAccountCreationInfo());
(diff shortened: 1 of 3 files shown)


ledger-core-ethereum/test/AccountTests.cpp
```diff
@@ -0,0 +1,55 @@
...
+TEST_F(EthereumAccounts, FirstEthAccountInfo) {
+    auto const currency = currencies::ethereum();
+
+    registerCurrency(currency);
+
+    auto wallet = wait(walletStore->createWallet("my_wallet", currency.name, api::DynamicObject::newInstance()));
+    auto info = wait(wallet->getNextAccountCreationInfo());
+    EXPECT_EQ(info.index, 0);
+    EXPECT_EQ(info.owners[0], "main");
+    // TODO: info.derivations[0] is equal to "44'/61'/0'". Is it a regression ?
```
gagbo: Looks like 61 is [ETC](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)

alekece: So the expected value is probably wrong ?

alekece: @KhalilBellakrid Can you confirm it ?

gagbo: Looks like a regression to me.
This [line](https://github.com/alekece/lib-ledger-core/blob/design/ethereum-account-tests/ledger-core-ethereum/src/ethereum/EthereumLikeCurrencies.cpp#L41) uses the same bip44 id as Ethereum classic although there is a `currencies::ethereum_classic()` function.

I *guess* eth classic also uses ETH as its native currency, but we have to be sure of this, since we also use the "ETH" name for it [here](https://github.com/alekece/lib-ledger-core/blob/design/ethereum-account-tests/ledger-core-ethereum/src/ethereum/EthereumLikeCurrencies.cpp#L57)

KhalilBellakrid: I agree , it should be `60` ! Nice catch @gagbo !

phaazon: The problem is in `EthereumLikeCurrencies.cpp`. The `currencies::ethereum()` function uses the `ETHEREUM_CLASSIC_COIN_ID` instead of `ETHEREUM_COIN_ID`!

alekece: Good catch ! Effectively we made a bad copy-paste in our currency definitions

PR #490, opened 2020-02-12
Add XTZ account test

alekece:
As the previous PR all you have to know is in the title.

BTW, we have a long-rude discussion about the final test (e.g `info.derivation[0] == "44'/1729'/0/0/0"`) and decide to comment it. The comment just above the test explains why.

## Mandatory picture of a cat
dancing_cat


gagbo:
I think it's at least safe to assert in tests that the derivation starts with `"44'/1729'/"` since it's what BIP44 suggests. All things afterwards are probably custom to various wallet implementers

PR #491, opened 2020-02-13
Add support for ordered migration rollbacks.

phaazon:
# Content

We need to ensure all coins’s database migrations are rolled back before
rolling back the ledger-core one. This is due to the fact internally,
tables might have relationship and foreign keys. Some database backends
will automatically erase lines in child tables but the structure — i.e.
the tables themselves — will remain in database.

In the doubt, it’s better to drop tables in order.

Relates to LLC-509.

# Mandatory picture of the best animal on Earth and the cosmos

### DIFF
--- a/ledger-core/inc/core/database/DatabaseSessionPool.hpp
+++ b/ledger-core/inc/core/database/DatabaseSessionPool.hpp
@@ -75,20 +75,48 @@ namespace ledger {
                 soci::transaction tr(sql);
                 Migration<T::CURRENT_VERSION, T>::forward(sql, version);
 
+                // register rollback migration
+                _rollbackMigrations.emplace_back([](soci::session& sql) {
+                    int version = getDatabaseMigrationVersion<T>(sql);
+
+                    soci::transaction tr(sql);
+                    Migration<T::CURRENT_VERSION, T>::backward(sql, version);
+                    tr.commit();
+                });
+
                 tr.commit();
             }
 
-            /// Rollback a migration for a given migration system.
-            template <typename T>
-            void rollbackMigration() {
+            /// Run migration forward for the ledger-core library.
+            ///
+            /// This is a special case because we don’t want to register ourselves for rollback,
+            /// since we will call that function explicitly.
+            template <>
+            void forwardMigration<CoreMigration>() {
                 soci::session sql(getPool());
-                int version = getDatabaseMigrationVersion<T>(sql);
+                int version = getDatabaseMigrationVersion<CoreMigration>(sql);
 
                 soci::transaction tr(sql);
-                Migration<T::CURRENT_VERSION, T>::backward(sql, version);
+                Migration<CoreMigration::CURRENT_VERSION, CoreMigration>::forward(sql, version);
 
-                unsetupMigrations(sql);
+                tr.commit();
+            }
 
+            /// Rollback all migrations for a given migration system.
+            void rollbackMigrations() {
+                soci::session sql(getPool());
+
+                // rollback coins first
+                for (auto rollbackCoinMigration : _rollbackMigrations) {
+                    rollbackCoinMigration(sql);
+                }
+
+                _rollbackMigrations.clear();
+
+                int version = getDatabaseMigrationVersion<CoreMigration>(sql);
+
+                soci::transaction tr(sql);
+                Migration<CoreMigration::CURRENT_VERSION, CoreMigration>::backward(sql, version);
                 tr.commit();
             }
 
@@ -100,6 +128,11 @@ namespace ledger {
             soci::connection_pool _pool;
             std::ostream* _logger;
             LoggerStreamBuffer _buffer;
+
+            // list of rollback migrations to play when dropping the database; this is a necessary
+            // object to ensure we drop all “registered coin” before trying to remove everything from
+            // the database
+            std::vector<std::function<void (soci::session& sql)>> _rollbackMigrations;
         };
     }
 }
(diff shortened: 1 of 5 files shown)


ledger-core/inc/core/database/Migrations.hpp
```diff
@@ -102,15 +102,15 @@ namespace ledger {
...
-                  if (currentVersion >= 0) {
-                      // after rolling back this migration, we won’t have anything left, so we only
-                      // update the version for > 0
-                      if (currentVersion != 0) {
-                          auto prevVersion = version - 1;
-                          sql << "UPDATE __database_meta__ SET id = :id, version = :version", soci::use(T::COIN_ID), soci::use(prevVersion);
-                      }
+                  if (currentVersion > 0) {
+                      // if the migration number is greater than 0, we just perform the regular migration
+                      auto prevVersion = version - 1;
```
alekece: You can use `constexpr` and use it as non-templated type : `Migration<prevVersion, T>`

ledger-core/inc/core/database/DatabaseSessionPool.hpp
```diff
@@ -75,20 +75,48 @@ namespace ledger {
...
+                soci::session sql(getPool());
+
+                // rollback coins first
+                for (auto rollbackCoinMigration : _rollbackMigrations) {
+                    rollbackCoinMigration(sql);
+                }
+
+                _rollbackMigrations.clear();
+
+                int version = getDatabaseMigrationVersion<CoreMigration>(sql);
```
alekece: Please add `const` qualifier

ledger-core/inc/core/database/DatabaseSessionPool.hpp
```diff
@@ -75,20 +75,48 @@ namespace ledger {
...
-                unsetupMigrations(sql);
+                tr.commit();
+            }
 
+            /// Rollback all migrations for a given migration system.
+            void rollbackMigrations() {
+                soci::session sql(getPool());
+
+                // rollback coins first
+                for (auto rollbackCoinMigration : _rollbackMigrations) {
```
alekece: You can avoid a copy by taking reference values in the range-based for loop :
```
for (auto& rollbackMigration : _rollbackMigrations)
```
Or use this beautiful line of code :
```
std::for_each(
   rollbackMigrations.cbegin(), 
   rollbackMigrations.cend(), 
   std::bind(
      &decltype(rollbackMigrations)::value_type::operator(),
      std::placeholders::_1,
      std::ref(sql)
   )
);
```
:trollface:

ledger-core/inc/core/database/Migrations.hpp
```diff
@@ -102,15 +102,15 @@ namespace ledger {
...
-                          sql << "UPDATE __database_meta__ SET id = :id, version = :version", soci::use(T::COIN_ID), soci::use(prevVersion);
-                      }
+                  if (currentVersion > 0) {
+                      // if the migration number is greater than 0, we just perform the regular migration
+                      auto constexpr prevVersion = version - 1;
+                      sql << "UPDATE __database_meta__ SET version = :version WHERE id = :id", soci::use(prevVersion), soci::use(T::COIN_ID);
 
-                      Migration<version - 1, T>::backward(sql, currentVersion - 1);
+                      Migration<prevVersion, T>::backward(sql, currentVersion - 1);
+                  } else if (currentVersion == 0) {
```
alekece: I am questioning myself about the type of `version`; why are we using `int` ? Do we allow negative version number ?
If not, I think you can change the type to any unsigned integer and change your `else if` in favor to a simple `else`.

ledger-core/inc/core/database/DatabaseSessionPool.hpp
```diff
@@ -75,20 +75,48 @@ namespace ledger {
...
-            void rollbackMigration() {
+            /// Run migration forward for the ledger-core library.
+            ///
+            /// This is a special case because we don’t want to register ourselves for rollback,
+            /// since we will call that function explicitly.
+            template <>
+            void forwardMigration<CoreMigration>() {
                 soci::session sql(getPool());
-                int version = getDatabaseMigrationVersion<T>(sql);
+                int version = getDatabaseMigrationVersion<CoreMigration>(sql);
```
alekece: Ditto

ledger-core/inc/core/database/DatabaseSessionPool.hpp
```diff
@@ -75,20 +75,48 @@ namespace ledger {
                 soci::transaction tr(sql);
                 Migration<T::CURRENT_VERSION, T>::forward(sql, version);
 
+                // register rollback migration
+                _rollbackMigrations.emplace_back([](soci::session& sql) {
+                    int version = getDatabaseMigrationVersion<T>(sql);
```
alekece: I missed this one; can you put also the `const` qualifier here ?

phaazon: You missed several actually. I’ll annotate them all.

ledger-core/inc/core/database/Migrations.hpp
phaazon: We use negative versions to encode non-existent data. We could use `0` instead, so I’ll make a switch.

alekece: Since the version is now an unsigned integer type and the first `if` statement checks all the values except zero; the `else if` should become and `else` because - in the current state of the code - the `if/else` statement suggests we can have another available values to check which is not the case.

PR #492, opened 2020-02-17
Sync token removal in reorganisation use case

KhalilBellakrid:
Unset sync token when falling in reorg situation because explorers are not trustworthy on that feature
**HODL:** I will add a config to deactivate sync tokens

### DIFF
--- a/core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
+++ b/core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
@@ -233,7 +233,7 @@ namespace ledger {
                 const auto deactivateToken =
                         buddy->configuration->getBoolean(api::Configuration::DEACTIVATE_SYNC_TOKEN).value_or(false);
                 auto getSyncToken = deactivateToken ? Future<void *>::successful(nullptr) : _explorer->startSession();
-                return getSyncToken.template map<Unit>(account->getContext(), [buddy, deactivateToken] (void * const& t) -> Unit {
+                return getSyncToken.template map<Unit>(account->getContext(), [buddy, deactivateToken] (void * const t) -> Unit {
                     buddy->logger->info("Synchronization token obtained");
                     if (!deactivateToken && t) {
                         buddy->token = Option<void *>(t);
@@ -340,7 +340,7 @@ namespace ledger {
                             return self->_explorer->startSession();
                         });
 
-                        return startSession.template flatMap<Unit>(ImmediateExecutionContext::INSTANCE, [=] (void * const &session) {
+                        return startSession.template flatMap<Unit>(ImmediateExecutionContext::INSTANCE, [=] (void * const session) {
                             if (!deactivateToken && session) {
                                 buddy->token = Option<void *>(session);
                             } else {
@@ -355,8 +355,8 @@ namespace ledger {
 
                             //Get its block/block height
                             auto &failedBatch = buddy->savedState.getValue().batches[currentBatchIndex];
-                            auto failedBlockHeight = failedBatch.blockHeight;
-                            auto failedBlockHash = failedBatch.blockHash;
+                            auto const failedBlockHeight = failedBatch.blockHeight;
+                            auto const failedBlockHash = failedBatch.blockHash;
 
                             if (failedBlockHeight > 0) {
                                 //Delete data related to failedBlock (and all blocks above it)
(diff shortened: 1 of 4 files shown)


core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -230,13 +230,21 @@ namespace ledger {
                 updateCurrentBlock(buddy, account->getContext());
 
                 auto self = getSharedFromThis();
-                return _explorer->startSession().template map<Unit>(account->getContext(), [buddy] (void * const& t) -> Unit {
+                const auto deactivateToken =
+                        buddy->configuration->getBoolean(api::Configuration::DEACTIVATE_SYNC_TOKEN).value_or(false);
+                auto getSyncToken = deactivateToken ? Future<void *>::successful(nullptr) : _explorer->startSession();
+                return getSyncToken.template map<Unit>(account->getContext(), [buddy, deactivateToken] (void * const& t) -> Unit {
```
alekece:
```suggestion
                return getSyncToken.template map<Unit>(account->getContext(), [buddy, deactivateToken] (void * const t) -> Unit {
```

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -322,52 +330,82 @@ namespace ledger {
...
+                        const auto deactivateToken =
+                                buddy->configuration->getBoolean(api::Configuration::DEACTIVATE_SYNC_TOKEN).value_or(false);
+                        auto startSession = Future<void *>::async(ImmediateExecutionContext::INSTANCE, [=](){
+                            if (deactivateToken) {
+                                return Future<void *>::successful(nullptr);
+                            }
+                            return self->_explorer->startSession();
+                        });
+
+                        return startSession.template flatMap<Unit>(ImmediateExecutionContext::INSTANCE, [=] (void * const &session) {
```
alekece:
```suggestion
                        return startSession.template flatMap<Unit>(ImmediateExecutionContext::INSTANCE, [=] (void * const session) {
```

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -322,52 +330,82 @@ namespace ledger {
...
 
-                            //Update savedState's batches
-                            for (auto &batch : buddy->savedState.getValue().batches) {
-                                if (batch.blockHeight > lastBlockHeight) {
-                                    batch.blockHeight = (uint32_t)lastBlockHeight;
-                                    batch.blockHash = lastBlockHash;
+                            //Get its block/block height
+                            auto &failedBatch = buddy->savedState.getValue().batches[currentBatchIndex];
+                            auto failedBlockHeight = failedBatch.blockHeight;
+                            auto failedBlockHash = failedBatch.blockHash;
```
alekece:
```suggestion
                            auto const failedBlockHeight = failedBatch.blockHeight;
                            auto const failedBlockHash = failedBatch.blockHash;
```

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -322,52 +330,82 @@ namespace ledger {
...
                                 }
-                            }
 
-                            //Save new savedState
-                            buddy->preferences->editor()->template putObject<BlockchainExplorerAccountSynchronizationSavedState>(
-                                    "state", buddy->savedState.getValue())->commit();
+                                //Update savedState's batches
+                                for (auto &batch : buddy->savedState.getValue().batches) {
+                                    if (batch.blockHeight > lastBlockHeight) {
+                                        batch.blockHeight = (uint32_t) lastBlockHeight;
```
alekece: We are performing a narrow cast which can result to an unexpected value.

I would like to introduce you `gsl` header-only library. `gsl::narrow_cast` performs some check during the cast and throws a `narrowing_error` if the resulting value is different from the uncasted value.

```suggestion
                                        batch.blockHeight = static_cast<uint32_t>(lastBlockHeight);
```

KhalilBellakrid: Really cool for `gsl` !! 
As discussed f2f, this is not something changed by this PR, Github is clearly messed up. but still the point is valid and should be treated in a dedicated PR for all our code ! I will create the corresponding ticket.

PR #494, opened 2020-02-18
Resync after wipe

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-531

### DIFF
--- a/core/src/wallet/ethereum/EthereumLikeAccount.cpp
+++ b/core/src/wallet/ethereum/EthereumLikeAccount.cpp
@@ -459,23 +459,16 @@ namespace ledger {
                 auto log = logger();
 
                 log->debug(" Start erasing data of account : {}", getAccountUid());
+
+                std::lock_guard<std::mutex> lock(_synchronizationLock);
+                _currentSyncEventBus = nullptr;
+
                 soci::session sql(getWallet()->getDatabase()->getPool());
+
                 //Update account's internal preferences (for synchronization)
-                auto savedState = getInternalPreferences()->getSubPreferences("BlockchainExplorerAccountSynchronizer")->getObject<BlockchainExplorerAccountSynchronizationSavedState>("state");
-                if (savedState.nonEmpty()) {
-                        //Reset batches to blocks mined before given date
-                        auto previousBlock = BlockDatabaseHelper::getPreviousBlockInDatabase(sql, getWallet()->getCurrency().name, date);
-                        for (auto& batch : savedState.getValue().batches) {
-                                if (previousBlock.nonEmpty() && batch.blockHeight > previousBlock.getValue().height) {
-                                        batch.blockHeight = (uint32_t) previousBlock.getValue().height;
-                                        batch.blockHash = previousBlock.getValue().blockHash;
-                                } else if (!previousBlock.nonEmpty()) {//if no previous block, sync should go back from genesis block
-                                        batch.blockHeight = 0;
-                                        batch.blockHash = "";
-                                }
-                        }
-                        getInternalPreferences()->getSubPreferences("BlockchainExplorerAccountSynchronizer")->editor()->putObject<BlockchainExplorerAccountSynchronizationSavedState>("state", savedState.getValue())->commit();
-                }
+                // Clear synchronizer state
+                eraseSynchronizerDataSince(sql, date);
+
                 auto accountUid = getAccountUid();
                 sql << "DELETE FROM operations WHERE account_uid = :account_uid AND date >= :date ", soci::use(accountUid), soci::use(date);
                 return Future<api::ErrorCode>::successful(api::ErrorCode::FUTURE_WAS_SUCCESSFULL);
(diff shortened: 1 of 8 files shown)


core/test/integration/synchronization/synchronization_tests.cpp
```diff
@@ -269,6 +269,120 @@ TEST_F(BitcoinLikeWalletSynchronization, TestNetSynchronization) {
...
+            pool->getEventBus()->subscribe(dispatcher->getMainExecutionContext(), receiver);
+
+            static std::function<void (int)> syncAccount = [accounts, receiver, this](int index){
+                auto localReceiver = make_receiver([=](const std::shared_ptr<api::Event> &event) {
+                    fmt::print("Received event {}\n", api::to_string(event->getCode()));
+                    if (event->getCode() == api::EventCode::SYNCHRONIZATION_STARTED)
+                        return;
+                    EXPECT_NE(event->getCode(), api::EventCode::SYNCHRONIZATION_FAILED);
+                    auto amount = wait(accounts[index]->getBalance());
+                    std::cout << "Amount: " << amount->toLong() << std::endl;
```
alekece: Just to be consistent with the rest of your test.

```suggestion
                    fmt::print("Amount: {}\n", amount->toLong());
```

core/test/integration/synchronization/synchronization_tests.cpp
```diff
@@ -269,6 +269,120 @@ TEST_F(BitcoinLikeWalletSynchronization, TestNetSynchronization) {
...
+                if (event->getCode() == api::EventCode::NEW_OPERATION) {
+                    auto uid = event->getPayload()->getString(
+                            api::Account::EV_NEW_OP_UID).value();
+                    EXPECT_EQ(emittedOperations.find(uid), emittedOperations.end());
+                }
+            });
+
+            pool->getEventBus()->subscribe(dispatcher->getMainExecutionContext(), receiver);
+
+            static std::function<void (int)> syncAccount = [accounts, receiver, this](int index){
```
alekece: The static storage isn't needed here.

```suggestion
            std::function<void (int)> syncAccount;
            
            syncAccount = [accounts, receiver, this](int index){
 
```

core/test/integration/synchronization/synchronization_tests.cpp
```diff
@@ -269,6 +269,120 @@ TEST_F(BitcoinLikeWalletSynchronization, TestNetSynchronization) {
...
+                    EXPECT_EQ(emittedOperations.find(uid), emittedOperations.end());
+                }
+            });
+
+            pool->getEventBus()->subscribe(dispatcher->getMainExecutionContext(), receiver);
+
+            auto date = "2000-03-27T09:10:22Z";
+            auto formatedDate = DateUtils::fromJSON(date);
+
+            static std::function<void (bool)> syncAccount = [formatedDate, account, this](bool shouldStop) {
```
alekece: Ditto

core/test/integration/synchronization/synchronization_tests.cpp
```diff
@@ -269,6 +269,120 @@ TEST_F(BitcoinLikeWalletSynchronization, TestNetSynchronization) {
...
+                auto localReceiver = make_receiver([=](const std::shared_ptr<api::Event> &event) {
+                    fmt::print("Received event {}\n", api::to_string(event->getCode()));
+                    if (event->getCode() == api::EventCode::SYNCHRONIZATION_STARTED)
+                        return;
+                    EXPECT_NE(event->getCode(), api::EventCode::SYNCHRONIZATION_FAILED);
+                    auto amount = wait(account->getBalance());
+                    std::cout << "Amount: " << amount->toLong() << std::endl;
```
alekece: Ditto

core/src/wallet/ethereum/EthereumLikeAccount.cpp
```diff
@@ -459,23 +459,16 @@ namespace ledger {
                 auto log = logger();
 
                 log->debug(" Start erasing data of account : {}", getAccountUid());
+
+                std::lock_guard<std::mutex> lock(_synchronizationLock);
```
phaazon: This lock mechanism effort is duplicated for all currencies. It can be easy to make a mistake or forget about it.
Maybe the `AbstractAccount::eraseDataSince` could have that logic instead of moving it down to the coin? Even though `AbstractAccount::eraseDataSince` is virtual pure, you can still provide it an implementation that coins could call in the implementation?

core/test/integration/synchronization/synchronization_tests.cpp
KhalilBellakrid: As discussed f2f: it's used for convenience and more readability

core/src/wallet/ethereum/EthereumLikeAccount.cpp
KhalilBellakrid: We agree but this needs moving the lock and event bus on abstract level, the impact could be huge, we should have a dedicated PR for this kind of changes (even should be part of segregation future tasks) !

PR #496, opened 2020-02-18
Backport PostgreSQL support 

alekece:
As the title above said this PR contains the backport of PostSQL support into the segregated `lib-ledger-core` and a method to know if the main `ledger-core` library was compiled with `PG_SUPPORT` (efficient way to instance the correct database backend in coin-specific library).

## Mandatory picture of a cat 
pink_cat

### DIFF
--- a/ledger-core-ripple/test/CMakeLists.txt
+++ b/ledger-core-ripple/test/CMakeLists.txt
@@ -30,4 +30,4 @@ add_test (NAME Ripple-address COMMAND ledger-core-ripple-tests --gtest_filter=Ri
 add_test (NAME Ripple-keychain COMMAND ledger-core-ripple-tests --gtest_filter=RippleKeychains.*)
 add_test (NAME Ripple-transaction COMMAND ledger-core-ripple-tests --gtest_filter=RippleMakeTransaction.*)
 add_test (NAME Ripple-sync COMMAND ledger-core-ripple-tests --gtest_filter=RippleLikeWalletSynchronization.*)
-add_test (NAME Ripple-account COMMAND ledger-core-ripple-test --gtest_filter=RippleAccounts.*)
+add_test (NAME Ripple-account COMMAND ledger-core-ripple-tests --gtest_filter=RippleAccounts.*)
--- a/ledger-core/src/core/database/DatabaseSessionPool.cpp
+++ b/ledger-core/src/core/database/DatabaseSessionPool.cpp
@@ -31,6 +31,10 @@
 
 #include <core/database/DatabaseSessionPool.hpp>
 #include <core/database/Migrations.hpp>
+#ifdef PG_SUPPORT
+#include <core/database/PostgreSQLBackend.hpp>
+#endif
+
 
 namespace ledger {
     namespace core {
@@ -39,8 +43,8 @@ namespace ledger {
             const std::shared_ptr<api::PathResolver> &resolver,
             const std::shared_ptr<spdlog::logger>& logger,
             const std::string &dbName,
-            const std::string &password
-        ) : _pool((size_t) backend->getConnectionPoolSize()), _backend(backend), _buffer("SQL", logger) {
+            const std::string &password) : 
+                _pool((size_t) backend->getConnectionPoolSize()), _backend(backend), _buffer("SQL", logger) {
             if (logger != nullptr && backend->isLoggingEnabled()) {
                 _logger = new std::ostream(&_buffer);
             } else {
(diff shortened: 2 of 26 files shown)


ledger-core-ripple/test/CMakeLists.txt
```diff
@@ -30,4 +30,4 @@ add_test (NAME Ripple-address COMMAND ledger-core-ripple-tests --gtest_filter=Ri
 add_test (NAME Ripple-keychain COMMAND ledger-core-ripple-tests --gtest_filter=RippleKeychains.*)
 add_test (NAME Ripple-transaction COMMAND ledger-core-ripple-tests --gtest_filter=RippleMakeTransaction.*)
 add_test (NAME Ripple-sync COMMAND ledger-core-ripple-tests --gtest_filter=RippleLikeWalletSynchronization.*)
-add_test (NAME Ripple-account COMMAND ledger-core-ripple-test --gtest_filter=RippleAccounts.*)
\ No newline at end of file
+add_test (NAME Ripple-account COMMAND ledger-core-ripple-tests --gtest_filter=RippleAccounts.*)
```
phaazon: :)

ledger-core/src/core/database/DatabaseSessionPool.cpp
```diff
@@ -39,8 +43,8 @@ namespace ledger {
             const std::shared_ptr<api::PathResolver> &resolver,
             const std::shared_ptr<spdlog::logger>& logger,
             const std::string &dbName,
-            const std::string &password
-        ) : _pool((size_t) backend->getConnectionPoolSize()), _backend(backend), _buffer("SQL", logger) {
+            const std::string &password) :
```
phaazon: Are you kidding me.

ledger-core/src/core/database/DatabaseSessionPool.cpp
```diff
@@ -55,6 +59,12 @@ namespace ledger {
                     session.set_log_stream(_logger);
             }
 
+#ifdef PG_SUPPORT
+            _type = std::dynamic_pointer_cast<PostgreSQLBackend>(backend) != nullptr ?
+                    api::DatabaseBackendType::POSTGRESQL : api::DatabaseBackendType::SQLITE3;
```
phaazon: This is okay for this version but potentially ill-designed, as if we at some time support three different backends, that will just be wrong.

ledger-core/src/core/database/PostgreSQLBackend.cpp
```diff
@@ -0,0 +1,76 @@
...
+                                  const std::string &dbName,
+                                  const std::string &password,
+                                  soci::session &session) {
+            _dbName = dbName;
+            setPassword(password, session);
+        }
+
+        void PostgreSQLBackend::setPassword(const std::string &password,
+                                         soci::session &session) {
+            if (_dbName.empty()) {
```
phaazon: That’s the kind of invariant that should be encoded at compile-time, not runtime. Kind of things I would like us to tackle when we have time to redesign the lib. :)

ledger-core/src/core/database/PostgreSQLBackend.cpp
```diff
@@ -0,0 +1,76 @@
...
+                throw make_exception(api::ErrorCode::DATABASE_EXCEPTION, "Database should be initiated before setting password.");
+            }
+            session.close();
+            session.open(*soci::factory_postgresql(), _dbName);
+        }
+
+        void PostgreSQLBackend::changePassword(const std::string & oldPassword,
+                                            const std::string & newPassword,
+                                            soci::session &session) {
+            throw make_exception(api::ErrorCode::IMPLEMENTATION_IS_MISSING, "Change of password is not handled with PostgreSQL backend.");
```
phaazon: So… not sure this is the right way to do. That would result in either a crash of the client product, or a necessary to know that this function can throw an exception (which is hard to see). As this is the modularization, this is fine, but I wish I saw that earlier…

phaazon:
BTW, have you tried it?

ledger-core/src/core/database/DatabaseSessionPool.cpp
alekece: Alright alright I just copy-pasted changes from PR (even if there are none modifications here).

Let's `clang-format` to that stuff for us.

alekece:
All the tests are green either with SQLite and PostgreSQL.

I kinda agree with all your comments - this changes comes from a validated PR so I suggest to let it just like that but create some ticket to treat what you reveal :)

ledger-core/src/core/database/DatabaseSessionPool.cpp
phaazon: Yeah, we need to introduce it! :)

ledger-core/src/core/database/DatabaseSessionPool.cpp
phaazon: See LLC-570.

ledger-core/src/core/database/PostgreSQLBackend.cpp
phaazon: LLC-571.

ledger-core/src/core/database/PostgreSQLBackend.cpp
phaazon: Already in LLC-513.

PR #498, opened 2020-02-21
Add support for failover behavior in soci connection pools.

phaazon:
With the current version of soci, when a `connection_pool` is created, a
buffer is allocated with, for each element in the buffer, a pair `(bool,
session*)`. The boolean value indicates whether the session is free or
not (this is pretty bad way to encode a pool, but whatever). When a
session is freed, the boolean gets changed and then opening a new
session will reuse that slot.

However, when a session has been idleing for a while, it’s possible that
the database backend (e.g. PostgreSQL) kills the connection held in the
session for timeout. The `soci` library doesn’t see that and assumes the
session is still valid -> we hit errors, which get propagated as
`std::runtime_error` in our codebase.

This commit fixes the problem by introducing a `is_alive()` function on
the session that allows a database backend to execute some logic to
check the status of the session before using it. SQLite3 doesn’t drop
sessions (no timeout) but PostgreSQL does), so we can implement that
logic in the PostgreSQL backend.

It should be fix LVS-92.

### DIFF
--- a/core/idl/database/database.djinni
+++ b/core/idl/database/database.djinni
@@ -205,6 +205,9 @@ DatabaseConnection = interface +j +n +o {
     # Create a new empty blob.
     # @return An empty blob
     newBlob(): DatabaseBlob;
+
+    # Check whether the connection is still alive.
+    is_alive(): bool;
 }
 
 # A pool of connections to a single database.
--- a/core/idl/database/database.djinni
+++ b/core/idl/database/database.djinni
@@ -207,7 +207,7 @@ DatabaseConnection = interface +j +n +o {
     newBlob(): DatabaseBlob;
 
     # Check whether the connection is still alive.
-    is_alive(): bool;
+    isAlive(): bool;
 }
 
 # A pool of connections to a single database.
--- a/core/lib/soci/core/session.cpp
+++ b/core/lib/soci/core/session.cpp
@@ -80,12 +80,20 @@ session::session(std::string const & connectString)
 session::session(connection_pool & pool)
     : logStream_(NULL), isFromPool_(true), pool_(&pool)
 {
-    poolPosition_ = pool.lease();
-    session & pooledSession = pool.at(poolPosition_);
+    while (true) {
+        poolPosition_ = pool.lease();
+        session & pooledSession = pool.at(poolPosition_);
 
-    once.set_session(&pooledSession);
-    prepare.set_session(&pooledSession);
-    backEnd_ = pooledSession.get_backend();
+        if (pooledSession.is_alive()) {
+            once.set_session(&pooledSession);
+            prepare.set_session(&pooledSession);
+            backEnd_ = pooledSession.get_backend();
+            break;
+        } else {
+            // invalidate the session in the pool, and try again
+            pool.give_back(poolPosition_);
+        }
+    }
 }
 
 session::~session()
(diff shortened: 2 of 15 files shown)


core/idl/database/database.djinni
```diff
@@ -205,6 +205,9 @@ DatabaseConnection = interface +j +n +o {
     # Create a new empty blob.
     # @return An empty blob
     newBlob(): DatabaseBlob;
+
+    # Check whether the connection is still alive.
+    is_alive(): bool;
```
alekece: `is_alive` to `isAlive` to respect the codebase naming convention

core/lib/soci/core/session.cpp
```diff
@@ -80,12 +80,20 @@ session::session(std::string const & connectString)
...
+        if (pooledSession.is_alive()) {
+            once.set_session(&pooledSession);
+            prepare.set_session(&pooledSession);
+            backEnd_ = pooledSession.get_backend();
+            break;
+        } else {
+            // invalidate the session in the pool, and try again
+            pool.give_back(poolPosition_);
+        }
+    }
```
alekece: I suggest to create a new configuration value such as `MAX_RETRY_SESSION_CONNECTION` or something similar to avoid an infinity loop. 
Even if you dislike exception the case where we all the tries failed can be considered as exceptional.

core/lib/soci/core/session.cpp
```diff
@@ -388,3 +396,10 @@ blob_backend * session::make_blob_backend()
 
     return backEnd_->make_blob_backend();
 }
+
+bool session::is_alive() const
```
alekece: Ditto on the method name

core/lib/soci/core/session.h
```diff
@@ -122,6 +122,12 @@ class SOCI_DECL session
     details::rowid_backend * make_rowid_backend();
     details::blob_backend * make_blob_backend();
 
+    // check the status of a session; some backends might drop idleing connections,
+    // resulting in invalidated connection_pool; when that occurs, this function
+    // should return false to indicate the connection_pool to discard the session
+    // and try again; the default implementation assumes no timeout
+    bool is_alive() const;
```
alekece: Ditto

core/lib/soci_postgresql/session.cpp
```diff
@@ -128,3 +128,14 @@ postgresql_blob_backend * postgresql_session_backend::make_blob_backend()
 {
     return new postgresql_blob_backend(*this);
 }
+
+bool postgresql_session_backend::is_alive() const {
+    auto res = PQexec(conn_, "SELECT 1");
```
alekece: Add `const` qualifier to that variable

core/lib/soci/core/session.cpp
phaazon: I agree.

alekece: Since the change is `soci` related maybe the configuration value cannot be applied in that scope

phaazon:
@alekece about your configuration suggestion, it could be, actually. We can just accept that as a `DATABASE_FAILOVER_RETRY` generic argument we use when connecting to our backends, and backends are free to use it or ignore it.

core/lib/soci_postgresql/session.cpp
phaazon: I can’t, because `PQresult` expects a `PQconn*`. I don’t want `const_cast` here, especially because `res` won’t be used anymore after that call.

phaazon:
Merging as all release tests pass and debug fail on explorers gateways errors.

PR #501, opened 2020-02-27
Fix OpenSSL’s BLAKE2b heap memory corruption.

phaazon:
The problem was that, in OpenSSL 1.1, they have completely integrated
BLAKE2b via the EVP interface, which is to be used — recommended by
OpenSSL developers I asked help to, and the actual documentation
provides link to _only_ `EVP_blake2b512` and `EVP_blake2s256`.

However, the EVP interface (and private code) of OpenSSL prevents
anybody from using length-variable digests. It’s an API regression that
has no workaround — I tried to browse the EVP interface, but there’s
nothing to do.

The fix is to vendor our own version of BLAKE2b. I put it in the
`ledger::core::hacky` namespace because otherwise, `node` would find
those symbols and replace them with the regression. Yeah I know. I
suffered.

Fixes LLC-566.


core/src/crypto/BLAKE.cpp
```diff
@@ -55,11 +59,32 @@ namespace ledger {
...
             std::vector<uint8_t> hash(outLength);
-            BLAKE2B_CTX blakeState;
             BLAKE2b_Init_default(&blakeState, outLength);
             BLAKE2b_Update(&blakeState, data.data() + offset, (data.size() - offset));
             BLAKE2b_Final(hash.data(), &blakeState);
+#else
+            // The OpenSSL official documentation (and developers too) recommend to use the
+            // EVP interface. The code for < 1.1 above was required because OpenSSL didn’t have
+            // direct support of it
+            EVP_MD_CTX * mdctx = EVP_MD_CTX_new();
```
teams2ua: Is this resource allocation, if yes were should we free it?
I know this is urgent, so put TODO to wrap it to smart pointer to have automatic resources handling.

phaazon: Yeah I’ll fix that in the PR too, it should be removed by the end of the function. We’ll need to find a better scheme later not to do those dynamic allocation on the fly, as they might be costly.

teams2ua: Yeap the same staff I tried to merge in https://github.com/LedgerHQ/lib-ledger-core/pull/307

PR #502, opened 2020-02-27
Fix build without PG_SUPPORT

alekece:
This PR aims to resolve a new bug occuring since the backport of PostgreSQL - quick and easy.

### DIFF
--- a/ledger-core/CMakeLists.txt
+++ b/ledger-core/CMakeLists.txt
@@ -31,8 +31,33 @@ set(CMAKE_MACOSX_RPATH 1)
 
 add_definitions("-DSQLITE_HAS_CODEC")
 
+set(INSTALL_TARGETS
+    ledger-core
+    ledger-core-static
+    ledger-core-obj
+    ledger-core-interface
+    bigd
+    fmt
+    crypto
+    soci_sqlite3
+    soci_postgresql
+    soci_core_static
+    leveldb
+    spdlog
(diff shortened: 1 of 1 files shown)


ledger-core/CMakeLists.txt
```diff
@@ -31,8 +31,33 @@ set(CMAKE_MACOSX_RPATH 1)
...
+    ledger-qt-host
+    ledger-core-integration-test
+    mongoose
+    ssl
+    gtest gtest_main
+)
+
 if (PG_SUPPORT)
     add_definitions("-DPG_SUPPORT")
+else()
```
phaazon: I’m not sure this is the right way to do. That seems odd.

gagbo: I would prefer adding to the list in the `if (PG_SUPPORT)` branch too.

ledger-core/CMakeLists.txt
```diff
@@ -31,6 +31,10 @@ set(CMAKE_MACOSX_RPATH 1)
 
 add_definitions("-DSQLITE_HAS_CODEC")
 
+set(INSTALL_TARGETS
+
+)
+
```
gagbo: What's the purpose of this variable ? Does it get filled automatically when calling `install` later ?

alekece: Good catch ! I will remove it :trollface:

PR #504, opened 2020-03-02
Do not include PG header if we’re not compiling with PG_SUPPORT.

phaazon:


ledger-core/src/CMakeLists.txt
```diff
@@ -159,7 +159,12 @@ target_include_directories(
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/lib/ethash/src>           $<INSTALL_INTERFACE:include/ledger-core/lib/ethash/src>
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/lib/fmt/include>          $<INSTALL_INTERFACE:include/ledger-core/lib/fmt/include>
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/lib/spdlog/include>       $<INSTALL_INTERFACE:include/ledger-core/lib/spdlog/include>
-    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/lib/soci_postgresql>      $<INSTALL_INTERFACE:include/ledger-core/lib/soci_postgresql>
+
+    # PG support
+    $<$<BOOL:${PG_SUPPORT}>:
+      $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/lib/soci_postgresql>
+      $<INSTALL_INTERFACE:include/ledger-core/lib/soci_postgresql>
```
teams2ua: Why do we need postgresql headers when we install library?

phaazon: That’s a very good question. The only reason I see is “if it’s needed in the coin projects.” It shouldn’t be, so very good catch. I’ll update.

PR #510, opened 2020-03-03
Better granularity of HTTP errors

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/LLC-586

### DIFF
--- a/core/src/net/HttpClient.hpp
+++ b/core/src/net/HttpClient.hpp
@@ -101,6 +101,21 @@ namespace ledger {
             std::shared_ptr<api::ExecutionContext> _context;
             Option<std::shared_ptr<spdlog::logger>> _logger;
 
+            static api::ErrorCode getErrorCode(int32_t statusCode) {
+                if (statusCode >= 200 && statusCode < 300 ) {
+                    return api::ErrorCode::FUTURE_WAS_SUCCESSFULL;
+                }
+                return statusCode >= 500 ? api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST :
+                statusCode >= 400 ? api::ErrorCode::HTTP_ERROR :
+                api::ErrorCode::TOO_MANY_REDIRECT;
+            };
+
+            static bool isHttpError(api::ErrorCode errorCode) {
+                return errorCode == api::ErrorCode::HTTP_ERROR ||
+                errorCode == api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST ||
+                errorCode == api::ErrorCode::TOO_MANY_REDIRECT;
+            };
+
             class ApiRequest : public api::HttpRequest {
             public:
                 ApiRequest(const std::shared_ptr<const ledger::core::HttpRequest>& self);
--- a/core/src/net/HttpClient.hpp
+++ b/core/src/net/HttpClient.hpp
@@ -102,10 +102,8 @@ namespace ledger {
             Option<std::shared_ptr<spdlog::logger>> _logger;
 
             static api::ErrorCode getErrorCode(int32_t statusCode) {
-                if (statusCode >= 200 && statusCode < 300 ) {
-                    return api::ErrorCode::FUTURE_WAS_SUCCESSFULL;
-                }
-                return statusCode >= 500 ? api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST :
+                return statusCode >= 200 && statusCode < 300 ? api::ErrorCode::FUTURE_WAS_SUCCESSFULL :
+                statusCode >= 500 ? api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST :
                 statusCode >= 400 ? api::ErrorCode::HTTP_ERROR :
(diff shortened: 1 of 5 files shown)


core/src/net/HttpClient.hpp
```diff
@@ -101,6 +101,21 @@ namespace ledger {
             std::shared_ptr<api::ExecutionContext> _context;
             Option<std::shared_ptr<spdlog::logger>> _logger;
 
+            static api::ErrorCode getErrorCode(int32_t statusCode) {
+                if (statusCode >= 200 && statusCode < 300 ) {
+                    return api::ErrorCode::FUTURE_WAS_SUCCESSFULL;
+                }
+                return statusCode >= 500 ? api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST :
+                statusCode >= 400 ? api::ErrorCode::HTTP_ERROR :
+                api::ErrorCode::TOO_MANY_REDIRECT;
```
alekece: Nested ternary statements are really hard to read and reason about; A main rule is to avoid to add cognitive complexity for the sake a smaller code.

```suggestion
                if (statusCode >= 200 && statusCode < 300 ) {
                    return api::ErrorCode::FUTURE_WAS_SUCCESSFULL;
                }
                else if (statusCode >= 500) {
                    return api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST;
                }
                else if (statusCoder >= 400) {
                    return api::ErrorCode::HTTP_ERROR :
                }
                else {
                    return api::ErrorCode::TOO_MANY_REDIRECT;
                }
```

core/src/net/HttpClient.hpp
```diff
@@ -101,6 +101,21 @@ namespace ledger {
...
+            };
+
+            static bool isHttpError(api::ErrorCode errorCode) {
+                return errorCode == api::ErrorCode::HTTP_ERROR ||
+                errorCode == api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST ||
+                errorCode == api::ErrorCode::TOO_MANY_REDIRECT;
```
alekece: Ditto

core/src/net/HttpClient.hpp
phaazon: I kinda disagree there. They read pretty easily:

```cpp
auto a = ifOkay ? dothis : insteadIfThis ? dothat : orJustDoThis;
```

Treating conditionals as expresisons is very natural to me (as it’s the native way to build conditionals in functional programming languages). It’s almost closer to how we speak. The change you suggest is very verbose for really nothing. Cognitive complexity is a thing, but structural noise is also something to take into account, and here, there is too much noise.

core/src/net/HttpClient.hpp
phaazon: Nah. Boolean chaining will yield better code than branching. That’s a correct usage of boolean algebra.

core/src/net/HttpClient.hpp
alekece: Discussion about native way to build multiple conditional statements in functional programming compare imperative one is out of scope.
In my opinion nested (and really packed) ternary statements can alway be express in a different and cleaner way; the one I suggest add 6 lines of __noises__ and allow the addition of future condition statements.

core/src/net/HttpClient.hpp
alekece: Agree. I made a jugement mistake here (I was stuck in ternary expression stuff)

core/src/net/HttpClient.hpp
KhalilBellakrid: @alekece ternary operators are made to make code more readable and are preferred to be used when you are setting or returning values depending on a simple condition, this is exactly the case here !

phaazon: > Discussion about native way to build multiple conditional statements in functional programming compare to imperative one is out of scope.

It’s not out of scope. Using a language paradigm from one language to another can be useful if it reduces noise and, by extent, maintainability and readability. You’re suggesting to add 6 lines of noise for something that can be expressed _almost as if we spoke it out loud_. I prefer to have a single `return <expr>` rather than several `return`. In that sense, yes, functional programming is beneficial here and C++ (and C) has been supporting that for decades.

alekece: Our opinion diverges on the usage of ternary statement; 
If you still want to use nested ternary statement I suggest to remove the additional `if` on top of the nested ternary statement such as :
```suggestion
                return statusCode >= 200 && statusCode < 300 ? api::ErrorCode::FUTURE_WAS_SUCCESSFULL : 
                statusCode >= 500 ? api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST :
                statusCode >= 400 ? api::ErrorCode::HTTP_ERROR :
                api::ErrorCode::TOO_MANY_REDIRECT;
```

KhalilBellakrid: As said before, I was using that for "simple conditions" but we can make it like that as it remains a pretty simple condition.

PR #512, opened 2020-03-05
Fix CMake exports so that we don’t export non-existing targets.

phaazon:
# Content

This PR moves _targets_ around so that we don’t include them if we’re not making them in the first place.

# Mandatory picture of an animal

### DIFF
--- a/ledger-core/CMakeLists.txt
+++ b/ledger-core/CMakeLists.txt
@@ -91,7 +91,6 @@ install(
         ledger-core-interface
         bigd
         fmt
-        crypto
         blake
         soci_sqlite3
         soci_core_static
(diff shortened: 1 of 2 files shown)


ledger-core/CMakeLists.txt
```diff
@@ -91,27 +91,57 @@ install(
...
+        EXPORT ledger-core
+        ARCHIVE       DESTINATION lib/ledger-core
+        LIBRARY       DESTINATION lib/ledger-core
+        FRAMEWORK     DESTINATION lib/ledger-core
+    )
+
+    if (SSL_SUPPORT)
+        install(
+            TARGETS
+                crypto
```
phaazon: Whoopsies, it should be `ssl`.

PR #513, opened 2020-03-06
Fix the Tezos’ fees field (actually fee) when getting fees from the external API.

phaazon:
Fixes LLC-589.

### DIFF
--- a/core/idl/wallet/tezos/tezos_like_wallet.djinni
+++ b/core/idl/wallet/tezos/tezos_like_wallet.djinni
@@ -154,10 +154,9 @@ TezosLikeOriginatedAccount = interface +c {
 }
 
 TezosConfiguration = interface +c {
-  const TEZOS_XPUB_CURVE: string = "TEZOS_XPUB_CURVE";
+  	const TEZOS_XPUB_CURVE: string = "TEZOS_XPUB_CURVE";
 	const TEZOS_PROTOCOL_UPDATE: string = "TEZOS_PROTOCOL_UPDATE";
 	const TEZOS_NODE: string = "TEZOS_NODE";
-  const TZSTATS_API_ENDPOINT: string = "TZSTATS_API_ENDPOINT";
 }
 
 TezosConfigurationDefaults = interface +c {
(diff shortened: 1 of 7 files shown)


phaazon:
@KhalilBellakrid sure! :)

gre:
just to be clear. this will not be complete in term of fixes needed. see https://ledgerhq.atlassian.net/browse/LLC-591

KhalilBellakrid:
@phaazon Another thing, it seems that even by fixing this, there are some issues on some amounts of txs etc ... We will have to investigate that because it might be related to same issue, what we can do is build a local version with this fix and run `live-common`'s tests and we'll realize the regressions !

core/idl/wallet/tezos/tezos_like_wallet.djinni
```diff
@@ -154,9 +154,10 @@ TezosLikeOriginatedAccount = interface +c {
 }
 
 TezosConfiguration = interface +c {
-  	const TEZOS_XPUB_CURVE: string = "TEZOS_XPUB_CURVE";
+  const TEZOS_XPUB_CURVE: string = "TEZOS_XPUB_CURVE";
 	const TEZOS_PROTOCOL_UPDATE: string = "TEZOS_PROTOCOL_UPDATE";
 	const TEZOS_NODE: string = "TEZOS_NODE";
+  const TZSTATS_API_ENDPOINT: string = "TZSTATS_API_ENDPOINT";
```
KhalilBellakrid: Why do we need this ? I think for comparaison you'll have to retrieve: `BLOCKCHAIN_EXPLORER_API_ENDPOINT` and compare it against default `TZSTATS_API_ENDPOINT` 🤔

phaazon:
> just to be clear. this will not be complete in term of fixes needed. see https://ledgerhq.atlassian.net/browse/LLC-591

Yes, that PR is only about the `fees` situation! :)

core/idl/wallet/tezos/tezos_like_wallet.djinni
phaazon: Oh you are right. Nice catch!

PR #516, opened 2020-03-09
Fix ssl not being properly imported whith BUILD_TESTS=ON and no SYS_OPENSSL=OFF.

phaazon:

### DIFF
--- a/ledger-core/CMakeLists.txt
+++ b/ledger-core/CMakeLists.txt
@@ -130,7 +130,7 @@ if(NOT SYS_OPENSSL)
         FRAMEWORK     DESTINATION lib/ledger-core
     )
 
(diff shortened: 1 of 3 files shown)


ledger-core/CMakeLists.txt
```diff
@@ -130,7 +130,7 @@ if(NOT SYS_OPENSSL)
         FRAMEWORK     DESTINATION lib/ledger-core
     )
 
-    if (SSL_SUPPORT)
+    if (SSL_SUPPORT OR BUILD_TESTS)
```
alekece: We need ssl library for our tests ?

phaazon: Yes, for HTTPS requests.

PR #518, opened 2020-03-10
Fix Http client with JSON SAX parser in case of error

pollastri-pierre:
- Apply changed made on HttpClient.cpp to HttpClient.hpp


core/src/net/HttpClient.hpp
```diff
@@ -73,7 +73,8 @@ namespace ledger {
             template <typename Success, typename Failure, typename Handler>
             Future<Either<Failure, std::shared_ptr<Success>>> json(Handler handler) const {
                 return operator()().recover(_context, [] (const Exception& exception) {
-                    if (exception.getErrorCode() == api::ErrorCode::HTTP_ERROR && exception.getUserData().nonEmpty()) {
+                    if (HttpRequest::isHttpError(exception.getErrorCode()) &&
```
phaazon: Nice catch.

PR #522, opened 2020-03-12
Bitcoin: add method to get addresses from range

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/BACK-527


core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -260,6 +260,16 @@ namespace ledger {
             }
         }
 
+        std::vector<std::shared_ptr<api::Address>> BitcoinLikeAccount::fromBitcoinAddressesToAddresses(const std::vector<std::shared_ptr<BitcoinLikeAddress>> &addresses) {
+            AbstractAccount::AddressList result(addresses.size());
+            auto i = 0;
+            for (auto& addr : addresses) {
+                result[i] = std::dynamic_pointer_cast<api::Address>(addr);
+                i += 1;
+            }
```
phaazon: I think you can simplify that function by using `reserve` and `push_back`:

```suggestion
AbstractAccount::AddressList result;
result.reserve(addresses.size());

for (auto const& addr : addresses) {
  result.push_back(std::dynamic_pointer_cast<api::Address>(addr);
}
```

That way, you don’t have to keep the logic of this `i` variable, and you still make only one allocation.

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -260,6 +260,16 @@ namespace ledger {
...
+            return result;
```
alekece:
```suggestion
            AbstractAccount::AddressList result(addresses.size());
            
            std::transform(std::begin(addresses), std::end(addresses), std::begin(result), [](auto address) {
                return std::dynamic_pointer_cast<api::Address>(address);
            });

            return result;
```

PR #429, opened 2019-12-19
[ETH] Use corrected endpoint for estimation of GasLimit

gagbo:
----

[HODL] This passes test but is obviously wrong.

Enabling modification of "amplifier", "from" and "data" fields in the
underlying explorer API call means the interface of the explorer needs
to change. Evaluation of impact on bindings is necessary before going on.

PR #523, opened 2020-03-12
Add missing explorer tests

alekece:
This PR includes missing explorer tests (basically linked to ethereum) and implement a `ExplorerFixture` class which is used in bitcoin and ethereum tests.
We need to merge this PR as soon as possible to unlock the backport of the PR #429.

## Mandatory picture of a cat
magic_cat

### DIFF
--- a/ledger-core-ethereum/test/ExplorerTests.cpp
+++ b/ledger-core-ethereum/test/ExplorerTests.cpp
@@ -1,0 +1,57 @@
+/*
+ *
+ * ExplorerTests
+ *
+ * ledger-core-ethereum
+ *
+ * Created by Alexis Le Provost on 04/02/2020.
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ledger
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


ledger-core-ethereum/test/ExplorerTests.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <gtest/gtest.h>
+
+#include <ethereum/EthereumNetworks.hpp>
+#include <ethereum/api/EthereumLikeNetworkParameters.hpp>
+#include <ethereum/explorers/LedgerApiEthereumLikeBlockchainExplorer.hpp>
+
+#include <integration/ExplorerFixture.hpp>
+
+class EthereumExplorers : public ExplorerFixture<LedgerApiEthereumLikeBlockchainExplorer, api::EthereumLikeNetworkParameters> {
```
phaazon: Why plural?

alekece: @phaazon To match with naming convention using in other tests.

phaazon: Do you have an example? I don’t find what you’re talking about.

alekece: Screenshot 2020-03-12 at 17 12 03
Screenshot 2020-03-12 at 17 12 32

phaazon: Thanks! That’s a weird convention but I agree with your decision. Good call.

PR #531, opened 2020-03-23
[LLC-607] Add staking-related operation types

gagbo:


core/idl/wallet/common/wallet.djinni
```diff
@@ -39,10 +39,20 @@ DerivationPath = interface +c {
...
 #
-# It’s possible to either send funds or receive ones.
+# It’s possible to either
+# send funds (getAmount() returns the sent amount),
+# receive funds (getAmount() returns the sent amount),
+# delegate funds (getAmount() returns the *delegated* amount) ; only getFees() are part of balanceHistory
+# undelegate funds (getAmount() returns the *undelegated* amount) ; only getFees() are part of balanceHistory
+# claim rewards (getAmount() returns 0 or the known amount of claimed rewards) ; only getFees() are part of balanceHistory
+#
+# The last 3 operations should not expect to change balanceHistory with their amount, but only their fees.
```
phaazon, line 49: Can you also add a small explanation of the `none` situation? Thanks!

gagbo: Will do

gagbo:
Added the NONE comment, merge when you see fit :)

phaazon:
@gagbo Before merging, I have a question. How does that relate to computing balance histories? Especially, a delegation (on Tezos for instance) is an operation that still is going to drain some coins (so it’s like a `send`). Or do you mean we should create a new type of operation out of that transaction that will have zero fees / amount, but just witness the operation?

gagbo:
The 2 approaches are possible actually, and I suppose we need to discuss with Live team for that.

- Either we can have a single `Delegate` Operation with some fees (through [getFees](https://github.com/LedgerHQ/lib-ledger-core/blob/d5e96ac9db3f1adfd0e5a85039ba25398a24b95f/core/idl/wallet/common/wallet.djinni#L117) ) ; I don't like this approach much because, in the case of Cosmos, we might have multiple operations on a single transaction, so how are we supposed to divide the common fees on the 2/3/4 operations in the transaction ?

- Or we can have a `Delegate` operation with no fee involved at all, and a separate `Send` operation for the fees associated with the tx (if the user paid for it). We can include the fees as the amount of the operation in the `Send` part (or in the fees, anything works here)


In both cases, the [amount](https://github.com/LedgerHQ/lib-ledger-core/blob/develop/core/idl/wallet/common/wallet.djinni#L114) of the `Delegate` operation is expected to be the actual amount of the delegation/vote, and not impact the balance computation at all.

phaazon:
LGTM, thanks for explaining!

PR #534, opened 2020-03-25
Static analyzer supports

alekece:
This PR introduces the following static analyzer tools :
* [cppcheck](http://cppcheck.sourceforge.net/)
* [clang-tidy](https://clang.llvm.org/extra/clang-tidy/)
* [include-what-you-use](https://include-what-you-use.org/)

The static analyzers can be combined together and should be enabled at `cmake` step such as : 
```
cmake -DENABLE_CPPCHECK=ON -DENABLE_CLANG_TIDY=ON -DENABLE_IWYU=ON ...
``` 

## Mandatory picture of a cat
giphy


gagbo:
Cool ! `ccache` next step ? :eyes:

alekece:
@gagbo Exactly then sanitizers :+1:

ledger-core-bitcoin/.clang-tidy
```diff
@@ -0,0 +1 @@
+../.clang-tidy
```
gagbo: Do we have to add the symlink here ? A root `.clang-tidy` should be enough isn't it ?

ledger-core-bitcoin/src/CMakeLists.txt
```diff
@@ -146,20 +87,21 @@ if(IS_IOS GREATER_EQUAL 0)
...
 else()
-    add_library(ledger-core-bitcoin-obj OBJECT
-        ${ledger-core-bitcoin-sources}
-        ${SRC_FILES}
-        ${CORE_SRC_FILES}
-        ${HEADERS_FILES}
-        ${CORE_HEADER_FILES})
+    add_library(ledger-core-bitcoin-api-obj OBJECT ${ledger-core-bitcoin-api-sources})
+
+    add_library(ledger-core-bitcoin-obj OBJECT ${SRC_FILES} ${CORE_SRC_FILES} ${CORE_HEADER_FILES})
```
gagbo: We lost `${HEADERS_FILES}` here, I suppose it's just because they are correctly in the `include_directories` of the target, but why do we need to add `${CORE_HEADER_FILES}` then ?

ledger-core-ethereum/.clang-tidy
```diff
@@ -0,0 +1 @@
+../.clang-tidy
```
gagbo: Do we have to add the symlink here ? A root .clang-tidy should be enough isn't it ?

ledger-core-ripple/.clang-tidy
```diff
@@ -0,0 +1 @@
+../.clang-tidy
```
gagbo: Do we have to add the symlink here ? A root .clang-tidy should be enough isn't it ?

ledger-core-tezos/.clang-tidy
```diff
@@ -0,0 +1 @@
+../.clang-tidy
```
gagbo: Do we have to add the symlink here ? A root .clang-tidy should be enough isn't it ?

ledger-core/.clang-tidy
```diff
@@ -0,0 +1 @@
+../.clang-tidy
```
gagbo: Do we have to add the symlink here ? A root .clang-tidy should be enough isn't it ?

ledger-core-ethereum/src/CMakeLists.txt
```diff
@@ -87,20 +87,21 @@ if(IS_IOS GREATER_EQUAL 0)
...
 else()
-    add_library(ledger-core-ethereum-obj OBJECT
-        ${ledger-core-ethereum-sources}
-        ${SRC_FILES}
-        ${CORE_SRC_FILES}
-        ${HEADERS_FILES}
-        ${CORE_HEADER_FILES})
+    add_library(ledger-core-ethereum-api-obj OBJECT ${ledger-core-ethereum-api-sources})
+
+    add_library(ledger-core-ethereum-obj OBJECT ${SRC_FILES} ${CORE_SRC_FILES} ${CORE_HEADER_FILES})
```
gagbo: We lost `${HEADERS_FILES}` here, I suppose it's just because they are correctly in the `include_directories` of the target, but why do we need to add `${CORE_HEADER_FILES}` then ?

ledger-core-ripple/src/CMakeLists.txt
```diff
@@ -87,20 +87,21 @@ if(IS_IOS GREATER_EQUAL 0)
...
 else()
-    add_library(ledger-core-ripple-obj OBJECT
-        ${ledger-core-ripple-sources}
-        ${SRC_FILES}
-        ${CORE_SRC_FILES}
-        ${HEADERS_FILES}
-        ${CORE_HEADER_FILES})
+    add_library(ledger-core-ripple-api-obj OBJECT ${ledger-core-ripple-sources})
+
+    add_library(ledger-core-ripple-obj OBJECT ${SRC_FILES} ${CORE_SRC_FILES} ${CORE_HEADER_FILES})
```
gagbo: We lost `${HEADERS_FILES}` here, I suppose it's just because they are correctly in the `include_directories` of the target, but why do we need to add `${CORE_HEADER_FILES}` then ?

ledger-core-tezos/src/CMakeLists.txt
```diff
@@ -87,20 +87,22 @@ if(IS_IOS GREATER_EQUAL 0)
...
-    add_library(ledger-core-tezos-obj OBJECT
-        ${ledger-core-tezos-sources}
-        ${SRC_FILES}
-        ${CORE_SRC_FILES}
-        ${HEADERS_FILES}
-        ${CORE_HEADER_FILES})
+
+    add_library(ledger-core-tezos-api-obj OBJECT ${ledger-core-tezos-api-sources})
+
+    add_library(ledger-core-tezos-obj OBJECT ${SRC_FILES} ${CORE_SRC_FILES} ${CORE_HEADER_FILES})
```
gagbo: We lost `${HEADERS_FILES}` here, I suppose it's just because they are correctly in the `include_directories` of the target, but why do we need to add `${CORE_HEADER_FILES}` then ?

ledger-core-bitcoin/.clang-tidy
alekece: The symlink is mandatory because of the current source directory is `ledger-core-bitcoin` so - in a `clang-tidy` point of view - there is no configuration file. 
BTW, you question opens my mind a little bit more and `clang-tidy` has an option to add a configuration path. I will use that instead of the symlink

ledger-core-bitcoin/src/CMakeLists.txt
alekece: I removed `${HEADERS_FILES}` because the variable is unset.

ledger-core-ethereum/.clang-tidy
alekece: ditto

cmake/StaticAnalyzers.cmake
```diff
@@ -0,0 +1,38 @@
+function(enable_static_analyzers target_name)
```
phaazon: The name of that function doesn’t exactly do what it’s implemented with — i.e. it can do nothing.

It can be good to isolate the scope of the target setting, and move the enable/disable logic to somewhere else.

ledger-core-bitcoin/src/CMakeLists.txt
```diff
@@ -34,66 +34,7 @@ if (TARGET_JNI)
...
-    bitcoin/transactions/BitcoinLikeUTXOPicker.cpp
-    # others
-    bitcoin/BitcoinLikeAccount.cpp
-    bitcoin/BitcoinLikeAddress.cpp
-    bitcoin/BitcoinLikeCurrencies.cpp
-    bitcoin/BitcoinLikeExtendedPublicKey.cpp
-    bitcoin/BitcoinLikeWallet.cpp
-    bitcoin/BitcoinNetworks.cpp
-)
+file(GLOB_RECURSE SRC_FILES *.cpp)
```
phaazon: This change is not part of the PR. Why do you need to do that? Also, why do we have an explicit listing here — ping @pollastri-pierre who might have that information.

ledger-core-bitcoin/src/CMakeLists.txt
```diff
@@ -146,20 +87,21 @@ if(IS_IOS GREATER_EQUAL 0)
...
-        ${CORE_SRC_FILES}
-        ${HEADERS_FILES}
-        ${CORE_HEADER_FILES})
+    add_library(ledger-core-bitcoin-api-obj OBJECT ${ledger-core-bitcoin-api-sources})
+
+    add_library(ledger-core-bitcoin-obj OBJECT ${SRC_FILES} ${CORE_SRC_FILES} ${CORE_HEADER_FILES})
     target_link_libraries(ledger-core-bitcoin-obj PUBLIC ledger-core-bitcoin-interface)
 
+    # enable static analyzers if requested
+    include(StaticAnalyzers)
```
phaazon: Can you conditionally include here? The problem is that you will always include that file even if we don’t want static analyzers.

ledger-core-ethereum/src/CMakeLists.txt
```diff
@@ -87,20 +87,21 @@ if(IS_IOS GREATER_EQUAL 0)
...
     target_link_libraries(ledger-core-ethereum-obj PUBLIC ledger-core-ethereum-interface)
 
+    # enable static analyzers if requested
+    include(StaticAnalyzers)
```
phaazon: Ditto.

ledger-core-ripple/src/CMakeLists.txt
```diff
@@ -87,20 +87,21 @@ if(IS_IOS GREATER_EQUAL 0)
...
     target_link_libraries(ledger-core-ripple-obj PUBLIC ledger-core-ripple-interface)
 
+    # enable static analyzers if requested
+    include(StaticAnalyzers)
```
phaazon: Ditto.

ledger-core-tezos/src/CMakeLists.txt
```diff
@@ -87,20 +87,22 @@ if(IS_IOS GREATER_EQUAL 0)
...
     target_link_libraries(ledger-core-tezos-obj PUBLIC ledger-core-tezos-interface)
 
+    # Enable static analyzers if requested
+    include(StaticAnalyzers)
```
phaazon: Ditto.

ledger-core/src/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 # Automatically add API files to the library
 file(
     GLOB ledger-core-api-sources
-    "${CMAKE_SOURCE_DIR}/inc/core/api/*.cpp"
+    "${CMAKE_SOURCE_DIR}/inc/core/api/*"
```
phaazon: Hm. Why do you need every file now? I’m not sure what `.hpp` are for here.

gagbo: api headers were included in `HEADERS_FILES` before
The change is 2 part, accumulated with 
```patch
-file(GLOB_RECURSE HEADERS_FILES ../inc/*.{h,hpp})
+file(GLOB_RECURSE HEADERS_FILES ../inc/*.hpp)
+list(REMOVE_ITEM HEADERS_FILES ../inc/api/*.hpp)
```
below.

Combined it has the effect of having `ledger-core-api-sources` be both `cpp` and `hpp` files, whereas `HEADERS_FILES` holds only non generated headers. The api headers were always included in the `ledger-core-obj` target.

Now that there are both a `ledger-core-obj` and a `ledger-core-api-obj` target, it makes sense to split the headers depending on being `api/` or `not`

ledger-core-bitcoin/src/CMakeLists.txt
alekece: I am the one who introduces the explicit listing at the first place - It was to ease the bitcoin segregation.

cmake/StaticAnalyzers.cmake
alekece: I agree the function name isn't that much relevant; do you have a suggestion ?

Since `StaticAnalyzers.cmake` is a helper, it have to hold the options and the logic itself.

ledger-core-bitcoin/src/CMakeLists.txt
alekece: I want to always include that file - I explained why in previous answer.

phaazon: Always including a file adds more parsing time and thus, more time to compile. Applying that rule everywhere leads to time-costly CMake configuration.

When we don’t want _any static analyzers_, we are going to pay a price for nothing here.

alekece: Are we really talking about time consuming in `cmake` part ? Launch `cmake` command takes few seconds.

What struggle me a lot with your suggestion is :
* We lose completely the benefit of `cmake` helper; isolate `cmake` logic and reuse it as needed.
* The check we have to do to enable static analyzers is to verify if at least one of them is enabled so if we want to add a new one, we have to change the condition in all `CMakeFiles.txt`.
* We make something at first simple harder to use.

PR #535, opened 2020-03-25
Fix undefined reference to dl on ubuntu

alekece:
This PR is quite simple : link `dl` library to `soci-core`.

## Mandatory picture of a cat
black_cat


ledger-core/lib/soci/core/CMakeLists.txt
```diff
@@ -111,6 +111,8 @@ if (SOCI_STATIC)
 
   add_library(${SOCI_CORE_TARGET_STATIC} STATIC ${SOCI_CORE_HEADERS} ${SOCI_CORE_SOURCES})
 
+  target_link_libraries(${SOCI_CORE_TARGET_STATIC} ${CMAKE_DL_LIBS})
```
phaazon: Please test the platform you’re on is a unix / linux. I’m not sure about the behavior of that stuff on Windows.

PR #538, opened 2020-03-26
Add compiler warnings

alekece:
The PR aims to enable a bunch of compiler warnings for `MSVC`, `Clang`, `AppleClang` and `Gcc`.

The compilation of `lib-ledger-core` - more precisely `ledger-core` and all the coins - triggers a lot of warnings so the compiler warnings feature is optional  and disable by default.

Compiler warnings can be enabled such as :
```
cmake -DENABLE_WARNINGS=ON -DWARNINGS_AS_ERRORS=ON ...
```

## Mandatory picture of an animal
hole_cat

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -1,0 +1,83 @@
+function(enable_warnings target_name)
+    option(ENABLE_WARNINGS "Enable compiler warnings" OFF)
+
+    if (ENABLE_WARNINGS)
+        set(MSVC_WARNINGS
+            /W4 # Baseline reasonable warnings
+            /w14242 # 'identfier': conversion from 'type1' to 'type1', possible loss of data
+            /w14254 # 'operator': conversion from 'type1:field_bits' to 'type2:field_bits', possible loss
+                    # of data
+            /w14263 # 'function': member function does not override any base class virtual member function
+            /w14265 # 'classname': class has virtual functions, but destructor is not virtual instances of
+                    # this class may not be destructed correctly
+            /w14287 # 'operator': unsigned/negative constant mismatch
+            /we4289 # nonstandard extension used: 'variable': loop control variable declared in the for-
+                    # loop is used outside the for-loop scope
+            /w14296 # 'operator': expression is always 'boolean_value'
+            /w14311 # 'variable': pointer truncation from 'type1' to 'type2'
+            /w14545 # expression before comma evaluates to a function which is missing an argument list
+            /w14546 # function call before comma missing argument list
+            /w14547 # 'operator': operator before comma has no effect; expected operator with side-effect
+            /w14549 # 'operator': operator before comma has no effect; did you intend 'operator'?
+            /w14555 # expression has no effect; expected expression with side- effect
+            /w14619 # pragma warning: there is no warning number 'number'
+            /w14640 # Enable warning on thread un-safe static member initialization
+            /w14826 # Conversion from 'type1' to 'type_2' is sign-extended. This may cause unexpected
+                    # runtime behavior.
+            /w14905 # wide string literal cast to 'LPSTR'
+            /w14906 # string literal cast to 'LPWSTR'
+            /w14928 # illegal copy-initialization; more than one user-defined conversion has been
+                    # implicitly applied
+        )
+
+        set(CLANG_WARNINGS
+            -Wall
+            -Wextra # reasonable and standard
+            -Wshadow # warn the user if a variable declaration shadows one from a parent context
+            -Wnon-virtual-dtor # warn the user if a class with virtual functions has a non-virtual
+                                # destructor. This helps catch hard to track down memory errors
+            -Wold-style-cast # warn for c-style casts
+            -Wcast-align # warn for potential performance problem casts
+            -Wunused # warn on anything being unused
+            -Woverloaded-virtual # warn if you overload (not override) a virtual function
+            -Wpedantic # warn if non-standard C++ is used
+            -Wconversion # warn on type conversions that may lose data
+            -Wsign-conversion # warn on sign conversions
+            -Wnull-dereference # warn if a null dereference is detected
+            -Wdouble-promotion # warn if float is implicit promoted to double
+            -Wformat=2 # warn on security issues around functions that format output (ie printf)
+        )
+        
+        if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+            set(CLANG_WARNINGS ${CLANG_WARNINGS} -O0 -g)
+            set(MSVC_WARNINGS ${MSVC_WARNINGS} /Od)
+        elseif(CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+            set(CLANG_WARNINGS ${CLANG_WARNINGS} -O3 -g)
+            set(MSVC_WARNINGS ${MSVC_WARNINGS} /O2d)
+        elseif(CMAKE_BUILD_TYPE STREQUAL "MinSizeRel")
+            set(CLANG_WARNINGS ${CLANG_WARNINGS} -Os)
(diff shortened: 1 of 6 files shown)


cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
+function(set_enabled_compiler_warnings target_name)
+    option(ENABLE_WARNINGS "Enable compiler warnings" OFF)
+    option(WARNINGS_AS_ERRORS "Treat warnings as erros" OFF)
+
+    if (ENABLE_WARNINGS)
+        set(MSVC_WARNINGS
+            /W4 # Baseline reasonable warnings
+            /w14242 # 'identfier': conversion from 'type1' to 'type1', possible loss of data
```
phaazon: identifier*

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
...
+            /w14254 # 'operator': conversion from 'type1:field_bits' to 'type2:field_bits', possible loss
+                    # of data
+            /w14263 # 'function': member function does not override any base class virtual member function
+            /w14265 # 'classname': class has virtual functions, but destructor is not virtual instances of
```
phaazon: instance*

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
...
+                    # this class may not be destructed correctly
```
phaazon: Weird phrasing, I think some punctuation is missing.

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
...
+            /w14905 # wide string literal cast to 'LPSTR'
+            /w14906 # string literal cast to 'LPWSTR'
+            /w14928 # illegal copy-initialization; more than one user-defined conversion has been
+                    # implicitly applied
+        )
+
+        set(CLANG_WARNINGS
+            -Wall
+            -Wextra # reasonable and standard
+            -Wshadow # warn the user if a variable declaration shadows one from a parent context
```
phaazon: Isn’t that in `-Wall` already?

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
...
+            -Wpedantic # warn if non-standard C++ is used
+            -Wconversion # warn on type conversions that may lose data
+            -Wsign-conversion # warn on sign conversions
+            -Wnull-dereference # warn if a null dereference is detected
+            -Wdouble-promotion # warn if float is implicit promoted to double
+            -Wformat=2 # warn on security issues around functions that format output (ie printf)
+        )
+        
+        if(WARNINGS_AS_ERRORS)
+            message("warnings as errors")
```
phaazon: Please provide a better message for the user interface (with a verb so that the user know what’s happening).

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
+function(set_enabled_compiler_warnings target_name)
+    option(ENABLE_WARNINGS "Enable compiler warnings" OFF)
+    option(WARNINGS_AS_ERRORS "Treat warnings as erros" OFF)
+
+    if (ENABLE_WARNINGS)
+        set(MSVC_WARNINGS
```
phaazon: Even though this is not going to be used on non-Windows, it’s a bit silly to declare variables when not needed. Not very important, but still a bit iffy.

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,90 @@
...
+        )
+
+        target_compile_options(
+            ${target_name} PRIVATE
+              $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
+              ${CLANG_WARNINGS}>
+              $<$<CXX_COMPILER_ID:MSVC>:
+              ${MSVC_WARNINGS}>
+              $<$<CXX_COMPILER_ID:GNU>:
+              ${GCC_WARNINGS}>)
```
phaazon: Can you format in a way it’s easier to read? Like:
```suggestion
            ${target_name} PRIVATE
                 $<$<OR:
                     $<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:
                         ${CLANG_WARNINGS}
                     >
                     $<$<CXX_COMPILER_ID:MSVC>:
                         ${MSVC_WARNINGS}
                     >
                     $<$<CXX_COMPILER_ID:GNU>:
                         ${GCC_WARNINGS}
                     >
        )
```

ledger-core-bitcoin/src/CMakeLists.txt
```diff
@@ -96,6 +96,10 @@ else()
     include(StaticAnalyzers)
     set_enabled_static_analyzers(ledger-core-bitcoin-obj)
 
+    # add compiler warnings
+    include(CompilerWarnings)
+    set_enabled_compiler_warnings(ledger-core-bitcoin-obj)
```
phaazon: Argh, I didn’t see those names, `set_enabled`, what does it mean? It’s a bit confusing… :confused:

ledger-core/src/CMakeLists.txt
```diff
@@ -94,6 +92,9 @@ if(IS_IOS GREATER_EQUAL 0)
     )
 else()
     add_library(ledger-core-api-obj OBJECT ${ledger-core-api-sources})
+    target_compile_options(ledger-core-api-obj PRIVATE
+        -Wno-return-type
```
phaazon: Why is this one like an exception?

cmake/CompilerWarnings.cmake
alekece: Not at all but after checking it one by one there are some duplicates.

cmake/CompilerWarnings.cmake
alekece: Concerning code formatting, I will introduce `cmake-format` tool.

ledger-core/src/CMakeLists.txt
alekece: Oops ! This is something I forgot.

ledger-core/src/CMakeLists.txt
```diff
@@ -163,7 +166,7 @@ target_link_libraries(ledger-core-interface INTERFACE ethash)
 target_link_libraries(ledger-core-interface INTERFACE ${SQLITE_LIB})
 
 target_include_directories(
-    ledger-core-interface INTERFACE
+    ledger-core-interface SYSTEM INTERFACE
```
phaazon: Why do you need `SYSTEM` here?

ledger-core/src/CMakeLists.txt
```diff
@@ -102,6 +100,11 @@ else()
     include(StaticAnalyzers)
     set_enabled_static_analyzers(ledger-core-obj)
 
+    # set compiler warnings
+    include(CompilerWarnings)
+    set_compiler_warnings(ledger-core-obj)
+    disable_compiler_warnings(ledger-core-api-obj)
```
phaazon: Yeah see, this why `set` / `enable` are not good names. It feels like you always disable warnings for API (which I think is okay) and always set warnings for our code. I’m not feeling comfortable with that naming, please find other names (`toggle` is still a good candidate to me).

cmake/CompilerWarnings.cmake
```diff
@@ -0,0 +1,93 @@
+function(set_compiler_warnings target_name)
```
phaazon: See my remark below.

ledger-core/src/CMakeLists.txt
alekece: The `SYSTEM` keyword is needed to avoid to propagate compiler warnings to our dependencies.
`cmake` will interpret `SYSTEM` by adding `-Isystem` to the compiler command line which will skip all headers after that option.

PR #543, opened 2020-04-02
LLC-572 — Fix missing internal transactions for Ethereum.

phaazon:
# Content

Those were missing because of two issues:

- We didn’t create internal transactions UUID correctly. We initially
  assumed that we have only one internal transaction (uuid =
  transaction_uuid + hash_data, for short). This fix adds the index of
  the internal transaction to generate different UUIDs and then get them
  all.
- We didn’t filter correctly transactions that don’t concern us. We then
  end up (at least on the core side) with several `NONE` transactions.
  The filtering here basically requires to discard internal transactions
  which type is set to `NONE`, since they don’t concern us — the
  explorers seem to still send us those transactions as they don’t
  filter them from the parent transaction.

# Mandatory picture of celestial animal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ compile_commands.json
 install_manifest.txt
 tags
 *.a
+/.clangd/
--- a/core/src/wallet/ethereum/EthereumLikeAccount.cpp
+++ b/core/src/wallet/ethereum/EthereumLikeAccount.cpp
@@ -107,9 +107,9 @@ namespace ledger {
                 throw Exception(api::ErrorCode::RUNTIME_ERROR, "Wallet reference is dead.");
             }
 
-            if (transaction.block.nonEmpty())
+            if (transaction.block.nonEmpty()) {
                 putBlock(sql, transaction.block.getValue());
-
+            }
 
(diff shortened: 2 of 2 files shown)


phaazon:
Adding @RomanWlm because I’m changing the way UUIDs of internal transactions are computed. Is that a problem for the Wallet Daemon?

phaazon:
After some internal discussions, we have potentially something else to do: make a data migration to recreate the previously persisted internal transactions’ UIDs with the new scheme. Even though it should be possible to do, I think it’s not a good move, because if some previousl synchronization missed internal transactions (because of the UID scheme bug), then they’ll still be missing them (the data transaction cannot introduce new internal transactions). So those users will still have to resync. And we don’t know who needs to do that, so I think we are just going to hide issues here.

phaazon:
I think this change — data migration — is not needed because if we already have associated internal operations, they already have a UID and it’s a random one. There is no case where we need to create those UIDs besides the code that got changed and collisions won’t happen (because of the added index).

I think we’re pretty safe without the data migration, @KhalilBellakrid.

PR #545, opened 2020-04-07
[iOS] Add x86_64 to fat framework

MortalKastor:
XCode had a recent change of heart regarding embedded library architectures allowed in a build so, long story short, we have to change the way we embed `ledger-core.framework` in Ledger Live mobile, meaning we don't use _thin_ frameworks anymore but only the _fat_ one, in which we need _all_ the architectures, including `x86_64`.

### DIFF
--- a/.circleci/deploy_fat_lib.sh
+++ b/.circleci/deploy_fat_lib.sh
@@ -6,8 +6,7 @@ if [ -n "$CIRCLE_TAG" ] || [ "$CIRCLE_BRANCH" == "master" -o "$CIRCLE_BRANCH" ==
 	echo "======> Build Fat Library"
 	pwd
 
-	#We don't put x86_64/ledger-core.framework/ledger-core otherwise we have a problem when pushing to AppStore
-	lipo -create armv7/ledger-core.framework/ledger-core arm64/ledger-core.framework/ledger-core -o ledger-core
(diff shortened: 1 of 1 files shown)


.circleci/deploy_fat_lib.sh
```diff
@@ -6,8 +6,7 @@ if [ -n "$CIRCLE_TAG" ] || [ "$CIRCLE_BRANCH" == "master" -o "$CIRCLE_BRANCH" ==
 	echo "======> Build Fat Library"
 	pwd
 
-	#We don't put x86_64/ledger-core.framework/ledger-core otherwise we have a problem when pushing to AppStore
-	lipo -create armv7/ledger-core.framework/ledger-core arm64/ledger-core.framework/ledger-core -o ledger-core
+	lipo -create armv7/ledger-core.framework/ledger-core arm64/ledger-core.framework/ledger-core x86_64/ledger-core.framework/ledger-core -o ledger-core
```
phaazon, line 9: The comment is pretty scary. Have you tried it? :D

MortalKastor: That comment made sense because we were embedding the full, untouched, fat framework with the app (and, rightly so, the App Store reject any PC arch to be shipped on phones), but we now changed the way we build and selectively strip any unneeded arch from our libs before embedding them.

gre:
@KhalilBellakrid is it mergeable? (check also the referenced PRs related to it) Thanks!

PR #546, opened 2020-04-08
Fix XRP: marker to be set to Uint64 instead of string

KhalilBellakrid:
@RomanWlm could you please complete with right ticket (could not find it).

**NOTE**: it seems that `The format of the marker field is intentionally undefined.` (from: https://xrpl.org/markers-and-pagination.html), so this is valid at least for `s3` and `s2`.


RomanWlm:
I Cannot edit your description

> @RomanWlm could you please complete with right ticket (could not find it).
> 
> **NOTE**: it seems that `The format of the marker field is intentionally undefined.` (from: https://xrpl.org/markers-and-pagination.html), so this is valid at least for `s3` and `s2`.

here is the ticker => https://ledgerhq.atlassian.net/browse/LLC-630

PR #548, opened 2020-04-08
Use soci::transaction to solve potential race conditions between observers and synchronizers

KhalilBellakrid:
This is related to: https://ledgerhq.atlassian.net/browse/LVS-140


KhalilBellakrid:
Will be included in : https://github.com/LedgerHQ/lib-ledger-core/pull/551

PR #550, opened 2020-04-10
LLC-433 — Remove external Tezos explorer fallback recover for HTTP errors

phaazon:
The recover feature was introduced for non-activated accounts, so that
they can return “default” values — such as zero balance, etc. The
problem is that this recover mechanism also gets in the way when we are
having HTTP issues (down explorers, for instance).

This commit catches HTTP errors and rethrows the exception so that no
fallback value hides the issue.


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -255,6 +255,15 @@ namespace ledger {
                                         return std::make_shared<BigInt>(value);
                                     })
                     .recover(getContext(), [fallbackValue] (const Exception &exception) {
+                        auto ecode = exception.getErrorCode();
+                        if (ecode == api::ErrorCode::HTTP_ERROR || ecode == api::ErrorCode::UNABLE_TO_CONNECT_TO_HOST) {
+                          // if it’s an HTTP error, it might be due to the host not being reachable or such,
+                          // so we re-run the error
+                          throw exception;
```
gre, line 262: this likely broke https://ledgerhq.atlassian.net/browse/LLC-674

PR #552, opened 2020-04-14
[develop] [Coin Integration] Add support for cosmoshub chain

gagbo:
# Scope

## Currency
Add support for CosmosHub chain and its native currency `ATOM`

## Backend used
Uses the explorer/node REST API given by the [gaiacli/gaiad combination](https://github.com/cosmos/gaia).

Currently the code is compatible with `v0.37.8`

## Features

- [x] Synchronize account data (sequence, account number, balances)
- [x] Get list of validators, delegations, pending undelegation, pending rewards, pending redelegations
- [x] Serialize messages for on-device signature and broadcast to network
- [x] Support for creating common "delegator-level" operations (send, delegate, undelegate, redelegate, claim)
- [x] Store all operation history state in database for incremental updates

### External resources

https://cosmos.network : Official site

https://github.com/gavinly/CosmosParametersWiki : Chain parameters wiki

https://github.com/cosmos/cosmos-sdk/ : cosmos-sdk source, which defines the REST API used in the explorer implementation.


gagbo:
I think I'll try to squash history here and there

gagbo:
I'll squash everything in one commit and try to split from there then. I don't think I'll be able to maintain CI checks with the splitting of the code though.

I'll try to split it this way :
- `idl` changes and generated C++ interface
- implementers of the interface with the helper structures and constants
- `database` migration and database helpers
- JNI interface

gagbo:
Another solution, is to rebase into multiple big commits, and this way in the review UI it should be possible to only see changes from "interfaces" commit, or the "database" commit etc.

This makes review a little weirder, but maintains CI

gagbo:
Closed by #556

PR #555, opened 2020-04-15
Undefined symbols related to dl library

alekece:
The PR aims to fix the undefined symbols related to `dl` library on Unix systems.

## Mandatory picture of a cat
slice_cat

### DIFF
--- a/core/lib/CMakeLists.txt
+++ b/core/lib/CMakeLists.txt
@@ -38,6 +38,9 @@ set(SOCI_STATIC ON)
 set(SOCI_SHARED OFF CACHE BOOL "USE ONLY STATIC" FORCE)
 
 add_subdirectory(soci)
+# Fix undefined symbol related to dl library on Unix system
+# The dl library is linked here to avoid to modify vendor cmake files
+target_link_libraries(soci_core_static ${CMAKE_DL_LIBS})
 add_subdirectory(soci_sqlite3)
 
 if (PG_SUPPORT)


core/lib/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ set(SOCI_STATIC ON)
 set(SOCI_SHARED OFF CACHE BOOL "USE ONLY STATIC" FORCE)
 
 add_subdirectory(soci)
+# Fix undefined symbol related to dl library on Unix system
+# The dl library is linked here to avoid to modify vendor cmake files
+target_link_libraries(soci_core_static ${CMAKE_DL_LIBS})
```
phaazon, line 43: Shouldn’t that be gated by the platform, at least `if (UNIX)`?

alekece: You're right !

gagbo: Can confirm (had a hackier version of the PR for a while) :

`  target_link_libraries(${SOCI_CORE_TARGET_STATIC} dl)` fails on Windows
`  target_link_libraries(${SOCI_CORE_TARGET_STATIC} ${CMAKE_DL_LIBS})` works on Windows

gagbo: (Well, CI too can confirm, I suppose)

phaazon: The [documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_DL_LIBS.html) doesn’t say anything about what it should do if it’s not on UNIX, which is why I got a bit scared there. I guess it should be fine…

alekece:
The PR is still blocked because I need at least two approvals - @gagbo your approval seems to be partial. 
@KhalilBellakrid @pollastri-pierre Can you take some time to review it please ?

phaazon:
@alekece to me we can merge it, @gagbo approval counts as a complete approval despite what GitHub says. Let’s wait for the CI and we’ll merge it when ready.

PR #557, opened 2020-04-15
Minor fixes on XLM implementation

pollastri-pierre:
- Fetch sequence from the explorer instead of relying on database. This ensures that the sequence is up to date with the network.
- Fix a minor issue on SQL create account query.

### DIFF
--- a/core/src/wallet/stellar/StellarLikeAccount.cpp
+++ b/core/src/wallet/stellar/StellarLikeAccount.cpp
@@ -499,13 +499,10 @@ namespace ledger {
 
         Future<BigInt> StellarLikeAccount::getSequence() {
             auto self = getSelf();
-            return async<BigInt>([=] () {
-                stellar::Account account;
-                soci::session sql(self->getWallet()->getDatabase()->getPool());
-                if (StellarLikeAccountDatabaseHelper::getAccount(sql, self->getAccountUid(), account))
-                    return BigInt::fromString(account.sequence);
-                return BigInt::ZERO;
-            });
+            return self->_params.explorer->getAccount(self->_params.keychain->getAddress()->toString())
+                .map<BigInt>(getContext(), [=] (const std::shared_ptr<stellar::Account>& account) {
+                    return BigInt::fromString(account->sequence);
+                });
         }
 
         void StellarLikeAccount::getSigners(const std::shared_ptr<api::StellarLikeAccountSignerListCallback> &callback) {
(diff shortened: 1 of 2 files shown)


core/src/wallet/stellar/StellarLikeAccount.cpp
```diff
@@ -499,13 +499,10 @@ namespace ledger {
...
         Future<BigInt> StellarLikeAccount::getSequence() {
             auto self = getSelf();
-            return async<BigInt>([=] () {
-                stellar::Account account;
-                soci::session sql(self->getWallet()->getDatabase()->getPool());
-                if (StellarLikeAccountDatabaseHelper::getAccount(sql, self->getAccountUid(), account))
-                    return BigInt::fromString(account.sequence);
-                return BigInt::ZERO;
-            });
+            return self->_params.explorer->getAccount(self->_params.keychain->getAddress()->toString())
```
phaazon, line 502: Since you removed the closure, you can get rid of the `self` parameter here.

pollastri-pierre: Indeed

core/src/wallet/stellar/database/StellarLikeAccountDatabaseHelper.cpp
```diff
@@ -122,8 +122,8 @@ namespace ledger {
                                                              int32_t accountIndex, const stellar::Account &in) {
             auto accountUid = AccountDatabaseHelper::createAccountUid(walletUid, accountIndex);
             sql << "INSERT INTO stellar_accounts VALUES (:uid, :wallet_uid, :idx, :address, :sequence, :count)",
-                    use(accountUid), use(walletUid), use(accountIndex), use(in.accountId), use(in.sequence),
-                    use(in.subentryCount);
+                    use(accountUid), use(walletUid), use(accountIndex), use(in.accountId),
+                    use(in.subentryCount), use(in.sequence);
```
alekece, line 126: The `:count` is the last value but you provide the sequence instead... The previous code looks like the correct way to create the account UID

phaazon: The names are not really important if you don’t name them in the `soci::use`. The fix here is to swap the names in the SQL query because otherwise it’s confusing. But that code will do the right thing.

pollastri-pierre: The label are not correct, I correct them. The new order is the good one. regarding the account table schema.

PR #559, opened 2020-04-16
LLC-484 – Turn 64-bit integers to 32-bit integers in the interface.

phaazon:
# Content

Most of those types are used for stuff like query limits, block heights
etc. which are all far from ever reaching the maximum value for a 32-bit
integers, and, also, we use 32-bit a lot in the rest of the codebase.

This commit is for LLC-484 where, on ReactNative, 64-bit precision
integers are not a possible thing — currently, the Ledger Live team has
an internal patch command they run to automatically turn the `long`
argument into `int` argument.

Once this commit is merged, we need to merge the ReactNative PR that
will update the interface so that they can use it.

# Mandatory picture of a cute doggo


phaazon:
I’m holding that PR because the changes are quite important, regarding the interface. It will impact all our clients, and I’m not completely sure it’s a good idea. We need to be _sure_ switching to 32-bit for those is not going to be a regression.

Since @gre told us the problem comes in only for a single function (query filter limit), maybe a first fix would be to just pass that one to 32-bit. The maximum value there would be `2147483647` (because of signed arithmetics), which I think should be fiiiiiiiiine for a query filter.

What do you guys think? @KhalilBellakrid @alekece @gre @teams2ua

gre:
i'm fine with the present choices but if in future we still want to keep some "long" api we will have the discussion again right?
alternatively, it could also be changed on the React Native binding to remap long->int with risk that we accept to lose precision on the React Native side. OR the binding could use a STRING instead of an integer for long values.
all are possible choices with different implications.

phaazon:
Yes, we can switch to a `BigInt` representation, which will be an “okay” impact. I thought about turning 64-bit -> 32-bit in the RN binding but it would hide the cast away, which might introduce bugs.

So, given the feedback from everybody, we should be able to merge this and if at some day we need more bits, `BigInt` should be a good candidate.

PR #560, opened 2020-04-17
Fix/coin selection

KhalilBellakrid:
**Fixed filterWithKnapsackSolver:**
 - estimate right size for signed UTXOs,
 - use dust amount + cost of spending a change as minimum change amount,
 - find adequate subset by targetting amount + fixed fees, instead of amount + fixed fees + fees of signed UTXOs of initial subset,
 - fix above is also fixing the change amount.

**Test:** add CoinSelectionP2PKH.CompareUTXOPickingStrategies: build N (here 20) txs with 3 different strategies (deep, merge and optimize) then compare generated fees and number outputs.

**Note:** this related to [LV-2340](https://ledgerhq.atlassian.net/browse/LV-2340)

**HODL:** removed, confirmed by QA


core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -190,7 +190,8 @@ namespace ledger {
 
             //Don't use this strategy for wipe mode (we have more performent strategies for this use case)
             if (buddy->request.wipe) {
-                throw make_exception(api::ErrorCode::ILLEGAL_ARGUMENT, "Strategy filterWithLowestFees can't be called when wiping funds to address, please use another strategy");
+                buddy->logger->debug("Strategy filterWithOptimizeSize with wipe to addreess mode, using filterWithDeepFirst");
```
alekece, line 193:
```suggestion
                buddy->logger->debug("Strategy filterWithOptimizeSize with wipe to address mode, using filterWithDeepFirst");
```

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -503,17 +505,37 @@ namespace ledger {
...
             buddy->logger->debug("Approximate Best Subset 1st try");
-            approximateBestSubset(vUTXOs, totalLower, BigInt((int64_t)amountWithFees), bestValues, bestValue);
-            if (bestValue != amountWithFees && totalLower >= amountWithFees + MIN_CHANGE) {
+            // Here we target the value amountWithFixedFees which is the amount of tx + fixed fees (fees of transaction without signed UTXOs)
+            approximateBestSubset(vUTXOs, totalLower, BigInt((int64_t)amountWithFixedFees), bestValues, bestValue);
+            if (bestValue != amountWithFixedFees && totalLower >= amountWithFixedFees + minimumChange) {
                 buddy->logger->debug("First approximation, bestValue {} with {} bestValues", bestValue, bestValues.size());
                 buddy->logger->debug("Approximate Best Subset 2nd try");
-                approximateBestSubset(vUTXOs, totalLower, BigInt((int64_t)amountWithFees + MIN_CHANGE), bestValues, bestValue);
+                approximateBestSubset(vUTXOs, totalLower, BigInt((int64_t)amountWithFixedFees + minimumChange), bestValues, bestValue);
```
alekece, line 513:
```suggestion
                approximateBestSubset(vUTXOs, totalLower, BigInt(static_cast<int64_t>(amountWithFixedFees) + minimumChange), bestValues, bestValue);
```

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -503,17 +505,37 @@ namespace ledger {
...
                 buddy->logger->debug("Second approximation, bestValue {} with {} bestValues", bestValue, bestValues.size());
             }
 
+            // We compute the the totalBest here because we need picked UTXOs in subset
```
alekece, line 517:
```suggestion
            // We compute the totalBest here because we need picked UTXOs in subset
```

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -503,17 +505,37 @@ namespace ledger {
...
+            // to be able to update amountWithFees (because after signing all UTXOs we might need to pick additional UTXOs to cover
+            // fees generated by adding this UTXOs and their signature)
+            int64_t totalBest = 0;
+            UTXODescriptorList tmpOut;
+            for (unsigned int i = 0; i < vUTXOs.size(); i++) {
+                if (bestValues[i])
+                {
+                    UTXODescriptor utxoDescriptor{vUTXOs[i]->getTransactionHash(), vUTXOs[i]->getOutputIndex(), std::get<1>(buddy->request.utxoPicker.getValue())};
```
alekece, line 525: Can you create a `constexpr` variable which represents the magic value `1` in the `std::get` ? Like that the code will be clearer

phaazon: I’m not quite sure what you mean here @alekece

andreishtamburg:
Hey guys!
Testing this fix.
One thing that is quite suspicious is that I'm getting the same fees for MERGE_OUTPUT and OPTIMIZE_SIZE. I understand that it basically means that the same UTXO are selected. 

I have the following distribution:
1 UTXO of 0.00664456
99 UTXO of 0.01 BTC
2 UTXO of 1 BTC
1 UTXO of 5 BTC
===============
2 questions:

1. Is that possible that with OPTIMIZE_SIZE and MERGE_OUTPUTS strategies fees would be the same with the aforementioned distribution?
2. If strategy is OPTIMIZE_OUTPUT. what UTXOs should be selected when trying to send 0.5 BTC?

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
alekece: I mean `std::get<1>` is opaque and prone to error; if the underlying types in the tuple change - for some reason - this code will be break even if it looks great at the first glance. 
With that duck conversation with myself, my previous suggestion doesn't look that great : even if we put the magic value to a variable, the code will look better but the code is still error-proned.
Maybe a plain-old function to get back the value from that tuple will be less error-proned, more readable and reusable.

KhalilBellakrid:
> Hey guys!
> Testing this fix.
> One thing that is quite suspicious is that I'm getting the same fees for MERGE_OUTPUT and OPTIMIZE_SIZE. I understand that it basically means that the same UTXO are selected.
> 
> # I have the following distribution:
> 1 UTXO of 0.00664456
> 99 UTXO of 0.01 BTC
> 2 UTXO of 1 BTC
> 1 UTXO of 5 BTC
> 2 questions:
> 
> 1. Is that possible that with OPTIMIZE_SIZE and MERGE_OUTPUTS strategies fees would be the same with the aforementioned distribution?
> 2. If strategy is OPTIMIZE_OUTPUT. what UTXOs should be selected when trying to send 0.5 BTC?

Depending on your UTXO distribution you could have the same for 3 strategies, the most important thing is that `OPTIMIZE_FEES <= DEEP_FEES <= MERGE_FEES`  same for number of UTXOs. Another thing is that `OPTIMIZE_SIZE` might not converge and just fallback on `DEEP_FIRST` so ... for all these reasons the main thing to test is above rule.

andreishtamburg:
> > Hey guys!
> > Testing this fix.
> > One thing that is quite suspicious is that I'm getting the same fees for MERGE_OUTPUT and OPTIMIZE_SIZE. I understand that it basically means that the same UTXO are selected.
> > # I have the following distribution:
> > 1 UTXO of 0.00664456
> > 99 UTXO of 0.01 BTC
> > 2 UTXO of 1 BTC
> > 1 UTXO of 5 BTC
> > 2 questions:
> > 
> > 1. Is that possible that with OPTIMIZE_SIZE and MERGE_OUTPUTS strategies fees would be the same with the aforementioned distribution?
> > 2. If strategy is OPTIMIZE_OUTPUT. what UTXOs should be selected when trying to send 0.5 BTC?
> 
> Depending on your UTXO distribution you could have the same for 3 strategies, the most important thing is that `OPTIMIZE_FEES <= DEEP_FEES <= MERGE_FEES` same for number of UTXOs. Another thing is that `OPTIMIZE_SIZE` might not converge and just fallback on `DEEP_FIRST` so ... for all these reasons the main thing to test is above rule.

Well. In my case I have the smallest fees for DEEP_OUTPUT_FIRST.
And the same fees for MERGE_OUTPUT AND OPTIMIZE_SIZE. Difference is significant. 5X

KhalilBellakrid:
> > > Hey guys!
> > > Testing this fix.
> > > One thing that is quite suspicious is that I'm getting the same fees for MERGE_OUTPUT and OPTIMIZE_SIZE. I understand that it basically means that the same UTXO are selected.
> > > # I have the following distribution:
> > > 1 UTXO of 0.00664456
> > > 99 UTXO of 0.01 BTC
> > > 2 UTXO of 1 BTC
> > > 1 UTXO of 5 BTC
> > > 2 questions:
> > > 
> > > 1. Is that possible that with OPTIMIZE_SIZE and MERGE_OUTPUTS strategies fees would be the same with the aforementioned distribution?
> > > 2. If strategy is OPTIMIZE_OUTPUT. what UTXOs should be selected when trying to send 0.5 BTC?
> > 
> > 
> > Depending on your UTXO distribution you could have the same for 3 strategies, the most important thing is that `OPTIMIZE_FEES <= DEEP_FEES <= MERGE_FEES` same for number of UTXOs. Another thing is that `OPTIMIZE_SIZE` might not converge and just fallback on `DEEP_FIRST` so ... for all these reasons the main thing to test is above rule.
> 
> Well. In my case I have the smallest fees for DEEP_OUTPUT_FIRST.
> And the same fees for MERGE_OUTPUT AND OPTIMIZE_SIZE. Difference is significant. 5X

Probably what you have is a deepeest UTXO with amount > to amount you're trying to spend.
So actually, there are cases, where it is normal that `DEEP_FIRST` would be advantageous but it will break huge UTXOs. `OPTIMIZE_SIZE` is fallbacking on knapsack solver so gather UTXOs with value under your amount first, here : 0.5 so it should gather 6 UTXOs of 0.6 hence the x5 difference. 

To sum up, the comparison should be made between `OPTIMIZE_SIZE` and `MERGE_OUTPUTS`, keep in mind that this is not just picking the biggest UTXOs and construct a tx with it, it takes into consideration the change amount to see if it's beneficial to create a change or not, so if picking 1 UTXO will create a change of `1000 statoshis` we would prefer pick > 1 UTXOs with no change or a change that worth to create (because 1000sat means that we'll probably spend 5000sat to spend it/ include it in a tx)

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
phaazon: Reusable? How so? Also, why do you want to make such a getter reusable? The problem here comes from the use of C++ tuples. It would be the same in another language, with a different syntax.

I think to either return a `struct`, or… just stick to that `get<1>` if it’s a tuple, which I think is okay for internal code.

PR #561, opened 2020-04-23
LLC-639 — Support origination for bakers.

phaazon:
# Content

This commit filters originated accounts (KT1*) to prevent including accounts that relate to us (delegation) but are not ours (i.e. their manager is someone else).

This case is typical for bakers.

# Mandatory picture of floof


phaazon:
Fix complete; 100% match on all the KT1 accounts I have tested with (even with failed transactions).

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
```diff
@@ -82,19 +92,47 @@ namespace ledger {
...
+
                         case api::OperationType::SEND: {
-                            auto fees = BigInt(op->getFees()->toString());
-                            result = result - (value + fees);
+                            // if sending a transaction has failed:
+                            //
+                            // - Then we only spend the fees pre- Babylon 2.0.
+                            // - We do nothing post- Babylon 2.0 as the fees will be spent by the
+                            //   parent account.
+                            auto isPreBabylon = op->getBlockHeight().value_or(std::numeric_limits<int64_t>::max()) < BABYLON_2_0_ACTIVATION_BLOCK_HEIGHT;
```
alekece, line 114:
```suggestion
                            auto const isPreBabylon = op->getBlockHeight().value_or(std::numeric_limits<int64_t>::max()) < BABYLON_2_0_ACTIVATION_BLOCK_HEIGHT;
```

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
```diff
@@ -131,14 +169,39 @@ namespace ledger {
 
                     static inline void update_balance(std::shared_ptr<api::Operation>& op, BigInt& sum) {
```
alekece, line 170: This method is similar to the inner range-based loop code in `getBalance` method.
Can you put that code into a static method/plain old function to avoid redundant code ?

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
phaazon: Done.

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
phaazon: Even though this is not part of the PR, I agree about the fact it’s the same code. More importantly: it’s duplicated logic. However, we cannot move that code out of the struct because of the agnostic code. However, I have moved the struct definition at the namespace level so that I can re-use it.

PR #564, opened 2020-05-05
Fix BigInt assign scalar

pollastri-pierre:
Current implementation of BigInt::assignScalar doesn't build when passing an unsigned integer because std::abs is not defined for unsigned type (and doesn't make much sense).

This PR add 2 different implementation of BigInt::assignScalar<T> for signed and unsigned type and a test to create BigInt from unsigned scalar types.

### DIFF
--- a/core/src/math/BigInt.h
+++ b/core/src/math/BigInt.h
@@ -43,6 +43,7 @@
 #include <vector>
 #include <bigd.h>
 #include <memory>
+#include <traits/arithmetic.hpp>
 #include "../utils/endian.h"
 
 namespace ledger {
@@ -223,7 +224,15 @@ namespace ledger {
 
             BigInt& assignI64(int64_t value);
 
-            template <typename T>
+            template <typename T, isUnsigned<T> = true>
+            BigInt& assignScalar(T value) {
+                auto bytes = endianness::scalar_type_to_array<T>(value, endianness::Endianness::BIG);
+                bdConvFromOctets(_bigd, reinterpret_cast<const unsigned char *>(bytes), sizeof(value));
+                std::free(bytes);
+                return *this;
+            }
+
+            template <typename T, isSigned<T> = true>
             BigInt& assignScalar(T value) {
                 auto bytes = endianness::scalar_type_to_array<T>(std::abs(value), endianness::Endianness::BIG);
                 bdConvFromOctets(_bigd, reinterpret_cast<const unsigned char *>(bytes), sizeof(value));
--- a/core/test/math/bigint_tests.cpp
+++ b/core/test/math/bigint_tests.cpp
@@ -31,6 +31,7 @@
 
 #include "gtest/gtest.h"
 #include "math/BigInt.h"
+#include <limits>
 
 using namespace ledger::core;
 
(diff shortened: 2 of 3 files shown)


core/src/math/BigInt.h
```diff
@@ -224,7 +224,15 @@ namespace ledger {
             BigInt& assignI64(int64_t value);
 
             template <typename T>
-            BigInt& assignScalar(T value) {
+            typename std::enable_if<std::is_unsigned<T>::value, BigInt&>::type assignScalar(T value) {
```
alekece, line 227: I suggest you to read that [article](https://www.fluentcpp.com/2019/08/23/how-to-make-sfinae-pretty-and-robust/) about how prettify SFINAE.

For instance we can create generic type traits somewhere such as : 
```
template <typename T>
using isSigned = std::enable_if_t<std::is_signed<T>::value, bool>;

template <typename T>
using isUnsigned = std::enable_it_t<std::is_unsigned<T>::value, bool>;
```

then use it like that : 
```
template <typename T, isUnsigned<T> = true>
BigInt& assignScalar(T value)  {
...
}

template <typename T, isSigned<T> = true>
BigInt& assignScalar(T value) {
... 
}
```
The code is more readable and we can reuse these type traits for future implementations.

core/test/math/bigint_tests.cpp
```diff
@@ -176,4 +176,10 @@ TEST(BigInt, FromFloatString) {
     auto t = witness.toByteArray();
     auto tt = subject.toByteArray();
     EXPECT_TRUE(witness.compare(subject) == 0);
+}
+
+TEST(BigInt, AssignFromUnsignedScalar) {
+    uint64_t value = 0xFFFFFFFFFFFFFF01UL;
+    BigInt bigInt = BigInt::fromScalar(value);
+    EXPECT_EQ(value, bigInt.toUint64());
```
alekece, line 184: There is a test somewhere for signed scalar ?

core/src/math/BigInt.h
pollastri-pierre: Nice way! I updated the code.

core/test/math/bigint_tests.cpp
pollastri-pierre: Indeed there is no direct test, I add one

core/src/traits/arithmetic_types_traits.hpp
```diff
@@ -0,0 +1,49 @@
+/*
```
alekece, line 1: I think `arithmetic.hpp` is a proper name since the file is contained in a `traits` folder.

core/test/math/bigint_tests.cpp
```diff
@@ -176,4 +177,16 @@ TEST(BigInt, FromFloatString) {
...
+TEST(BigInt, AssignFromUnsigned64bitScalar) {
+    auto value = std::numeric_limits<uint64_t>::max();
+    BigInt bigInt = BigInt::fromScalar(value);
+    EXPECT_EQ(value, bigInt.toUint64());
+}
+
+TEST(BigInt, AssignFromSigned64bitScalar) {
+    uint64_t value = std::numeric_limits<uint64_t>::min();
+    BigInt bigInt = BigInt::fromScalar(value);
+    EXPECT_EQ(value, bigInt.toUint64());
```
alekece, line 191: Why use `BigInt::fromScalar` instead of `assignScalar` ? I suggest to be as close as possible of the tested method/function in unit test.

PR #565, opened 2020-05-06
LLC–626 Add support for getting recipients relating to one’s account only from an operation.

phaazon:
# Content

See LLC-626 for further details. This commit also adds support for
getting keychains via an abstract interface to check whether an address
belongs to the keychain, as it’s necessary to implement correctly the
getSelfRecipients feature.

Context can be found in LVS-137.

# Mandatory picture of cat boop


core/src/wallet/common/api_impl/OperationApi.cpp
```diff
@@ -69,6 +70,29 @@ namespace ledger {
...
+            std::copy_if(
+                _backend.recipients.cbegin(),
+                _backend.recipients.cend(),
+                std::back_inserter(recipients),
+                [&](std::string const& addr) -> bool
+            {
+                return keychain->contains(addr);
+            });
+
+            return recipients;
```
phaazon, line 93: Maybe do a `shrink_to_fit` here? What do you think @alekece?

phaazon:
I need to add integration tests about that, but I don’t have the data. Anybody with a transaction outputting to different recipients? A contract one should be enough, most of the time.

core/src/wallet/common/api_impl/OperationApi.cpp
alekece: I do not recommand to use `shrink_to_fit` in such a context. The implementation is free to optimize the memory space :

> The request is non-binding, and the container implementation is free to optimize otherwise and leave the vector with a capacity greater than its size.
This may cause a reallocation, but has no effect on the vector size and cannot alter its elements.

So either we perform a reallocation for few bytes or do barely nothing; however the code tell the opposite for the second case which can lead to misunderstanding.

In general, we use `shrink_to_fit` when a container is often modified such as a bunch of addition followed by a bunch of deletion. In such a case, `shrink_to_fit` can save a lot of memory space.

phaazon: We use `shrink_to_fit` to remove memory we know we will never use anymore. This might be important here, as some transactions (especially contracts) might have a _lot_ of outputs, and the user might have only a few (if not a single one) in the list.

If the implementation decides not to shrink, then it’s like a no-op. If it decides to optimize, then we end up with a cleaner memory scheme.

On a more general note, shrinking sizes like that is not recommended in hot and tight loop, but it’s not the case here, so I don’t see why you don’t recommend it.

teams2ua: If we are worrying about memory, why can't we just don't do `reserve` ?

phaazon: I use `reserve` to prevent allocating memory on the fly in the loop, but given the fact the account shouldn’t have a lot of addresses in the list, I think you’re right @teams2ua. I’ll remove the reserve.

PR #566, opened 2020-05-07
LLC-648 + VSD-26 — Fix stack exhaustion with recursive futures in the Bitcoin UTXO picker.

phaazon:
# Content

The problem was due to immediate contexts being used, which stack
fillInputs calls and eventually lead to a segmentation fault. We
detected the problem in VSD-26, leading to LLC-648 after seeing first a
520+-deep callstack on the JVM / Wallet Daemon, and I reproduced locally
by limitting the size of the stack to 1MB on native tests (SIGSEGV after
a 8k+-deep callstack).

This commit doesn’t do anything more than fixing the stack problem. Slow
issues are to be solved in another PR and new issues.

# Mandatory picture of boop boop


phaazon:
@KhalilBellakrid this PR is only about the segfault and, yes, there are issues with performances, but not covered by this PR. I think @RomanWlm might have more information about the exact scenario occurring with the performance issues — I think it can lead to timeouts?

PR #567, opened 2020-05-12
Fix bad block height (ledger index) on XRP integration

pollastri-pierre:
Current implementation of XRP isn't setting a block height which leads to a situation where
all transactions end up in the same block in database and are not retrieved.

The fix is simply to use block height as block hash.

The rational behind this PR is that we can get rid of block hash in this precise case and replace it with block height (ledger index).
Blocks (called ledgers on XRP) cannot be reorganized on the chain, once a block is inserted it remains here therefore keeping the hash is not needed. Another approach would be to individually fetch all ledgers, this would however kill performances for something which is not useful.


phaazon:
Roses are blue,
CI is red,
Le me is sad.

gagbo:
> CI is red,

```
The following tests FAILED:

	118 - ledger-core-integration-CoinSelectionP2PKH.CompareUTXOPickingStrategies (Failed)
```

Might be worth it to run CI again. Also changing the hash of a block *might* change the output to ledger live afterwards. Did we check if LL tests need to be changed as the xrp_synchronization ones here ?

phaazon:
@gagbo that one is a regression introduced lately by #560. I think it can be ignored until we fix it.

pollastri-pierre:
CI is full red, almost the definition of a bad day 😡 .

pollastri-pierre:
> > CI is red,
> 
> ```
> The following tests FAILED:
> 
> 	118 - ledger-core-integration-CoinSelectionP2PKH.CompareUTXOPickingStrategies (Failed)
> ```
> 
> Might be worth it to run CI again. Also changing the hash of a block _might_ change the output to ledger live afterwards. Did we check if LL tests need to be changed as the xrp_synchronization ones here ?

Current implementation is in JS on Ledger Live.

pollastri-pierre:
Hodl this time, CI fails for real..

phaazon:
> Hodl this time, CI fails for real..

core/src/wallet/ripple/explorers/api/RippleLikeBlockParser.h
```diff
@@ -49,14 +49,13 @@ namespace ledger {
...
-            bool String(const rapidjson::Reader::Ch *str, rapidjson::SizeType length, bool copy) {
-                std::string value = std::string(str, length);
-                if (getLastKey() == "close_time_human") {
-                    _block->time = DateUtils::fromJSON(value);
+                } else if (getLastKey() == "close_time") {
+                    std::string number(str, length);
+                    BigInt value = BigInt::fromString(number);
+                    // rippled is using another epoch time than UNIX
+                    // https://xrpl.org/basic-data-types.html#specifying-time
+                    std::chrono::system_clock::time_point time(std::chrono::seconds(value.toUint64() + 946684800));
```
alekece, line 57: Do that magic number is used somewhere else ? In that case we probably should store it within `constexpr` variable.

core/src/wallet/ripple/utils/RippleLikeUtils.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+#include "RippleLikeUtils.hpp"
+
+namespace {
+    constexpr uint64_t kRippleEpochSeconds = 946684800;
```
phaazon, line 35: k? What is that new convention? :D

core/src/wallet/ripple/utils/RippleLikeUtils.hpp
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#ifndef LEDGER_CORE_RIPPLELIKEUTILS_HPP
+#define LEDGER_CORE_RIPPLELIKEUTILS_HPP
+
+#include <chrono>
+
+namespace ledger {
+    namespace core {
+        class RippleLikeUtils {
```
phaazon, line 39: Why not simply an `xrp` namespace?

core/src/wallet/ripple/utils/RippleLikeUtils.cpp
alekece: I think I am the one introducing that convention when I was working on Cosmos integration. 
It's a old convention but effective to avoid name clashing in local TU; a other way is to use upper snake case but I think this one should remain for macro

pollastri-pierre: Yep, sorry I'm working on a  C++ side project for friends on mine, and it's part of their convention I got used to it. I change it to upper snake.

core/src/wallet/ripple/utils/RippleLikeUtils.hpp
pollastri-pierre: I'm for it, but what's the new convention regarding file name. Previously I put free function in a file named in lower case (C like). What's the convention now ?

core/src/wallet/ripple/utils/RippleUtils.hpp
```diff
@@ -0,0 +1,54 @@
...
+        namespace xrp_utils {
+            /**
+             * Convert a XRP timestamp to a C++ time point. Rippled is using another epoch time than UNIX
+             * therefore it needs to be translated to UNIX timestamp before converting it to time point.
+             * https://xrpl.org/basic-data-types.html#specifying-time
+             * @param timestamp A XRP timestamp
+             * @param time Reference to the time point to inflate.
+             * @return The translated time point
+             */
+             void xrpTimestampToTimePoint(uint64_t timestamp, std::chrono::system_clock::time_point &time);
```
alekece, line 48: There are a lot of redundant information in function/file name :
* `ripple/utils/RippleUtils` -> `ripple/utils/Time.*` : the source code is related to time helper
* `xrpTimestampToTimePoint` -> `toTimePoint`

I also suggest some improvement to make the plain function generic enough.
```suggestion
             template<typename T>
             using IsClock = std::enable_if_t<std::is_same<T, std::chrono::system_clock>::value ||
                                              std::is_same<T, std::chrono::high_resolution_clock>::value ||
                                              std::is_same<T, std::chrono::steady_clock>::value, bool>;
                                                                 
             template <typename T, IsClock<T> = true>
             std::chrono::time_point<T> toTimePoint(uint64_t timestamp)
             {
                return std::chrono::time_point<T>{std::chrono::seconds(timestamp + RIPPLE_EPOCH_SECONDS));
             }
```

phaazon:
> We're done ! There is only one thing we can add to that PR but not really essential :
> 
> ```
> auto toSystemTimePoint(uint64_t timestamps)
> {
>    return toTimePoint<std::chrono::system_clock>(timestamps);
> }
> ```

Please don’t use `auto` in type signature. It makes the code harder for people to read and guess what types are. If you have to read the body of a function to understand what it does: it’s wrong.

PR #568, opened 2020-05-18
Add wrapper for SHA-512/256 crypto algorithm

haammar-ledger:
Add support of SHA-512/256 algorithm to libcore crypto capabilities.

:information_source: Note that SHA-512/256 is a different algorithm from SHA-512 and SHA-256. It may be required (to be confirmed) for Algorand integration.

:information_source: SHA-512/256 requires OpenSSL 1.1.1 or above, so the implementation is preprocessor-directive-protected.

:warning: On my machine, `design/modularization` doesn't compile without the following modification in `ledger-core/test/integration/CMakeLists.txt`:
```
-    target_link_libraries(ledger-core-integration-test OpenSSL::crypto)
+    target_link_libraries(ledger-core-integration-test OpenSSL::Crypto)
```
Not sure this change is normal, so I'm just mentioning it here.

### DIFF
--- a/ledger-core/inc/core/crypto/SHA512256.hpp
+++ b/ledger-core/inc/core/crypto/SHA512256.hpp
@@ -1,0 +1,48 @@
+/*
+ * SHA512256
+ *
+ * Created by Hakim Aammar on 20/04/2020.
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020 Ledger
+ *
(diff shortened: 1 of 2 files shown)


alekece:
@haammar-ledger I confirm you the modification is normal; this is something we spot out few days ago

haammar-ledger:
Discarded: this was required for Algorand, but since OpenSSL 1.1.1 is not currently available, we have implemented a workaround that will live in the coin module.

PR #569, opened 2020-05-18
Port support of both OpenSSL 1.0.0 & 1.1.0 for HMAC from libcore v1 to v2

haammar-ledger:
Port to libcore v2 the support of both OpenSSL 1.0.0 & 1.1.0 for HMAC.

This should be the same changes in `HMAC.cpp` as in https://github.com/LedgerHQ/lib-ledger-core/commit/64247e992fcc2447565628b6469ae4acb1d86fb1


phaazon:
I had done that in another branch for the packaging, but I think we can merge that PR is it’s minimal.

PR #571, opened 2020-05-20
[cosmos] Return empty derivationPath for CosmosLikeAddress

gagbo:
This *might* fix an iOS specific error

### DIFF
--- a/core/src/cosmos/CosmosLikeAddress.cpp
+++ b/core/src/cosmos/CosmosLikeAddress.cpp
@@ -81,7 +81,7 @@ std::string CosmosLikeAddress::toBech32()
 
 std::experimental::optional<std::string> CosmosLikeAddress::getDerivationPath()
 {
-    return {};  // _derivationPath.toOptional();
+    return _derivationPath.toOptional();
 }
 
 std::string CosmosLikeAddress::toString()
--- a/core/src/wallet/cosmos/factories/CosmosLikeKeychainFactory.cpp
+++ b/core/src/wallet/cosmos/factories/CosmosLikeKeychainFactory.cpp
@@ -41,6 +41,20 @@ std::shared_ptr<CosmosLikeKeychain> CosmosLikeKeychainFactory::build(
     const std::shared_ptr<Preferences> &accountPreferences,
     const api::Currency &currency)
 {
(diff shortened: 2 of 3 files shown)


gagbo:
This *also* recreates an old bug we quick-fixed earlier :sob:

gagbo:
Final decision : see commit message. Basically, we just never return `None`, but return `Some("")` like it was decided on Tezos.

core/src/cosmos/CosmosLikeAddress.cpp
```diff
@@ -81,7 +81,7 @@ std::string CosmosLikeAddress::toBech32()
 
 std::experimental::optional<std::string> CosmosLikeAddress::getDerivationPath()
 {
-    return {};  // _derivationPath.toOptional();
+    return _derivationPath.toOptional();
```
alekece, line 84: The old line of code do the exact same thing, right ? What is the rational behind this change ?

core/src/wallet/cosmos/factories/CosmosLikeKeychainFactory.cpp
```diff
@@ -41,6 +41,20 @@ std::shared_ptr<CosmosLikeKeychain> CosmosLikeKeychainFactory::build(
     const std::shared_ptr<Preferences> &accountPreferences,
     const api::Currency &currency)
 {
+    // NOTE : The keychain is supposed to be linked to an account, but this check
+    // makes the very strong assumption that the keychain is supposed have the pubkey / path
```
alekece, line 45:
```suggestion
    // makes the very strong assumption that the keychain is supposed to have the pubkey / path
```

core/src/cosmos/CosmosLikeAddress.cpp
gagbo: _derivationPath is constructed to be `Some("")` instead of `None` by the Keychain, so the change here is to make that upstream change visible in the interface

gagbo: >  instead of ~~None~~

instead of the old "full Path" that we had to override here with a `None` to prevent bad account construction in Ledger Live*

PR #572, opened 2020-05-21
Fix tezos infinite synchronization

pollastri-pierre:
Context
--------
If a tezos has more than 100 operations and synchronization token is deactivated in configuration.
The infinite loop comes from the fact that Tezos is using the session object to paginate its API 
calls. However since DEACTIVATE_SYNC_TOKEN bypasses the session creation it will prevent 
the synchronizer to paginate and will fetch the same page over and over again.

Fix
---
Beware this PR is a dirty fix to the problem. It just overrides DEACTIVATE_SYNC_TOKEN in the configuration object. 
A proper fix would be to get rid of AbstractBlockchainExplorer, this inheritance brings very few advantage and make synchronization code for Tezos hard to understand. Most parts of AbstractBlockchainExplorer (and synchronizer) was taken from the BTC implementation which is more complex. Tezos seems need an easy synchronization algorithm. Moreover configuration for BTC doesn't fit actual needs for Tezos.


core/src/wallet/tezos/TezosLikeWallet.cpp
```diff
@@ -68,6 +68,10 @@ namespace ledger {
             _observer = observer;
             _keychainFactory = keychainFactory;
             _synchronizerFactory = synchronizer;
+            // TODO This is a dirty patch because Tezos uses the AbstractBlockchainExplorer and synchronizer
+            // Most of the code of these classes were made for BTC and offer very few advantage for simpler
+            // currencies (mainly it avoid rewriting a reorg case)
+            configuration->putBoolean(api::Configuration::DEACTIVATE_SYNC_TOKEN, false);
```
phaazon, line 74: Hm, seems fine to me but do we have a better plan to clean this?

pollastri-pierre: Either we can fine tune AbstractBlockchainExplorer to handle this through inheritance or configuration. Which has the advantage of avoiding rewriting lot of stuffs.

Or we remove AbstractBlockchainExplorer and rewrite some parts to make this class standalone. I'm more for rewriting, because the benefits of having a common ground for BTC, ETH, XTZ is not very high and explorer/synchronizer becomes difficult to read even when XTZ synchronization seems to be pretty straightforward.

PR #573, opened 2020-05-22
[cosmos] Handle fees properly in balanceHistory

gagbo:
- The fees need to be properly computed on inflation if we have the
actual gasUsed in the transaction.
- The fees are always included as an operation.amount so the
balanceHistory from CosmosLikeAccount should never care about
operation.fees from the received operationQuery

### DIFF
--- a/core/src/wallet/cosmos/CosmosLikeAccount.cpp
+++ b/core/src/wallet/cosmos/CosmosLikeAccount.cpp
@@ -41,6 +41,7 @@
 #include <database/soci-option.h>
 #include <events/Event.hpp>
 #include <math/Base58.hpp>
+#include <numeric>
 #include <soci.h>
 #include <utils/DateUtils.hpp>
 #include <utils/Option.hpp>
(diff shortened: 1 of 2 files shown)


core/src/wallet/cosmos/CosmosLikeAccount.cpp
```diff
@@ -318,13 +320,20 @@ void CosmosLikeAccount::inflateOperation(
...
-        std::for_each(tx.fee.amount.begin(), tx.fee.amount.end(), [&](cosmos::Coin amount) {
+        std::for_each(tx.fee.amount.begin(), tx.fee.amount.end(), [&](const cosmos::Coin &amount) {
             assert(amount.denom == "uatom");  // FIXME Temporary until all units correctly supported
             fees += BigInt::fromDecimal(amount.amount).toInt();
         });
     }
+    // The value is updated to the fees actually paid if the transaction has a GasUsed value to use.
+    if (fees != 0 && tx.gasUsed) {
+        auto feeConsumptionRatio = static_cast<float>(tx.gasUsed.getValue().toInt()) /
+                                   static_cast<float>(tx.fee.gas.toInt());
```
alekece, line 334: Potential division by zero detected here ! 
Can `tx.fee.gas` be zero ? Even if it's not the case now, this invariant can be change in the future and I suggest to protect the code from that UB.

core/src/wallet/cosmos/CosmosLikeAccount.cpp
```diff
@@ -318,13 +320,20 @@ void CosmosLikeAccount::inflateOperation(
...
     if (cosmos::stringToMsgType(msg.type.c_str()) == api::CosmosLikeMsgType::MSGFEES) {
-        std::for_each(tx.fee.amount.begin(), tx.fee.amount.end(), [&](cosmos::Coin amount) {
+        std::for_each(tx.fee.amount.begin(), tx.fee.amount.end(), [&](const cosmos::Coin &amount) {
             assert(amount.denom == "uatom");  // FIXME Temporary until all units correctly supported
             fees += BigInt::fromDecimal(amount.amount).toInt();
         });
     }
+    // The value is updated to the fees actually paid if the transaction has a GasUsed value to use.
+    if (fees != 0 && tx.gasUsed) {
+        auto feeConsumptionRatio = static_cast<float>(tx.gasUsed.getValue().toInt()) /
```
alekece, line 333:
```suggestion
        auto const feeConsumptionRatio = static_cast<float>(tx.gasUsed.getValue().toInt()) /
```

core/src/wallet/cosmos/CosmosLikeAccount.cpp
gagbo: Theoretically it cannot be. A 0 gas transaction is either :
- Never broadcast by the node (because of spam protection mechanisms)
- Invalid, as the signature verification (or the `gas per byte` cost) will immediately trigger an `OutOfGas` network error.

An invalid or absent transaction from the blockchain is not picked up by the explorer, so Operations cannot be inflated with such transactions. I'll add a throw branch with a comment for these cases

alekece: I think it's an additional reason to check the zero value : it's both functional and technical potential bug.

gagbo:
I just reviewed all the changes and found a small bug in the convention used for `Operation::amount` vs `Operation::fees` values for the `Cosmos::MsgFees` operations. It should be clean now.

PR #574, opened 2020-05-25
[cosmos] Prevent trying to index non-arrays in JSON parsers

gagbo:
The API might give a `null` object or an empty list for a few keys that
are supposed to be arrays. Therefore we need to guard these accesses to
avoid segmentation faults.

The parser functions will leave the values in the target structures
unchanged, probably equal to the default-constructed values.

Typical example
`/distribution/validators/{cosmosvaloperAddress}` usually returns
```
curl "https://cosmos.coin-proxy.dev.aws.ledger.fr/distribution/validators/cosmosvaloper1clpqr4nrk4khgkxj78fcwwh6dl3uw4epsluffn" | jq
{
  "height": "2017695",
  "result": {
    "operator_address": "cosmos1clpqr4nrk4khgkxj78fcwwh6dl3uw4ep4tgu9q",
    "self_bond_rewards": [
      {
        "denom": "uatom",
        "amount": "6150868794.599042500050577296"
      }
    ],
    "val_commission": [
      {
        "denom": "uatom",
        "amount": "4316080153.204048262984009636"
      }
    ]
  }
}

```
but it also might return : 
```
curl "https://cosmos.coin-proxy.dev.aws.ledger.fr/distribution/validators/cosmosvaloper1fhr7e04ct0zslmkzqt9smakg3sxrdve6ulclj2" | jq
{
  "height": "2017348",
  "result": {
    "operator_address": "cosmos1fhr7e04ct0zslmkzqt9smakg3sxrdve6etv27e",
    "self_bond_rewards": [
      {
        "denom": "uatom",
        "amount": "14874758.445311305213563000"
      }
    ],
    "val_commission": null
  }
}
```


gagbo:
HODL : one array made it through somehow

gagbo:
> May you put the remaining raw strings into constant variables ?
No problem, I think everything is covered now, or at least most of it.

> ```
> const auto &balances = node[kCoins].GetArray();
> parseCoinVector(balances, account.balances);
> ```
We are just manipulating const references in any case, but I opted for this syntax where possible to reduce the cognitive complexity of the `parse*` calls in this file.


I also added a small fix for the `CosmosLikeReward` API, since now it can receive an empty structure as data attribute.

PR #579, opened 2020-06-02
Create synchronization result for NRT project

alekece:
This PR aims to introduce a synchronization result available from `ProgressNotifier` - which is already returned when a synchronization is performed but with a dummy `Unit` object - and from `Event` for bitcoin only for now.

The result contains the block height from where the reorganization started (this value is obviously optional and will not be added in the payload event in some case), the last block height and the numbers of new operations.

The `FLAG_TRANSACTION_CREATED_EXTERNAL_OPERATION` value is changed because all the flags specified in coin are moved to `api::Account` and the old value was already taken by a previous flag. So far the code use the flag value for comparison and assignment but we should probably check the usage in bindings.

# Mandatory picture of a cat
bitten_cat

**_The tests are missing and we should find a proper way to test a deterministic synchronization_**


core/idl/wallet/common/wallet.djinni
```diff
@@ -333,8 +333,17 @@ TimePeriod = enum {
 # Class representing an account.
 Account = interface +c {
     # Key of the synchronization duration time in the synchronize event payload.
-    # The value is stored in a int 64 time expressed in miliseconds.
+    # The value is stored in a int64 time expressed in miliseconds.
     const EV_SYNC_DURATION_MS: string = "EV_SYNC_DURATION_MS";
+    # Indicates the last block height fetch during synchronization
+    # The value is stored as a int32
+    const EV_SYNC_LAST_BLOCK_HEIGHT: string = "EV_SYNC_LAST_BLOCK_HEIGHT";
```
RomanWlm, line 340: Super, this is our consistency point, we can specify to the client last block height.
Can you clarify here (comment) if we are talking about **block height from most recent operation** related to the account or about **last height known by explorer** at the synchronization time ?

core/idl/wallet/common/wallet.djinni
```diff
@@ -348,6 +357,20 @@ Account = interface +c {
     const EV_NEW_OP_WALLET_NAME: string = "EV_NEW_OP_WALLET_NAME";
     const EV_NEW_OP_ACCOUNT_INDEX: string = "EV_NEW_OP_ACCOUNT_INDEX";
     const EV_NEW_OP_UID: string = "EV_NEW_OP_UID";
+
+    # Returned flag when a transaction is put in DB
+    # Some of those are unrelevant for specific coins
+    # but for the sake of generic mechanism we put all those
+    # flags together
+    const FLAG_TRANSACTION_IGNORED: i32 = 0;
```
RomanWlm, line 365: When this is used ? when there is no new transaction to retrieve ?

core/idl/wallet/common/wallet.djinni
```diff
@@ -348,6 +357,20 @@ Account = interface +c {
     const EV_NEW_OP_WALLET_NAME: string = "EV_NEW_OP_WALLET_NAME";
     const EV_NEW_OP_ACCOUNT_INDEX: string = "EV_NEW_OP_ACCOUNT_INDEX";
     const EV_NEW_OP_UID: string = "EV_NEW_OP_UID";
```
RomanWlm, line 359: Is it possible to retrieve UID when operation is removed from DB when a reorg happen ??

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -84,22 +85,28 @@ namespace ledger {
             }
         };
 
+        struct BlockchainExplorerAccountSynchronizationResult {
+            Option<uint32_t> reorgBlockHeight;
+            uint32_t lastBlockHeight;
+            uint32_t newOperations;
+        };
```
RomanWlm, line 92: Ideally we would like to have the list of removed operations during sync here inside this result payload **or** by Event Bus means

core/idl/wallet/common/wallet.djinni
alekece: It's the default value used by all coins when we try to put a transaction on an account

alekece: I cannot certify that the meaning is the same for all the coins

core/idl/wallet/common/wallet.djinni
alekece: I will add that as soon as possible

core/idl/wallet/common/wallet.djinni
luchenyuxx: @RomanWlm you mean this message indicate a reorg point? I am not very clear how it works.

core/idl/wallet/common/wallet.djinni
```diff
@@ -333,8 +333,17 @@ TimePeriod = enum {
...
     # Key of the synchronization duration time in the synchronize event payload.
-    # The value is stored in a int 64 time expressed in miliseconds.
+    # The value is stored in a int64 time expressed in miliseconds.
     const EV_SYNC_DURATION_MS: string = "EV_SYNC_DURATION_MS";
+    # Indicates the last block height fetch during synchronization
+    # The value is stored as a int32
+    const EV_SYNC_LAST_BLOCK_HEIGHT: string = "EV_SYNC_LAST_BLOCK_HEIGHT";
+    # Indicates the number of new operations during the synchronization
+    # The value is stored as a int32
+    const EV_SYNC_NEW_OPERATIONS: string = "EV_SYNC_NEW_OPERATIONS";
```
luchenyuxx, line 343: When is this message been sent? Is it once per sync?

alekece: Exactly, at the end of the sync

luchenyuxx:
Could we also send event about the deleted transaction and blocks in case of reorg?
Vault may need it

core/idl/wallet/common/wallet.djinni
alekece: We're talking about the **last block height from most recent operation** related to the account here

alekece:
> Could we also send event about the deleted transaction and blocks in case of reorg?
> Vault may need it

I am adding currently the list of deleted operations UIDs to the event at the end of the sync. Concerning the blocks in case of reog, you can already found it with `EV_SYNC_REORG_BLOCK_HEIGHT` @luchenyuxx

luchenyuxx:
@alekece Do we have any progress on this ticket?

PR #582, opened 2020-06-05
Update SBT to version 1.3.10 on the Debian CI.

phaazon:
That commit should fix the situation with failing sbt downloads.

### DIFF
--- a/.circleci/setup_macos.sh
+++ b/.circleci/setup_macos.sh
@@ -22,7 +22,6 @@ if [ "$BUILD_CONFIG" == "Release" ]; then
 	brew install awscli
 	echo "========> Install sbt"
 	brew install sbt
-	sbt sbtVersion
 fi
 
 echo "========> Install C++ dependencies"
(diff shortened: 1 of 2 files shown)


phaazon:
Looks like this doesn’t solve our problems.

phaazon:
Looks like the `sbt` problem is now gone. However, we’re still failing with network issues with explorers… It should be fine to merge this PR though as it fixes what it intends to.

.circleci/setup_macos.sh
```diff
@@ -22,7 +22,7 @@ if [ "$BUILD_CONFIG" == "Release" ]; then
 	brew install awscli
 	echo "========> Install sbt"
 	brew install sbt
-	sbt sbtVersion
+	#sbt sbtVersion
```
pollastri-pierre, line 25:
```suggestion
```
since we don't use it we can remove the line I think.

phaazon: Sure.

phaazon: Done.

PR #587, opened 2020-06-09
[cosmos] Add support for ExtendedKeyAccountCreationInfo

gagbo:
This support is necessary for Live to handle synchronizations from
"cleaned cache"


gagbo:
This can be also fixed on Live side, so feel free to close (to be discussed later)

PR #591, opened 2020-06-10
Parallelize ctest in CI

gagbo:
## Purpose

Got to fail faster.

## Changes

- [x] Run all ctest in CI in parallel
- [x] Only produce the output from tests if failing (otherwise it gets messy really quickly)
- [x] Randomize the uids in tests : `pool` names, `wallet` names so that no database conflicts occur when reading/writing from sqlite databases
- [ ] Add pragmas to sqlite3Backend to help filesystem handle the additional load on Disk I/O coming from parallel tests
- [x] Use `WalletPool::eraseDataSince` as much as possible instead of using `freshResetAll` to reduce filesystem load *and* to prevent race conditions on PgSQL backend that use a single `database`

## Known issues (aka HODL status)

- [x] GetAccountAfterPoolReopen fails : `newDefaultPool(poolName)` with an already used `poolName` fails to get the older database. But only in parallel runs. This means all tests using a second call to `newDefaultPool` to check wallets/operations are "additional failures" compared to sequential runs. This is under investigation
- [ ] The real bottleneck here is filesystem, sqlite3Backend currently does not fail gracefully if it needs to wait for filesystem/Disk bandwidth.
- [ ] The sqlite3 backend performance pragmas should only be enabled in the tests

## Notes

- This is a draft, just to see if we encounter the same I/O problems in CI environments
- Testing on PostgresQL backend : tests that can switch automatically from sqlite to postgresQL do the switch and never fail. The test database is not randomized (because it needs to be created Ahead Of Time) so to avoid issues with `freshResetAll()` racing against `WalletPool::newInstance()` the use of the former one has been reduced as much as possible (only one call left in the test that actually has `FreshResetAll` in its name) and `eraseDataSince(epoch)` is used instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,7 +138,7 @@ jobs:
             command: |
                 cd ../lib-ledger-core-build
 
-                ~/cmake-3.16.5-Linux-x86_64/bin/ctest -VV -timeout 60
+                ~/cmake-3.16.5-Linux-x86_64/bin/ctest -j 5 --output-on-failure -timeout 60
   build_macos_release:
     macos:
       xcode: "10.0.0"
@@ -232,7 +232,7 @@ jobs:
           name: Run_Tests
           command: |
               cd ../lib-ledger-core-build
-              ctest -VV -timeout 60
+              ctest -j 5 --output-on-failure -timeout 60
   publish_jar:
     macos:
       xcode: "10.0.0"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -129,7 +129,7 @@ build_script:
 - if %TOOL_CONFIG% == vs2017  if %BUILD_CONFIG% == Release cp "..\lib-ledger-core-build\%CRYPTO_DLL_PATH%" ".\%LIB_VERSION%\win\%TOOL_CONFIG%" || echo "FAIL TO COPY CRYPTO DLL"
 
 #Launch tests
-- if %BUILD_CONFIG% == Debug ctest --output-on-failure -VV
+- if %BUILD_CONFIG% == Debug ctest -j 4 --output-on-failure
 
 #Build is failing because of configuration
 ##Create node module
--- a/core/idl/wallet/cosmos/wallet.djinni
+++ b/core/idl/wallet/cosmos/wallet.djinni
@@ -223,7 +223,7 @@ CosmosLikeRedelegationEntry = interface +c {
 }
 
 CosmosConfigurationDefaults = interface +c {
-        const COSMOS_DEFAULT_API_ENDPOINT: string = "https://cosmos.coin.ledger.com";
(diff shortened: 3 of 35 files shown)


gagbo:
This is the kind of things that happen when the tests go too fast :

```
[ RUN      ] WalletPoolTest.InitializeCurrencies
 
unknown file: Failure
 
C++ exception with description "IO error: /root/lib-ledger-core-build/core/test/integration/build/preferences___2859ef56-f757-408e-bfc5-69b73f03a5ca__preferences.db/000001.dbtmp: No such file or directory" thrown in the test body.
 
TEAR DOWN
 
[  FAILED  ] WalletPoolTest.InitializeCurrencies (42 ms)
```

Most other errors are related to the new pools not finding the old databases/wallets when taking an already existing name.

It looks like this for the record :
```
[ RUN      ] WalletTests.GetAccountAfterPoolReopen
 
2020-06-10T09:42:33Z+00:00 I: Building wallet instance '67cb2b2b-ef3d-4a85-951a-4c27b3866af4' for bitcoin with parameters: 
 
unknown file: Failure
 
C++ exception with description "Wallet '67cb2b2b-ef3d-4a85-951a-4c27b3866af4' doesn't exist." thrown in the test body.
 
TEAR DOWN
 
[  FAILED  ] WalletTests.GetAccountAfterPoolReopen (3086 ms)
```

There's also
- the regression on UTXO, 
- a stellar fix that did not get rebased on correctly
- a bug probably related to the uid generation system for operations 
```
[ RUN      ] QueryBuilderTest.SimpleOperationQuery
 
2020-06-10T09:43:12Z+00:00 I: Building wallet instance '990ff34c-949a-46ec-aa43-1063d85ff37f' for bitcoin with parameters: 
 
/root/lib-ledger-core/core/test/database/query_builder_tests.cpp:90: Failure
 
Expected equality of these values:
 
  count
 
    Which is: 0
 
  1
 
[  FAILED  ] QueryBuilderTest.SimpleOperationQuery (2212 ms)
```

pollastri-pierre:
Great PR!

gagbo:
Will do ! I'll revert the cosmos default node address too, I thought it'd be up by now but it is not

gagbo:
I just added comments to point to the RFC for the implementation and explaining the binary representations of the random number generated.

The main reason I chose to use UUIDv4 is that it has a specification, and it's very clear (when you know to look for the version byte in the UUID), that : 
- it's chosen at random (with 122 bits of entropy ; 6 are fixed in uuid v4)
- it's meant to be a unique identifier

Those are the 2 only properties I wanted (unique, at random) I wanted to get, and following a known RFC for these constraints help reader to know what was wanted

twilgenbus-ledger:
@gagbo re-did that when integrating Nix, blast from the past.

PR #599, opened 2020-06-15
[cosmos] Fetch block data on each transaction fetch

gagbo:
If we do not fetch the block data each time a transaction is
fetched (although saving HTTP calls), then :

- The blockHash of the transaction is empty, therefore no block_uid can
be computed from the tx.block
- The block_uid column of the transaction in the database therefore
stays `NULL`
- When querying the operations later, the LEFT JOIN of
cosmos_transactions with blocks will fail to find a block
- Final state : the queried operation has no "height" information

Having no "height" information is a problem because this is used by
Ledger Live to compute the count of confirmations for a given Operation.

### DIFF
--- a/core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
@@ -98,7 +98,7 @@ const std::vector<CosmosLikeBlockchainExplorer::TransactionFilter>
     return GAIA_FILTER;
 }
 
-FuturePtr<cosmos::Block> GaiaCosmosLikeBlockchainExplorer::getBlock(uint64_t &blockHeight)
+FuturePtr<cosmos::Block> GaiaCosmosLikeBlockchainExplorer::getBlock(uint64_t &blockHeight) const
 {
     return _http->GET(fmt::format(kGaiaBlocksEndpoint, blockHeight), ACCEPT_HEADER)
         .json(true)
@@ -212,24 +212,28 @@ void addMsgFeesTo(cosmos::Transaction &transaction, const T &node)
     transaction.messages.push_back(msgFees);
 }
 
+}  // namespace
+
 template <typename T>
-void parseTransactionWithPosttreatment(const T &node, cosmos::Transaction &transaction)
+void GaiaCosmosLikeBlockchainExplorer::parseTransactionWithPosttreatment(
+    const T &node, cosmos::Transaction &transaction) const
 {
     rpcs_parsers::parseTransaction(node, transaction);
     addMsgFeesTo(transaction, node);
 }
 
-}  // namespace
 
 FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransactions(
     const CosmosLikeBlockchainExplorer::TransactionFilter &filter, int page, int limit) const
 {
+    // NOTE : the amount of memory involved here asks for a second pass to make
+    // sure that the least amount of copies is done.
     return _http->GET(fmt::format(kGaiaTransactionsWithPageLimitEnpoint, filter, page, limit), ACCEPT_HEADER)
         .json(true)
-        .mapPtr<cosmos::TransactionsBulk>(
-            getContext(), [](const HttpRequest::JsonResult &response) {
-                auto result = std::make_shared<cosmos::TransactionsBulk>();
-
+        .flatMap<cosmos::TransactionsBulk>(
+            getContext(), [this](const HttpRequest::JsonResult &response)
+            -> Future<cosmos::TransactionsBulk> {
+                cosmos::TransactionsBulk result;
                 const auto &document = std::get<1>(response)->GetObject();
                 if (!document.HasMember("txs") || !document[kTxArray].IsArray()) {
                     throw make_exception(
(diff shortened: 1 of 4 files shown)


core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
-                return result;
-            });
+                return Future<cosmos::TransactionsBulk>::successful(result);
+                })
+                .flatMapPtr<cosmos::TransactionsBulk>(
+                getContext(),
+                [this](const cosmos::TransactionsBulk& inputBulk) mutable -> FuturePtr<cosmos::TransactionsBulk> {
+                    auto retval = inputBulk;
+                    auto inputTxs = inputBulk.transactions;
+                    std::vector<FuturePtr<cosmos::Transaction>> filledTxsFutures;
```
alekece, line 270:
```suggestion
                    std::vector<FuturePtr<cosmos::Transaction>> filledTxsFutures;
                    filledTxsFutures.reserve(inputTxs.size());
```

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
                 }
-                return result;
-            });
+                return Future<cosmos::TransactionsBulk>::successful(result);
+                })
+                .flatMapPtr<cosmos::TransactionsBulk>(
+                getContext(),
+                [this](const cosmos::TransactionsBulk& inputBulk) mutable -> FuturePtr<cosmos::TransactionsBulk> {
+                    auto retval = inputBulk;
+                    auto inputTxs = inputBulk.transactions;
```
alekece, line 269: The copy is necessary here ? If I understand correctly that variable is more an variable alias than something else.

```suggestion
                    auto const &inputTxs = inputBulk.transactions;
```

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
+                        inputTxs.cbegin(),
+                        inputTxs.cend(),
+                        std::back_inserter(filledTxsFutures),
+                        [this](const cosmos::Transaction &inputTx) {
+                            return this->inflateTransactionWithBlockData(inputTx);
+                        });
+                    return async::sequence(getContext(), filledTxsFutures)
+                        .flatMapPtr<cosmos::TransactionsBulk>(
+                            getContext(),
+                            [retval](
```
alekece, line 281:
```suggestion
                            [inputBulk = std::move(inputBulk)](
```

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
+                    result.hasNext = (count < total_count);
                 }
-                return result;
-            });
+                return Future<cosmos::TransactionsBulk>::successful(result);
+                })
+                .flatMapPtr<cosmos::TransactionsBulk>(
+                getContext(),
+                [this](const cosmos::TransactionsBulk& inputBulk) mutable -> FuturePtr<cosmos::TransactionsBulk> {
+                    auto retval = inputBulk;
```
alekece, line 268: You can remove that variable if you follow my suggestion below (capturing by move semantic in the lambda)

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
+                        std::back_inserter(filledTxsFutures),
+                        [this](const cosmos::Transaction &inputTx) {
+                            return this->inflateTransactionWithBlockData(inputTx);
+                        });
+                    return async::sequence(getContext(), filledTxsFutures)
+                        .flatMapPtr<cosmos::TransactionsBulk>(
+                            getContext(),
+                            [retval](
+                                const std::vector<std::shared_ptr<cosmos::Transaction>> &filledTxsList) mutable {
+                                retval.transactions.clear();
```
alekece, line 283: The reallocation isn't guarantee with `clear` method. If you really need to free some memory space you can use swap method with an empty vector.

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
+                                        return *filledTx;
+                                    });
+                                return FuturePtr<cosmos::TransactionsBulk>::successful(
+                                    std::make_shared<cosmos::TransactionsBulk>(retval));
+                            });
+                });
+}
+
+FuturePtr<cosmos::Transaction> GaiaCosmosLikeBlockchainExplorer::inflateTransactionWithBlockData(const cosmos::Transaction& inputTx) const {
+            auto retval = cosmos::Transaction(inputTx);
```
alekece, line 298: What `retval` is ? Can you use proper variable name here ?

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -241,36 +245,86 @@ FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransac
...
+            if (!retval.block) {
+                return FuturePtr<cosmos::Transaction>::successful(
+                    std::make_shared<cosmos::Transaction>(std::move(retval)));
+            }
+            auto height = retval.block.getValue().height;
+            return getBlock(height).flatMapPtr<cosmos::Transaction>(
+                getContext(), [retval](const auto &blockData) mutable -> FuturePtr<cosmos::Transaction> {
```
alekece, line 305:
```suggestion
                getContext(), [retval = std::move(retval)](const auto &blockData) mutable -> FuturePtr<cosmos::Transaction> {
```

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
gagbo: Yes will do ; I used to have code invalidating that reference closed to it earlier, I think that's why I had a copy, and forgot to change it after

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
gagbo: It's not really for the memory space that I do it ; I'm not sure I can actually use `transform` to mutate the content of `retval/inputBulk.transactions`, so instead I just clear the vector and use the back_inserter. I don't know if there's a better way to do it, maybe with a swap in the transform function

alekece: So you can do something like that :
```
retval.transactions.erase(
   std::transform(filledTxsList.cbegin(), filledTxsList.cend(), retval.transactions.begin(), ...),
   retval.transactions.end());
```

alekece: Oh wait ! Since `filledTxsList` has the same size that `retval.transactions` - you construct the first container with the second one - you can skip the `erase` call which is result to a no operation.

gagbo: Seems good :)

PR #601, opened 2020-06-18
Add a PreferencesBackend api

barjonr:
* I did not "duplicate" the `newInstance` method in `WalletPool` api. Maybe it would be better to add a second one with the two `PreferencesBackend` not to impact clients of the current code.


pollastri-pierre:
Why is the CI blocked again ><

core/idl/preferences/preferences.djinni
```diff
@@ -74,3 +76,58 @@ PreferencesEditor = interface +c {
...
+}
+
+PreferencesChange = record {
+    type: PreferencesChangeType;
+    key: binary;
+    value: binary;
+}
+
+# Interface describing the behaviour of the backend used by Preferences.
+PreferencesBackend = interface +c {
```
alekece, line 92: Add `const` qualifier on all getters

core/idl/preferences/preferences.djinni
```diff
@@ -74,3 +76,58 @@ PreferencesEditor = interface +c {
     # Clear all preferences.
     clear();
 }
+
+PreferencesChangeType = enum {
+    PUT_TYPE;
+    DELETE_TYPE;
```
alekece, line 82: The type suffix is redundant with the enumeration name
```suggestion
    PUT;
    DELETE;
```

barjonr: I'll do this

core/idl/preferences/preferences.djinni
barjonr: You are right

barjonr:
> Thanks for that work ! Can you explain the rational behind the deletion of `iterate` method ?

They were not referenced anywhere in the libcore (actually, once, in a test) and were difficult to add to the djinni API as they used `std::function` parameters. I told with Pierre about it before removing them

alekece:
> > Thanks for that work ! Can you explain the rational behind the deletion of `iterate` method ?
> 
> They were not referenced anywhere in the libcore (actually, once, in a test) and were difficult to add to the djinni API as they used `std::function` parameters. I talked with Pierre about it before removing them

Perfect. I suggest to indicate the rational - in the PR description - for a change to avoid that case

PR #603, opened 2020-06-21
Fix synchronization for RBF transaction

pollastri-pierre:
- Fix synchronization to replace transaction in user accounts when they are replaced in the mempool
- The code is not optimal, there is multiple thing to rework regarding the transaction and how the library deals with mempool transaction

This PR brings the following changes to the synchronization:
- Mempool is not handled by doing a diff between the mempool stored in DB and the transaction fetched from the network. Now it removes all mempool operations at the beginning of the synchronization (DB mempool is kept in memory in case of failures during synchronization, to be able to restore it).
- It accumulates all mempool transaction fetched on network in memory
- Filter transactions once (remove replaced transaction) and get all input addresses from foreign addresses (addresses which are not part of the account)
- Fetch foreign addresses transactions on the mempool
- Filter once more with replaceable transactions with foreign addresses transactions
- Insert transactions in DB

There is a lot of "pirouettes", due to the common implementation between ETH, XRP, XTZ and BTC. Also DB schema forces to do a lot of queries to be able to properly delete transactions in DB. 

Current Qt host has bugs regarding threading, this PR brings another ThreadDispatcher using libuv which is stable for now but there is no thread pool implementation (it returns sequential execution context).

Last thing this PR change is the use of TTL cache, BTC balance is computed at the end of the synchronization and stored in database. TTL cache is hard to manage when multiple instances of libcore on the same DB.

### DIFF
--- a/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
+++ b/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
@@ -719,5 +719,9 @@ namespace ledger {
         std::shared_ptr<api::Keychain> BitcoinLikeAccount::getAccountKeychain() {
           return _keychain;
         }
+
+        void BitcoinLikeAccount::dropTransactions(const std::vector<std::string> &hashes) {
+            // TODO implement SQL + publish
+        }
     }
 }
--- a/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
+++ b/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
@@ -308,6 +308,7 @@ namespace ledger {
                 if (result.isSuccess()) {
                     code = !isEmpty && wasEmpty ? api::EventCode::SYNCHRONIZATION_SUCCEED_ON_PREVIOUSLY_EMPTY_ACCOUNT
                                                 : api::EventCode::SYNCHRONIZATION_SUCCEED;
+                    self->getWallet()->invalidateBalanceCache(self->getIndex());
                 } else {
                     code = api::EventCode::SYNCHRONIZATION_FAILED;
                     payload->putString(api::Account::EV_SYNC_ERROR_CODE, api::to_string(result.getFailure().getErrorCode()));
@@ -718,10 +719,6 @@ namespace ledger {
 
         std::shared_ptr<api::Keychain> BitcoinLikeAccount::getAccountKeychain() {
           return _keychain;
-        }
-
-        void BitcoinLikeAccount::dropTransactions(const std::vector<std::string> &hashes) {
-            // TODO implement SQL + publish
         }
     }
 }
--- a/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
+++ b/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
@@ -256,7 +256,43 @@ namespace ledger {
         void
         BitcoinLikeTransactionDatabaseHelper::getMempoolTransactions(soci::session &sql, const std::string &accountUid,
                                                                      std::vector<BitcoinLikeBlockchainExplorerTransaction> &out) {
-            
+            // Query all transaction
+            rowset<row> txRows = (sql.prepare <<
+                    "SELECT  tx.hash, tx.version, tx.time, tx.locktime, "
+                    "block.hash, block.height, block.time, block.currency_name "
+                    "FROM bitcoin_transactions AS tx "
+                    "LEFT JOIN blocks AS block ON tx.block_uid = block.uid "
+                    "WHERE tx.hash IN ("
+                      "SELECT tx.hash FROM operations AS op "
+                      "JOIN bitcoin_operations AS bop ON bop.uid = op.uid  "
+                      "JOIN bitcoin_transactions AS tx ON tx.hash = bop.transaction_hash "
+                      "WHERE op.account_uid = :uid AND op.block_uid IS NULL"
+                    ")", use(accountUid));
+            // Inflate all transactions
+            for (const auto& txRow : txRows) {
+                BitcoinLikeBlockchainExplorerTransaction tx;
+                inflateTransaction(sql, txRow, accountUid, tx);
+                out.emplace_back(std::move(tx));
+            }
+        }
+
+        void BitcoinLikeTransactionDatabaseHelper::removeAllMempoolOperation(soci::session &sql,
+                                                                             const std::string &accountUid) {
+            rowset<std::string> rows = (sql.prepare <<
+                    "SELECT transaction_hash FROM bitcoin_operations AS bop "
+                    "JOIN operations AS op ON bop.uid = op.uid "
+                    "WHERE op.account_uid = :uid AND op.block_uid IS NULL", use(accountUid)
+            );
+            std::vector<std::string> txToDelete(rows.begin(), rows.end());
+            if (!txToDelete.empty()) {
+                sql << "DELETE FROM bitcoin_inputs WHERE uid IN ("
+                       "SELECT input_uid FROM bitcoin_transaction_inputs "
+                       "WHERE transaction_hash IN(:hashes)"
+                       ")", use(txToDelete);
+                sql << "DELETE FROM operations WHERE account_uid = :uid AND block_uid is NULL", use(accountUid);
+                sql << "DELETE FROM bitcoin_transactions "
+                       "WHERE hash IN (:txs)", use(txToDelete);
+            }
         }
 
     }
--- a/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
+++ b/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
@@ -253,5 +253,11 @@ namespace ledger {
             return true;
         }
 
+        void
+        BitcoinLikeTransactionDatabaseHelper::getMempoolTransactions(soci::session &sql, const std::string &accountUid,
+                                                                     std::vector<BitcoinLikeBlockchainExplorerTransaction> &out) {
+            
+        }
+
     }
 }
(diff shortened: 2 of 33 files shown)


alekece:
> Last thing this PR change is the use of TTL cache, BTC balance is computed at the end of the synchronization and stored in database. TTL cache is hard to manage when multiple instances of libcore on the same DB.

Our TTL cache implementation is actually to naive to manage such a case. As discuss internally a tool like memcached or Redis will be more adapted.

core/src/database/migrations.cpp
```diff
@@ -955,5 +955,14 @@ namespace ledger {
             sql << "DROP TABLE cosmos_currencies";
 
         }
+
+        template <> void migrate<21>(soci::session& sql, api::DatabaseBackendType type) {
+            sql << "ALTER TABLE bitcoin_accounts ADD balance VARCHAR(255) DEFAULT '0'";
```
phaazon, line 960: Hm. That makes the core library even more stateful as possible. I’m strongly against that.

pollastri-pierre: What do you mean by more stateful ? 
The current implementation is very stateful, it computes the balance and put it in a memory cache. Therefore with multiple instances on the same DB you end up with different results because the library keeps the state. The change makes the state handled by the database instead of the core library. Now it just read the value in database and this value is only written by the synchronization algorithm.

phaazon: Also, this denormalizes the DB.

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
```diff
@@ -95,5 +135,191 @@ namespace ledger {
...
+                    }
+                }
+                // Mark output
+                for (auto& output : transaction.outputs) {
+                    if (output.address.hasValue()) {
+                        keychain->markAsUsed(output.address.getValue());
+                        flag |= BitcoinLikeAccount::FLAG_TRANSACTION_CREATED_SENDING_OPERATION;
+                    }
+                }
+                return flag;
```
alekece, line 162: We can avoid redundant code by doing something like :

```suggestion    
    auto const markAddress = [keychain](auto &txIo) {
        auto flag = 0;
       
        for (auto &io : txIo) {
            if (io.address.hasValue()) {
                keychain->markAsUsed(io.address.getValue());
                flag |= BitcoinLikeAccount::FLAG_TRANSACTION_CREATED_SENDING_OPERATION;
            }
        }

        return flag;
    };

    return markAddress(transaction.inputs) | markAddress(transactoin.outputs);
```

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
```diff
@@ -95,5 +135,191 @@ namespace ledger {
...
+                    notReplaceableTxs.emplace_back(txWithHash.second.first);
+                }
+            }
+            // Use the list of addresses to get all mempool transaction for those addresses
+            return getMempoolTransactions(buddy,
+                    std::vector<std::string>(foreignAddresses.begin(), foreignAddresses.end()))
+            .map<Unit>(buddy->account->getContext(), [=] (const auto& foreignTxs) {
+                // Do another round of filtering on replaceable transaction vs foreign transactions
+                TransactionMap filteredTxs;
+                std::list<Transaction> txs(replaceableTxs.begin(), replaceableTxs.end());
```
alekece, line 278: Why use `std::list` ?

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
```diff
@@ -95,5 +135,191 @@ namespace ledger {
...
+            auto bitcoinBuddy = std::static_pointer_cast<BitcoinSynchronizationBuddy>(buddy);
+            return resolveMempool(bitcoinBuddy);
+        }
+
+        Future<Unit> BlockchainExplorerAccountSynchronizer::recoverFromFailedSynchronization(
+                const std::shared_ptr<CommonBuddy> &commonBuddy) {
+            auto buddy = std::static_pointer_cast<BitcoinSynchronizationBuddy>(commonBuddy);
+            return Future<Unit>::async(buddy->account->getContext(), [buddy] () {
+                soci::session sql(buddy->account->getWallet()->getDatabase()->getPool());
+                for (const auto& tx : buddy->previousMempool) {
```
teams2ua, line 308: Why do you want to insert transactions from mempool? We should drop them each time we want to do synchronizetion, no?

core/src/wallet/bitcoin/database/BitcoinLikeAccountDatabaseHelper.cpp
```diff
@@ -53,5 +53,18 @@ namespace ledger {
...
+            auto amount = balance.toString();
+            sql << "UPDATE bitcoin_accounts SET balance = :balance WHERE uid = :uid", use(amount), use(accountUid);
+        }
+
+        void BitcoinLikeAccountDatabaseHelper::queryAccountBalance(soci::session &sql, const std::string &accountUid,
+                                                                   BigInt &out) {
+            std::string amount;
+            sql << "SELECT balance FROM bitcoin_accounts WHERE uid = :uid", use(accountUid), into(amount);
+            out = BigInt::fromString(amount);
+        }
```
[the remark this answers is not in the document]
pollastri-pierre: The first will do nothing, the second will throw.

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
pollastri-pierre: no valuable reason, I'll change it to vector.

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
```diff
@@ -95,5 +135,191 @@ namespace ledger {
...
+
+        inline bool isInputReplaced(const Input& input, uint32_t conflictingInputSequence) {
+            return input.sequence < conflictingInputSequence;
+        }
+
+        inline bool isInputReplaceable(const Input& input) {
+            return input.sequence < std::numeric_limits<uint32_t>::max();
+        }
+
+        template <template<class, class> class List, typename Alloc>
```
teams2ua, line 180: Can we just have a normal std::vector in the parameters (and on all the places where we use std::list)?

core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -719,5 +730,9 @@ namespace ledger {
         std::shared_ptr<api::Keychain> BitcoinLikeAccount::getAccountKeychain() {
           return _keychain;
         }
+
+        void BitcoinLikeAccount::dropTransactions(const std::vector<std::string> &hashes) {
+            // TODO implement SQL + publish
```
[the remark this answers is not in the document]
pollastri-pierre: I'll remove it was part of a previous attempt.

core/src/database/migrations.cpp
teams2ua: @pollastri-pierre  How does this change relates to RBF?

pollastri-pierre: I'll remove this part, it's another fix which does not concern RBF. 
However I think we'll have to move all integration to this kind of logic:
 - Synchronization is the only process which can write data
 - Getter should be as simple as possible and return result based on state holder (i.e. the DB)
Balances should be moved to the synchronization part because there are heavily hit by clients and require a cache, also there are part of the state of the account which is updated by the synchronization. Only remaining call which can stay lazy are the one related to transaction creation (basically fee computation and sequence/nonce)

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
pollastri-pierre: yes, I'll replace to std::vector

core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
```diff
@@ -253,5 +253,57 @@ namespace ledger {
             return true;
         }
 
+        void
+        BitcoinLikeTransactionDatabaseHelper::getMempoolTransactions(soci::session &sql, const std::string &accountUid,
+                                                                     std::vector<BitcoinLikeBlockchainExplorerTransaction> &out) {
+            // Get all transaction hashes with
+            std::vector<std::string> transactionHashes;
```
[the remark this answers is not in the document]
pollastri-pierre: Indeed, I'll merge them, no need for this.

core/src/database/migrations.cpp
teams2ua: +1 for moving it to another PR. About heavy operations - I am ok to have a cache. And maybe in this case it is also OK to populate at during write (synchonization) 🤔

core/src/wallet/bitcoin/synchronizers/BlockchainExplorerAccountSynchronizer.cpp
pollastri-pierre: For the specific case of recoverFromFailedSynchronization:
I don't want a fail transaction to change the database  (it can fail because internet is down for example). But since I'm deleting all mempool transaction at the beginning of the synchronization, I restore the database to its previous state (artificially restore the database to its previous state in fact).

recoverFromFailedSynchronization should be only call if the synchronization process failed.

core/src/database/migrations.cpp
pollastri-pierre: I'll still need to make a minor change to the balance cache system, because it's failing tests. The cache is not invalidated after a synchro.

PR #605, opened 2020-06-24
Fix SQL query to remove operation in mempool for a given account

pollastri-pierre:
The SQL queries to remove mempool operation were wrong.
Since they were based on transaction hash it actually partially removes transactions even for other accounts.

The fix uses transaction_uid to drop only transaction bound to the account and avoid side effects on other accounts.

### DIFF
--- a/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
+++ b/core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
@@ -279,7 +279,7 @@ namespace ledger {
         void BitcoinLikeTransactionDatabaseHelper::removeAllMempoolOperation(soci::session &sql,
                                                                              const std::string &accountUid) {
             rowset<std::string> rows = (sql.prepare <<
-                    "SELECT transaction_hash FROM bitcoin_operations AS bop "
+                    "SELECT transaction_uid FROM bitcoin_operations AS bop "
                     "JOIN operations AS op ON bop.uid = op.uid "
                     "WHERE op.account_uid = :uid AND op.block_uid IS NULL", use(accountUid)
             );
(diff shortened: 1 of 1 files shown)


core/src/wallet/bitcoin/database/BitcoinLikeTransactionDatabaseHelper.cpp
```diff
@@ -279,19 +279,19 @@ namespace ledger {
         void BitcoinLikeTransactionDatabaseHelper::removeAllMempoolOperation(soci::session &sql,
                                                                              const std::string &accountUid) {
             rowset<std::string> rows = (sql.prepare <<
-                    "SELECT transaction_hash FROM bitcoin_operations AS bop "
+                    "SELECT transaction_uid FROM bitcoin_operations AS bop "
```
teams2ua, line 282: BTW this queries can be united to one.

pollastri-pierre: What do you mean ?
The result of this query is used in two different queries, I kept them in one query to avoid doing this query multiple time.
Also I need to remove operations then bitcoin_transactions due to foreign key constraints. That why I get uids before doing so.

teams2ua: I meant something like this
```sql << "DELETE FROM bitcoin_inputs WHERE uid IN ("
       "SELECT input_uid FROM bitcoin_transaction_inputs "
       "WHERE transaction_uid IN ("
       "SELECT transaction_uid FROM bitcoin_operations AS bop "
       "JOIN operations AS op ON bop.uid = op.uid "
       "WHERE op.account_uid = :uid AND op.block_uid IS NULL", use(accountUid))"
       ")", use(;
sql << "DELETE FROM operations WHERE account_uid = :uid AND block_uid is NULL", use(accountUid);
sql << "DELETE FROM bitcoin_transactions "
       "WHERE transaction_uid IN ("
       "SELECT transaction_uid FROM bitcoin_operations AS bop "
       "JOIN operations AS op ON bop.uid = op.uid "
       "WHERE op.account_uid = :uid AND op.block_uid IS NULL", use(accountUid))"```
It will execute the same subquery twice, I agree, but IMO it is better then sending data between DB and client. But up to you.

PR #606, opened 2020-06-25
Add a method in BTC outputs to know if the output could be replaced (RBF)

pollastri-pierre:
This PR adds a new function to BitcoinLikeOutput to check if the output may be replaced or not.

```
   # Check if the transaction (which created this output) is replaceable (RBF).
    # An output can be replaceable if the transaction has at least one RBF input
    # and if the transaction is not a block.
    # @return true if the output is replaceable, false otherwise
    isReplaceable(): bool;
```


teams2ua:
Why this PR contains commits from previous, already merged?

pollastri-pierre:
> Why this PR contains commits from previous, already merged?

I didn't rebase on develop.

core/src/database/migrations.cpp
```diff
@@ -955,5 +955,17 @@ namespace ledger {
...
         }
+
+        template <> void migrate<21>(soci::session& sql, api::DatabaseBackendType type) {
+           sql << "ALTER TABLE bitcoin_outputs ADD replaceable INTEGER DEFAULT 0";
+        }
+
+        template <> void rollback<21>(soci::session& sql, api::DatabaseBackendType type) {
+            // SQLite doesn't handle ALTER TABLE DROP
+            if (type != api::DatabaseBackendType::SQLITE3) {
+                sql << "ALTER TABLE bitcoin_outputs DROP replaceable";
```
alekece, line 966: Can you use the trick we did before to alter the table even for `sqlite` ?
When we get completely rid of `sqlite` support we can take such a shortcut

core/idl/wallet/bitcoin/bitcoin_like_wallet.djinni
```diff
@@ -97,6 +97,11 @@ BitcoinLikeOutput = interface +c {
     getAddress(): optional<string>;
     getDerivationPath(): optional<DerivationPath>;
     getBlockHeight(): optional<i64>;
+    # Check if the transaction (which created this output) is replaceable (RBF).
+    # An output can be replaceable if the transaction has at least one RBF input
+    # and if the transaction is not a block.
+    # @return true if the output is replaceable, false otherwise
+    isReplaceable(): bool;
```
alekece, line 104:
```suggestion
    const isReplaceable(): bool;
```

PR #608, opened 2020-06-26
Fix rollback script for migration 21

pollastri-pierre:
This PR fixes a typo in the SQL query for doing a rollback of database schema version 21

### DIFF
--- a/core/src/database/migrations.cpp
+++ b/core/src/database/migrations.cpp
@@ -976,7 +976,7 @@ namespace ledger {
                        "PRIMARY KEY (idx, transaction_uid)"
                        ")";
                 sql << "INSERT INTO bitcoin_outputs_swap "
-                       "SELECT idx, transaction_uid, transaction_hash, amount, scripts, "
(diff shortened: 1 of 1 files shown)


gre:
can we merge?

core/src/database/migrations.cpp
```diff
@@ -976,7 +976,7 @@ namespace ledger {
                        "PRIMARY KEY (idx, transaction_uid)"
                        ")";
                 sql << "INSERT INTO bitcoin_outputs_swap "
-                       "SELECT idx, transaction_uid, transaction_hash, amount, scripts, "
+                       "SELECT idx, transaction_uid, transaction_hash, amount, script, "
```
phaazon, line 979: Was that migration released yet? If so, we shouldn’t update migrations like that, as it will go full Kitchen Nightmare™ on production (since they have already applied the migration).

gre:
but that migration never worked right? there were a typo 🤔 

BTW linux is so slow than it reached timeout :( 

```
Build timed out after 5h0m0s
```

PR #612, opened 2020-07-03
Fix signed int vs unsigned int comparison

teams2ua:
Make types more obvious to avoid the problem
https://stackoverflow.com/questions/5416414/signed-unsigned-comparisons


alekece:
This PR has been move to PR #598

PR #618, opened 2020-07-07
[cosmos] Remove gas ratio multiplier in fees

gagbo:
Apparently all advertised/signed fees are consumed in transaction and no
ratio of actual gas usage is used regarding fee computation


alekece (approved):
I always enjoy when we remove a bunch of line of codes

PR #619, opened 2020-07-07
[Tezos] Fix bug when sending multiple transactions in a row

llehmann-ledger:
Save tx counter and use it when explorer_counter < saved_counter, should solve the bug in which only the first tx will be accepted by the networks because the explorer didn't have the time to increment it

### DIFF
--- a/core/src/wallet/tezos/TezosLikeAccount2.cpp
+++ b/core/src/wallet/tezos/TezosLikeAccount2.cpp
@@ -367,8 +367,18 @@ namespace ledger {
                                                 if (!counter) {
                                                     throw make_exception(api::ErrorCode::RUNTIME_ERROR, "Failed to retrieve counter from network.");
                                                 }
-                                                // We should increment current counter
-                                                tx->setCounter(std::make_shared<BigInt>(++(*counter)));
+                                                auto explorer_counter = counter->toInt64();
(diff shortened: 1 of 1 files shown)


core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -367,8 +367,18 @@ namespace ledger {
                                                 if (!counter) {
                                                     throw make_exception(api::ErrorCode::RUNTIME_ERROR, "Failed to retrieve counter from network.");
                                                 }
-                                                // We should increment current counter
-                                                tx->setCounter(std::make_shared<BigInt>(++(*counter)));
+                                                auto explorer_counter = counter->toInt64();
+                                                auto saved_counter = self->getInternalPreferences()->getLong("counter", 0);
```
gagbo, line 371: Not sure we should use internalPreferences database for this. Should we store that, probably an additional column in tezosAccounts is better ?

llehmann-ledger: Sure that would be a viable alternative, I first begun by modifying migrate/rollback in migrations but I finally used  internalPreferences because I liked this solution better, I guess it's more a matter of personal preference in the end ?

PR #620, opened 2020-07-07
Make preferences API retro-compatible

pollastri-pierre:
New PreferencesBackend update broke the API for WalletPool::newInstance. I reverted the API to keep the same API.
For custom preferences API, users will have to use WalletPoolBuilder

### DIFF
--- a/core/src/wallet/pool/WalletPool.cpp
+++ b/core/src/wallet/pool/WalletPool.cpp
@@ -146,7 +146,7 @@ namespace ledger {
                     name, password, httpClient, webSocketClient, pathResolver,
                     logPrinter, dispatcher, rng, backend, configuration,
                     externalPreferencesBackend,
-                    externalPreferencesBackend
+                    internalPreferencesBackend
             ));
 
             // Initialization
--- a/core/src/wallet/pool/WalletPool.cpp
+++ b/core/src/wallet/pool/WalletPool.cpp
@@ -141,12 +141,12 @@ namespace ledger {
             const std::shared_ptr<api::DatabaseBackend> &backend,
             const std::shared_ptr<api::DynamicObject> &configuration,
             const std::shared_ptr<api::PreferencesBackend> &externalPreferencesBackend,
-            const std::shared_ptr<api::PreferencesBackend> &internalPreferencesBackend
-        ) {
+            const std::shared_ptr<api::PreferencesBackend> &internalPreferencesBackend) {
             auto pool = std::shared_ptr<WalletPool>(new WalletPool(
                     name, password, httpClient, webSocketClient, pathResolver,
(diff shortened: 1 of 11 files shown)


core/src/wallet/pool/WalletPool.cpp
```diff
@@ -141,12 +141,12 @@ namespace ledger {
...
             const std::shared_ptr<api::PreferencesBackend> &externalPreferencesBackend,
-            const std::shared_ptr<api::PreferencesBackend> &internalPreferencesBackend
-        ) {
+            const std::shared_ptr<api::PreferencesBackend> &internalPreferencesBackend) {
             auto pool = std::shared_ptr<WalletPool>(new WalletPool(
                     name, password, httpClient, webSocketClient, pathResolver,
                     logPrinter, dispatcher, rng, backend, configuration,
-                    externalPreferencesBackend, internalPreferencesBackend
+                    externalPreferencesBackend,
+                    externalPreferencesBackend
```
alekece, line 149: The `internalPreferencesBackend` isn't use anymore. If this change is done in purpose, can you remove the variable name to the method parameter list or put a comment to indicate - in both case - why such a variable is unused.

core/src/wallet/pool/WalletPool.hpp
```diff
@@ -124,6 +124,8 @@ namespace ledger {
                 const std::shared_ptr<api::PreferencesBackend> &internalPreferencesBackend
             );
 
+
+
```
alekece, line 128: We should avoid to introduce formatting changes since a formatting tool will be integrated soon

core/src/wallet/pool/WalletPool.cpp
pollastri-pierre: 😨 good catch, internal is actually the important one.

core/src/wallet/pool/api/WalletPoolApi.cpp
```diff
@@ -52,10 +53,18 @@ namespace ledger {
             const std::shared_ptr<api::ThreadDispatcher> &dispatcher,
             const std::shared_ptr<api::RandomNumberGenerator> &rng,
             const std::shared_ptr<api::DatabaseBackend> &backend,
-            const std::shared_ptr<api::DynamicObject> &configuration,
-            const std::shared_ptr<api::PreferencesBackend> &externalPreferencesBackend,
-            const std::shared_ptr<api::PreferencesBackend> &internalPreferencesBackend
+            const std::shared_ptr<api::DynamicObject> &configuration
         ) {
+            auto externalPreferencesBackend = std::make_shared<ledger::core::PreferencesBackend>(
```
pollastri-pierre, line 58: This part is not necessary there is already a check in WalletPool constructor.

PR #623, opened 2020-07-10
Create Keychain API

hlafet-ledger:
new keychain api

### DIFF
--- a/ledger-core-bitcoin/idl/addresses.djinni
+++ b/ledger-core-bitcoin/idl/addresses.djinni
@@ -80,12 +80,4 @@ BitcoinLikeExtendedPublicKey = interface +c {
     getRootPath(): string;
 }
 
-# A keychain can be seen as an address database that can be queried to get lists of addresses,
-# ask whether an address is contained, or generate new addresses.
-Keychain = interface +c {
-    # Check whether an address is contained.
-    const contains(address: string): bool;
-	# get all contained adresses.
-    getAllAddresses(): list<Address>;
-}
 
--- a/ledger-core-bitcoin/idl/addresses.djinni
+++ b/ledger-core-bitcoin/idl/addresses.djinni
@@ -79,5 +79,3 @@ BitcoinLikeExtendedPublicKey = interface +c {
     # Get the root path of the xPUB.
     getRootPath(): string;
 }
-
-
--- a/ledger-core-bitcoin/idl/addresses.djinni
+++ b/ledger-core-bitcoin/idl/addresses.djinni
@@ -79,3 +79,13 @@ BitcoinLikeExtendedPublicKey = interface +c {
     # Get the root path of the xPUB.
     getRootPath(): string;
 }
+
+# A keychain can be seen as an address database that can be queried to get lists of addresses,
+# ask whether an address is contained, or generate new addresses.
+Keychain = interface +c {
+    # Check whether an address is contained.
+    const contains(address: string): bool;
+	# get all contained adresses.
+    getAllAddresses(): list<Address>;
+}
+
--- a/ledger-core-bitcoin/idl/idl.djinni
+++ b/ledger-core-bitcoin/idl/idl.djinni
@@ -334,6 +334,8 @@ BitcoinLikeAccount = interface +c {
     # Get addresses given a range of indices
     # Note: this will return public and change addresses
     getAddresses(from: i64, to: i64, callback: ListCallback<Address>);
+    # get all contained adresses.
+    getAllAddresses(): list<Address>;
 }
 
 # A bitcoin-like wallet.
--- a/ledger-core-bitcoin/src/bitcoin/BitcoinLikeAccount.cpp
+++ b/ledger-core-bitcoin/src/bitcoin/BitcoinLikeAccount.cpp
@@ -695,5 +695,10 @@
             void BitcoinLikeAccount::getAddresses(int64_t from, int64_t to, const std::shared_ptr<api::AddressListCallback> & callback) {
                 return getAddresses(from, to).callback(getMainExecutionContext(), callback);
             }
+
+            AbstractAccount::AddressList BitcoinLikeAccount::getAllAddresses() {
+                auto keychain = getKeychain();
+                return fromBitcoinAddressesToAddresses(keychain->getAllAddresses());
+            }
         }
     }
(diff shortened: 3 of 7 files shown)


ledger-core-bitcoin/idl/addresses.djinni
```diff
@@ -79,3 +79,13 @@ BitcoinLikeExtendedPublicKey = interface +c {
     # Get the root path of the xPUB.
     getRootPath(): string;
 }
+
+# A keychain can be seen as an address database that can be queried to get lists of addresses,
+# ask whether an address is contained, or generate new addresses.
+Keychain = interface +c {
```
phaazon, line 85: On the legacy, I have added that interface in the “common” part of the library, [here](https://github.com/LedgerHQ/lib-ledger-core/blob/develop/core/idl/wallet/common/wallet.djinni#L317-L323). So I think this should belong to the `ledger-core` project, not necessarily to `ledger-core-bitcoin`.

hlafet-ledger: It was the recommendation of Pierre to add it to the bitcoin interface only, may be because it does not make sense when we have only one adresse to return with the other currencies.

ledger-core-bitcoin/idl/idl.djinni
```diff
@@ -334,6 +334,8 @@ BitcoinLikeAccount = interface +c {
     # Get addresses given a range of indices
     # Note: this will return public and change addresses
     getAddresses(from: i64, to: i64, callback: ListCallback<Address>);
+    # get all contained adresses.
+    getAllAddresses(): list<Address>;
```
alekece, line 338: Please add `const` qualifier

ledger-core-bitcoin/src/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -695,5 +695,10 @@
             void BitcoinLikeAccount::getAddresses(int64_t from, int64_t to, const std::shared_ptr<api::AddressListCallback> & callback) {
                 return getAddresses(from, to).callback(getMainExecutionContext(), callback);
             }
+
+            AbstractAccount::AddressList BitcoinLikeAccount::getAllAddresses() {
+                auto keychain = getKeychain();
```
alekece, line 700: Please add `const` qualifier

ledger-core-bitcoin/src/bitcoin/keychains/CommonBitcoinLikeKeychains.cpp
```diff
@@ -227,6 +227,18 @@ namespace ledger {
...
+        std::vector<BitcoinLikeKeychain::Address> CommonBitcoinLikeKeychains::getAllAddresses() {
+            std::vector<BitcoinLikeKeychain::Address> result;
+            result.reserve(_state.maxConsecutiveChangeIndex + 1 + _state.maxConsecutiveReceiveIndex + 1);
+            for (auto i = 0; i <= _state.maxConsecutiveChangeIndex; i++) {
+                result.push_back(derive(KeyPurpose::CHANGE, i));
+            }
+            for (auto i = 0; i <= _state.maxConsecutiveReceiveIndex; i++) {
+                result.push_back(derive(KeyPurpose::RECEIVE, i));
+            }
+            return result;
```
alekece, line 239: You don't need the two extra slots when reserving memory. BTW I suggest you to remove redundant codes even in local function for readability, maintenance and to ease the cognitive complexity. I also suggest to use better name for variables - `result` name is too generic for such a case.

```suggestion
            std::vector<BitcoinLikeKeychain::Address> addresses;
            addresses.reserve(_state.maxConsecutiveChangeIndex + _state.maxConsecutiveReceiveIndex);
            
            auto fetchAddressesFrom = [&addresses](auto const keyPurpose, auto const maxIndex) {
                  for (auto i = 0; i <= maxIndex; ++i) {
                      addresses.push_back(derive(keyPurpose, i));
                  }
            };
            
            fetchAddressesFrom(KeyPurpose::CHANGE, _state.maxConsecutiveChangeIndex);
            fetchAddressesFrom(KeyPurpose::RECEIVE, _state.maxConsecutiveReceiveIndex);
            
            return addresses;
```

ledger-core-bitcoin/idl/addresses.djinni
```diff
@@ -79,3 +79,5 @@ BitcoinLikeExtendedPublicKey = interface +c {
     # Get the root path of the xPUB.
     getRootPath(): string;
 }
+
+
```
alekece, line 83: Please remove these new lines

ledger-core-bitcoin/src/bitcoin/keychains/CommonBitcoinLikeKeychains.cpp
hlafet-ledger: - for (auto i = 0; i <= maxIndex; ++i) => maxIndex+1 items
- Do we really need a delta expression just for a loop ? For me, it affects fluidity when reading the code

ledger-core-bitcoin/idl/idl.djinni
hlafet-ledger: I keept the same logic as getAddresses(from, to)

alekece: There is a lot of missing `const` qualifiers in our interface - for now we let it like that because a lot of changes are needed to add it.
However for new methods we must add that `const` qualifer.

gre:
great work. Would it be possible to have more detail in the description of examples of what this allows? I would like to make sure we are all on the same page (what does the contract looks like + an example on bitcoin?)

will this solve the possibility to do this? https://ledgerhq.atlassian.net/browse/LLC-171 

thanks!

ledger-core-bitcoin/idl/idl.djinni
hlafet-ledger: Adding the const qualifier will be more complicated than it seems: All nested functions must be changed to const too.

alekece: This is a great indicator than - maybe - the name of the function is not good because a getter should - in most of the case - never mutate any member of an instance so :
- Can you list which methods/functions have to be `const` to compile the library ?
- Then if these methods/functions are not `const` by decision so rename your getter to something more appropriate

hlafet-ledger: All functions reachable by a call to getAllAddresses must be changed to const (derive, ...)
Refactoring the code could be done in another commit, but it has to touch several similar methods. 
The name must be kept as we have some similar functions like getAddresses (from, to)

PR #627, opened 2020-07-17
LLC-692 | Fix the way we check whether a Tezos account is funded.

phaazon:
We now depend on the explorer instead of looking at the transactions.
Doing a transaction check is tricky, because of the fact failed
transactions are still recorded on the blockchain (someone has to pay
the fees), but they don’t credit the recipient account; so it still
stays unfunded.

### DIFF
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -304,5 +304,23 @@ namespace ledger {
                                                             getRPCNodeEndpoint());
         }
 
+        Future<bool> ExternalTezosLikeBlockchainExplorer::isFunded(const std::string &address) {
+            return
+                _http->GET(fmt::format("account/{}", address))
+                    .json().map<bool>(getExplorerContext(), [=](const HttpRequest::JsonResult &result) {
+                        auto& json = *std::get<1>(result);
+
+                        // look for the is_funded field
+                        //Is there a fees field ?
+                        auto field = "is_funded";
+                        if (!json.IsObject() || !json.HasMember(field) ||
+                            !json[field].IsBool()) {
+                            throw make_exception(api::ErrorCode::HTTP_ERROR,
+                                                 "Failed to get is_funded from network, no (or malformed) field \"result\" in response");
+                        }
+
+                        return json[field].GetBool();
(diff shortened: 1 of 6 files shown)


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -304,5 +304,23 @@ namespace ledger {
...
 
+        Future<bool> ExternalTezosLikeBlockchainExplorer::isFunded(const std::string &address) {
+            return
+                _http->GET(fmt::format("account/{}", address))
+                    .json().map<bool>(getExplorerContext(), [=](const HttpRequest::JsonResult &result) {
+                        auto& json = *std::get<1>(result);
+
+                        // look for the is_funded field
+                        //Is there a fees field ?
+                        auto field = "is_funded";
```
alekece, line 315: Add `const` here ;)

core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
```diff
@@ -289,5 +289,24 @@ namespace ledger {
...
+                        // look for the is_funded field
+                        //Is there a fees field ?
+                        auto field = "is_funded";
+                        if (!json.IsObject() || !json.HasMember(field) ||
+                            !json[field].IsBool()) {
+                            throw make_exception(api::ErrorCode::HTTP_ERROR,
+                                                 "Failed to get is_funded from network, no (or malformed) field \"result\" in response");
+                        }
+
+                        return json[field].GetBool();
```
alekece, line 308: Since that chunk of code seems to be the exact same than the one in `ExternalTezosLikeBlockchainExplorer`, can you put the code in a free function instead ?

phaazon: No, because even though it’s similar, it might change regarding what we do on an explorer or the other – and I’m not completely sure `is_funded` is present on our explorers. I wait for Chenyu for this.

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
phaazon: Sure.

PR #630, opened 2020-07-22
Fix segfault related to openssl

llehmann-ledger:
This commit fix a segfault on Linux occuring when the system openssl is different from the libcore embedded openssl. It prevents openssl from automatically export its symbols which was causing the segfault in the first place.


llehmann-ledger:
I don't have write access on the repository, so I cannot merge even if the change is approved

PR #631, opened 2020-07-22
[algorand] integration in libcore v2

barjonr:
Algorand integtration in libcore v2, up to date with the code from the PR to integrate in v1


pollastri-pierre:
Drop support of libcore v2

PR #638, opened 2020-08-12
LLC-696: Tezos regression: can no longer build a transaction to a new…

hlafet-ledger:
… account address

### DIFF
--- a/core/src/net/HttpClient.hpp
+++ b/core/src/net/HttpClient.hpp
@@ -89,7 +89,7 @@ namespace ledger {
                 });
             }
 
-            Future<JsonResult> json(bool parseNumbersAsString = false) const;
+            Future<JsonResult> json(bool parseNumbersAsString = false, bool ignoreStatusCode = false) const;
             std::shared_ptr<api::HttpRequest> toApiRequest() const;
 
 
--- a/core/src/net/HttpClient.cpp
+++ b/core/src/net/HttpClient.cpp
@@ -153,7 +153,7 @@ namespace ledger {
             });
         }
 
-        Future<HttpRequest::JsonResult> HttpRequest::json(bool parseNumbersAsString) const {
+        Future<HttpRequest::JsonResult> HttpRequest::json(bool parseNumbersAsString, bool ignoreStatusCode) const {
             return operator()().recover(_context, [] (const Exception& exception) {
                 if (HttpRequest::isHttpError(exception.getErrorCode()) &&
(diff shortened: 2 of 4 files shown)


core/src/net/HttpClient.hpp
```diff
@@ -89,7 +89,7 @@ namespace ledger {
                 });
             }
 
-            Future<JsonResult> json(bool parseNumbersAsString = false) const;
+            Future<JsonResult> json(bool parseNumbersAsString = false, bool ignoreStatusCode = false) const;
```
alekece, line 92: I am really not comfortable with integer/boolean arguments as parameter because the readability at the call site is really terrible and lead to mistake. I suggest you to encapsulate such a boolean like :
```
class WatchedValue {   
public:
    constexpr WatchedValue(bool v) : _watched(v) {};

    // Returns true if the value is watched otherwise false
    constexpr operator bool() const {
      return _watched;
    }
private:

    bool const _watched;
};
```

Then define some constants into `HttpClient` : 
```
constexpr auto IgnoredStatusCode = WatchedValue(false);
constexpr auto WatchedStatusCode = WatchedValue(true);

Future<JsonResult> json(bool parseNumbersAsString = false, WatchedValue statusCode = false) const;
```

And finally at the call site :
```
_http->GET(fmt::format("account/{}", address))
                    .json(false, IgnoredStatusCode).map<bool>(getExplorerContext(), [=](const HttpRequest::JsonResult  &result) {
...

Which is more easy to read and reason about. 
We can do the exact same thing for `parseNumberAsString` but this is a subject for another PR.

Comment (hlafet-ledger) [reply core/src/net/HttpClient.hpp:92]: I'm not comfortable with this proposal for many reasons:
1- We have to keep consistency with the other parameter (parseNumbersAsString): I am not OK to mix objects encapsulating a boolean value and native booleans
2- By adding a new object that just takes a boolean and then return it, we are complicating the work of readers that will working later on the project. we have to keep the code simple and clean and avoid adding useless classes
3- if you want to avoid confusion when calling the json function, a good solution would be to use an enum like this:
```
`enum StatusCodePolicy {
	IgnoreStatusCode,
	CheckdStatusCode
};` 
```
With an enum, its strongly typed, and the behaviour will be clear when calling the function

alekece: As discuss internally let's merge that PR and move that refactorization into another one when the team find a consensus about such a modification.

PR #643, opened 2020-08-20
[COIN 584] [XTZ] Improve Fees and GasLimit using a RPC node

gagbo:
- HttpClient : Send error payloads as well as the http code from Http exceptions (because it make debugging easier down the road)
- TezosLikeTransactionApi : Add a JSON serializer specific for the `run_operation` endpoint on a RPC node. The binary serialization did not work well with the endpoint and the error messages are harder to act on
- TezosLikeExplorers : add interface (implemented only on TZSTATS type of explorer) to fetch the current gasPrice
- TezosLikeExplorers : add interface (implemented only on TZSTATS type of explorer for now but can be improved) to fetch the gasLimit for a transaction through Node simulation. The code only works for tx that use Reveal and/or Transaction type operations for now
- TezosLikeAccount : give an interface to fetch the current gasPrice on the network (will throw on non tzstats explorers), and allow the transaction builder to dynamically compute the fees for a transaction.

To make the automatic estimation this is currently an admittedly bad hack to keep the current request API backwards compatible:
- the _signal_ is to put the *GasLimit* to `"0"` in the request
- the _fee price_ you want to pay should be in the *Fees* of the request, *in picoTez (10^-12 XTZ) per gasUnit*

Then on transaction building, Core detects the 2 info to actually make a transaction with the correct gasLimit and the associated fees according to your fee price. Note that there are no guarantees that the fees amount won't be too high for the sender's balance, callers should check this before signing/broadcasting.


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -83,21 +98,47 @@ namespace ledger {
...
+                        const auto totalFees = api::BigInt::fromDecimalString(json[feesField].GetString(), 6, ".");
+                        std::string fees = api::TezosConfigurationDefaults::TEZOS_DEFAULT_FEES;
+                        if (fees != "0" && totalTx->intValue() != 0) {
+                            fees = totalFees->divide(totalTx)->toString(10);
                         }
-                        // Since nodes are giving some awkward values, we set a threshold to avoid having really fees
-                        // Factor for threshold is inspired from other XTZ wallets
+                        // Since nodes are giving some awkward values, we set a threshold to avoid
+                        // having really fees Factor for threshold is inspired from other XTZ
+                        // wallets
```
haammar-ledger, line 117: Need a `.` between `fees` and `Factor`, otherwise it doesn't make any sense :grammarnazi:

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -83,21 +98,47 @@ namespace ledger {
...
+                        auto &json = *std::get<1>(result);
+
+                        if (!json.IsObject() || !json.HasMember(gasPriceField) ||
+                            !json[gasPriceField].IsString()) {
+                            throw make_exception(api::ErrorCode::HTTP_ERROR,
+                                                 fmt::format("Failed to get gas_price from network, no (or malformed) field \"{}\" in response", gasPriceField));
+                        }
+                        const std::string apiGasPrice = json[gasPriceField].GetString();
+                        const std::string picoTezGasPrice = api::BigInt::fromDecimalString(apiGasPrice, 6, ".")->toString(10);
+                        return std::make_shared<BigInt>(std::stoi(picoTezGasPrice));
```
haammar-ledger, line 138: Why not just return the result of `api::BigInt::fromDecimalString` here?

core/src/wallet/tezos/explorers/NodeTezosLikeBlockchainExplorer.cpp
```diff
@@ -250,7 +256,84 @@ namespace ledger {
             );
         }
 
-        Future<std::shared_ptr<BigInt>> NodeTezosLikeBlockchainExplorer::getStorage(const std::string &address) {
+        Future<std::shared_ptr<BigInt>>
+        NodeTezosLikeBlockchainExplorer::getEstimatedGasLimit(
```
haammar-ledger, line 260: If I'm not mistaken, the code of `getEstimatedGasLimit` is exactly the same for `ExternalTezosLikeBlockchainExplorer` and `NodeTezosLikeBlockchainExplorer`, so maybe it could belong to `TezosLikeBlockchainExplorer` to avoid duplication?

haammar-ledger:
(Only the comment on `TezosLikeTransactionApi.cpp` needs double checking IMO, the rest is probably unimportant).

core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -369,15 +369,34 @@ namespace ledger {
...
+                                                                self->getGasPrice()
+                                                                :
+                                                                FuturePtr<BigInt>::successful(request.fees);
+
+                                                            return gasPrice_fut.flatMapPtr<api::TezosLikeTransaction>(self->getMainExecutionContext(), [self, filledTx] (const std::shared_ptr<BigInt>&gasPrice) -> FuturePtr<api::TezosLikeTransaction> {
+                                                                return self->estimateGasLimit(filledTx).flatMapPtr<api::TezosLikeTransaction>(self->getMainExecutionContext(), [filledTx, gasPrice] (const std::shared_ptr<BigInt> &gas) -> FuturePtr<api::TezosLikeTransaction> {
+                                                                    // 0.000001 comes from the gasPrice being in picoTez
+                                                                    const auto fees = std::make_shared<BigInt>(static_cast<int64_t>(1 + gas->toInt64() * static_cast<double>(gasPrice->toInt64()) * 0.000001));
```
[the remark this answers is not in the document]
gagbo: +1 is to get the `ceil` of the value instead of the `floor`, and I need to cast the gasPrice to double to make sure that multiplication by `0.000001` is not truncated to 0. I'll cast everything inside

core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -422,6 +441,30 @@ namespace ledger {
...
+
+        FuturePtr<BigInt> TezosLikeAccount::getGasPrice() {
+            return _explorer->getGasPrice();
+        }
+
+        FuturePtr<BigInt> TezosLikeAccount::estimateGasLimit(const std::shared_ptr<TezosLikeTransactionApi>& tx, double adjustment_factor) {
+            return _explorer->getEstimatedGasLimit(tx).flatMapPtr<BigInt>(
+                getMainExecutionContext(),
+                [adjustment_factor](const std::shared_ptr<BigInt>& consumedGas){
+                    auto adjusted_gas = static_cast<int64_t>(1 + consumedGas->toInt64() * adjustment_factor);
```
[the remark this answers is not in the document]
gagbo: +1 to ceil again

core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
```diff
@@ -328,6 +331,185 @@ namespace ledger {
...
+
+                // Fake sign
+                static const auto bogusSignature =
+                    "edsigtkpiSSschcaCt9pUVrpNPf7TTcgvgDEDD6NCEHMy8NNQJCGnMfLZzYoQj74yLjo9wx6MPVV29"
+                    "CvVzgi7qEcEUok3k7AuMg";
+                vString.SetString(
+                    bogusSignature,
+                    static_cast<SizeType>(std::strlen(bogusSignature)),
+                    allocator);
+                opObject.AddMember("signature", vString, allocator);
```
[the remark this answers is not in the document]
gagbo: the node refuses to make the operation simulation with an error

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
gagbo: because when I use BigInt->toInt64 later I don't know if I get `500` or `50` for a value like `"0.50"`

core/src/wallet/tezos/TezosLikeAccount2.cpp
```diff
@@ -369,15 +369,34 @@ namespace ledger {
...
+                                                        if (request.gasLimit->toInt() == 0) {
+                                                            auto filledTx = tx;
+                                                            auto gasPrice_fut = request.fees->toInt() == 0 ?
+                                                                self->getGasPrice()
+                                                                :
+                                                                FuturePtr<BigInt>::successful(request.fees);
+
+                                                            return gasPrice_fut.flatMapPtr<api::TezosLikeTransaction>(self->getMainExecutionContext(), [self, filledTx] (const std::shared_ptr<BigInt>&gasPrice) -> FuturePtr<api::TezosLikeTransaction> {
+                                                                return self->estimateGasLimit(filledTx).flatMapPtr<api::TezosLikeTransaction>(self->getMainExecutionContext(), [filledTx, gasPrice] (const std::shared_ptr<BigInt> &gas) -> FuturePtr<api::TezosLikeTransaction> {
+                                                                    // 0.000001 comes from the gasPrice being in picoTez
```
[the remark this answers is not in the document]
gagbo: `gasPrice->toInt64` is the value that's guaranteed to be in picoTez, I'll clarify

core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.h
```diff
@@ -127,8 +127,13 @@ namespace ledger {
             virtual Future<std::shared_ptr<BigInt>>
             getEstimatedGasLimit(const std::string &address) = 0;
 
-            virtual Future<std::shared_ptr<BigInt>>
-            getEstimatedGasLimit(const std::shared_ptr<TezosLikeTransactionApi> &transaction) = 0;
+            virtual Future<std::shared_ptr<BigInt>> getEstimatedGasLimit(
+                const std::shared_ptr<TezosLikeTransactionApi> &tx) = 0;
+
+            Future<std::shared_ptr<BigInt>> getEstimatedGasLimit(
+                const std::shared_ptr<HttpClient> &_http,
```
haammar-ledger, line 134: Should be `http` not `_http`

haammar-ledger (approved):
Seems all good, for a non-Tezos-expert though.
Thanks for the answers to my questions.

haammar-ledger (approved):

gagbo:
> HODL it blocks reveal for now

All issues are fixed and accounted for, but for the time being it's a pile of hacks in `TezosLikeExtendedPublicKey::toBase58` and this needs to be fixed properly and defined in unit tests

gagbo:
TODO: investigate on which branches this code is live

PR #644, opened 2020-08-21
Protect json reads in Algorand Explorer to prevent crashes

haammar-ledger:
Algorand Explorer causes crashes when there is no network connectivity --> sanitize JSON in case of no data received.


alekece (approved):
LGTM

PR #646, opened 2020-08-24
Remove optional numbers from Djinni records

haammar-ledger:
iOS doesn't tolerate `optional` numbers in method parameters (see screenshot below)...

[A first PR](https://github.com/LedgerHQ/lib-ledger-core/pull/642) removed all instance of `optional` parameters in functions explicitely defined in djinni, however it missed generated functions, such as records constructors... eg. the one of `CosmosGasLimitRequest` which causes the issue below with `CoreLGCosmosGasLimitRequest.init`

This PR changes all records fields that are optional numbers to optional strings, which are accepted by iOS.

image


core/idl/wallet/algorand/idl.djinni
```diff
@@ -35,12 +35,12 @@ AlgorandAssetParams = record {
...
     url: optional<string>;
     # Specifies whether holdings in this asset are frozen by default
-    defaultFrozen: optional<bool>;
+    defaultFrozen: bool;
 
     # The total number of units of this asset
     total: optional<string>;
     # The number of digits to use after the decimal point when displaying this asset
-    decimals: optional<i32>;
+    decimals: optional<string>;
```
alekece, line 43: Instead of use a string which brings conversation mechanism and probably new unmanaged errors, I suggest to key the initial type which is `i32` and create placeholder value.
This placeholder can be used to skip the value if wanted

core/idl/wallet/cosmos/wallet.djinni
```diff
@@ -120,7 +120,7 @@ CosmosGasLimitRequest = record {
     # The transaction is more expensive usually because we send a smaller
     # payload than usual to the simulation backend (and gas is also consumed
     # per tx byte)
-    amplifier: optional<f64>;
+    amplifier: optional<string>;
```
alekece, line 123: Ditto.

core/idl/wallet/ethereum/ethereum_like_wallet.djinni
```diff
@@ -145,7 +145,7 @@ EthereumGasLimitRequest = record {
     # Current gas price
     gasPrice: optional<string>;
     # Amplifier coefficient for safety
-    amplifier: optional<f64>;
+    amplifier: optional<string>;
```
alekece, line 148: Ditto.

alekece (changes requested):
Nice catch ! I am quite surprised this error has detected just right now.

gagbo (approved):
Shouldn't impact Cosmos as far as I understand

core/idl/wallet/algorand/idl.djinni
barjonr: We used `string` everywhere for numbers in algorand (because we couldn't use u64), so I think it wouldn't make a lot of sense to use a placeholder here

haammar-ledger: Thinking twice, I think I agree with @barjonr: in all 3 cases (Algorand, Cosmos, Ethereum) most values are passed as optional<string>, so it would be more consistent to adapt the few ones that were passed as number?

PR #650, opened 2020-08-31
LLC-677: Rework dust filtering for BTC family

hlafet-ledger:
I created "dust policies": 
- the "per kbyte" policy: where the dust is calculated dynamically depending to the size of the transaction (for the moment applied only on the BTC and BTC TEST currencies) (Please check with me if the estimated size is calculated correctly)
- the "fixed" policy (the old mode) which is kept for the moment for the other currencies, but can be migrated easily (by editing core/src/wallet/bitcoin/networks.cpp)


core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.h
```diff
@@ -144,6 +146,9 @@ namespace ledger {
                                                    const api::Currency &currency,
                                                    const std::string &keychainEngine);
 
+            static int64_t computeDustAmount(const api::Currency &currency, int32_t size);
```
alekece, line 149: What is the rational to have a signed `size` here ? I think negative size doesn't make any sense both for the computation and the blockchain protocol.

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -425,9 +431,21 @@ namespace ledger {
 
             // Minimum amount from which we are willing to create a change for it
             // We take the dust as a reference plus the cost of spending this change
-            const int64_t minimumChange = currency.bitcoinLikeNetworkParameters->DustAmount + signedUTXOCost;
 
-            buddy->logger->debug("Start filterWithKnapsackSolver, target range is {} to {}", amountWithFixedFees, amountWithFixedFees + minimumChange);
+            const int64_t dustAmount_fixedAndOutputPart = BitcoinLikeTransactionApi::computeDustAmount(currency, 
+                                    (fixedSize.Max + (buddy->request.outputs.size() * oneOutputSize)));
+            
+            const int64_t dustAmount_OneInputPart = BitcoinLikeTransactionApi::computeDustAmount(currency,
```
alekece, line 438: Keep same naming convention for variable.

```suggestion
            const int64_t dustAmountFixedAndOutputPart = BitcoinLikeTransactionApi::computeDustAmount(currency, 
                                    (fixedSize.Max + (buddy->request.outputs.size() * oneOutputSize)));
            
            const int64_t dustAmountOneInputPart = BitcoinLikeTransactionApi::computeDustAmount(currency, 
```

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeStrategyUtxoPicker.cpp
```diff
@@ -425,9 +431,21 @@ namespace ledger {
 
             // Minimum amount from which we are willing to create a change for it
             // We take the dust as a reference plus the cost of spending this change
-            const int64_t minimumChange = currency.bitcoinLikeNetworkParameters->DustAmount + signedUTXOCost;
 
-            buddy->logger->debug("Start filterWithKnapsackSolver, target range is {} to {}", amountWithFixedFees, amountWithFixedFees + minimumChange);
+            const int64_t dustAmount_fixedAndOutputPart = BitcoinLikeTransactionApi::computeDustAmount(currency,
```
alekece, line 435: Same question, why use signed integers here ?

core/src/wallet/bitcoin/transaction_builders/BitcoinLikeUtxoPicker.cpp
```diff
@@ -237,7 +241,10 @@ namespace ledger {
         BitcoinLikeUtxoPicker::createFilteredUtxoFunction(const BitcoinLikeTransactionBuildRequest &request,
                                                           const std::shared_ptr<BitcoinLikeKeychain> &keychain,
                                                           const BitcoinLikeGetUtxoFunction &getUtxo) {
-            auto minAmount = getCurrency().bitcoinLikeNetworkParameters.value().DustAmount;
+
+            const auto basicTransactionSize = BitcoinLikeTransactionApi::estimateSize(1, 1, getCurrency(), keychain->getKeychainEngine());
+            int64_t dustAmount = BitcoinLikeTransactionApi::computeDustAmount(getCurrency(), basicTransactionSize.Max);
```
alekece, line 246: Ditto

alekece (changes requested):
Great first iteration ! Basically you need to clarify your choices about variable types

pollastri-pierre (approved):

core/src/wallet/bitcoin/api_impl/BitcoinLikeTransactionApi.h
hlafet-ledger: I agree with you. But the problem here is that we are forced to use signed integers to manipulate sizes (for example EstimatedSize object) because we are exposing those objects and functions via djinni (and unfortunately djinni doesn't support unsigned integers to keep compliance with java)

PR #655, opened 2020-09-09
LLC-707: Replace QT http client by cpp-httplib in tests

hlafet-ledger:


core/test/lib/libledger-test/CppHttpLibClient.cpp
```diff
@@ -0,0 +1,111 @@
...
+                
+                httplib::Result res = 
+                    (request->getMethod() == core::api::HttpMethod::POST) ? cli.Post(path.c_str(), body, "text/plain") :
+                    (request->getMethod() == core::api::HttpMethod::GET) ?  cli.Get(path.c_str()) :
+                    (request->getMethod() == core::api::HttpMethod::PUT) ?  cli.Put(path.c_str(), body, "text/plain") :
+                                                                            cli.Delete(path.c_str());
+                
+                if (res.error() != httplib::Error::Success || res->status < 200 || res->status >= 300 ) {
+                    request->complete(nullptr, std::experimental::optional<core::api::Error>(
+                        core::api::Error(core::api::ErrorCode::HTTP_ERROR, res->reason)));
```
haammar-ledger, line 91: For the second param here, you should build a string containing both the status and reason, something like:

`fmt::format("{}: {}", res->status, res->reason)`

Both will be useful info in case of failure.

core/test/lib/libledger-test/CppHttpLibClient.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define CPPHTTPLIB_OPENSSL_SUPPORT
+
+#include <functional>
+#include <api/HttpClient.hpp>
+#include <api/HttpUrlConnection.hpp>
+#include <api/HttpRequest.hpp>
+#include <api/HttpReadBodyResult.hpp>
+#include <api/HttpMethod.hpp>
+#include <api/ExecutionContext.hpp>
+#include <httplib.h>
```
haammar-ledger, line 12: Some of these headers can be included in the `.cpp` file.

haammar-ledger (commented):
Looks good overall!

Main issue is to modify the headers generation instead of the generated files, the rest is just details.

gagbo:
Seems really useful !

Would it be possible to log the http calls (either using a logger or just bare `print` controlled by an attribute) ? 

Having the calls, duration and optionnally the response body would be useful in test output instead of having to litter feature code with print statements around http calls

PR #658, opened 2020-09-30
[jira/LLC-710]Remove qt dependency in cleanFs

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LLC-710


core/src/utils/FilesystemUtils.cpp
```diff
@@ -29,19 +29,31 @@
...
+#include <experimental/filesystem>
 
-void ledger::qt::FilesystemUtils::clearFs(const std::string& path) {
-    QDir root(QString::fromStdString(path));
-    for (const auto& file : root.entryInfoList(QDir::AllEntries | QDir::NoDotAndDotDot)) {
-        if (file.isDir() && file.absoluteFilePath().compare(root.absolutePath()) != 0) {
-            clearFs(file.absoluteFilePath().toStdString());
-        } else if (!file.isExecutable()) {
-            QFile::remove(file.absoluteFilePath());
+using namespace std;
```
gagbo, line 36: doesn't look like it is used here

core/src/utils/FilesystemUtils.cpp
```diff
@@ -29,19 +29,31 @@
  *
  */
 
-#include "FilesystemUtils.h"
-#include <QDir>
-#include <QDebug>
-#include <QCoreApplication>
+#include "FilesystemUtils.hpp"
 #include <iostream>
+#include <experimental/filesystem>
```
gagbo, line 34: We'll need to make sure this works on all targets, I remember that some experimental features did not compile well on some CI targets

core/src/utils/FilesystemUtils.cpp
```diff
@@ -29,19 +29,31 @@
...
+using namespace std;
+namespace fs = std::experimental::filesystem::v1;
+
+bool ledger::core::FilesystemUtils::isExecutable(const std::string& path){
+    fs::path filePath{path};
+#ifdef _WIN32
+	auto extension=filePath.extension().string();
+    return (extension==".exe") || (extension==".bat") || (extension==".com")
+#else
+    return (status(filePath).permissions() & fs::perms::owner_exec)!=fs::perms::none;
```
gagbo, line 45: Only `owner_exec` ? Why not checking `group_exec` and `other_exec` as well ?

core/src/utils/FilesystemUtils.hpp
```diff
@@ -1,13 +1,13 @@
...
+ * FilesystemUtils
  * ledger-core
  *
- * Created by Pierre Pollastri on 20/09/2017.
+ * Created by Huiqi ZHENG on 16/12/2016.
  *
  * The MIT License (MIT)
  *
- * Copyright (c) 2017 Ledger
+ * Copyright (c) 2016 Ledger
```
gagbo, line 10: Should be 2020

core/src/utils/FilesystemUtils.cpp
hzheng-ledger: Yes, I am testing on linux and Mac and Windows on my side to make sure it works.

core/src/utils/FilesystemUtils.cpp
```diff
@@ -29,19 +29,31 @@
...
+void ledger::core::FilesystemUtils::clearFs(const std::string& path) {
+    fs::path filePath{path};
+    for (const auto & file : fs::recursive_directory_iterator(path))
+    {
+        if (!fs::is_directory(file.path())) {
+            if (!FilesystemUtils::isExecutable(file.path().string())){
+        	    fs::remove(file.path());
+            }
         }
     }
```
gagbo: This loop could also delete the empty directories it meets along the way.

It seems that this loop will only delete non executable files in the given path. It is not consistent with the name of the function

core/src/utils/FilesystemUtils.hpp
```diff
@@ -28,20 +28,22 @@
  * SOFTWARE.
  *
  */
+#ifndef LEDGER_CORE_FILESYSTEMUTILS_HPP
+#define LEDGER_CORE_FILESYSTEMUTILS_HPP
```
gagbo, line 32: We use `#pragma once` as much as possible now, so when we change the include guards we change to pragma

hzheng-ledger: fixed

core/src/utils/FilesystemUtils.hpp
hzheng-ledger: fixed

core/src/utils/FilesystemUtils.cpp
hzheng-ledger: fixed

core/src/utils/FilesystemUtils.cpp
hzheng-ledger: fixed

core/src/utils/FilesystemUtils.cpp
```diff
@@ -42,7 +41,10 @@ bool ledger::core::FilesystemUtils::isExecutable(const std::string& path){
     auto extension=filePath.extension().string();
     return (extension==".exe") || (extension==".bat") || (extension==".com")
 #else
-    return (status(filePath).permissions() & fs::perms::owner_exec)!=fs::perms::none;
+    auto permissions=status(filePath).permissions();
+    return ((permissions & fs::perms::owner_exec)!=fs::perms::none)
+    ||((permissions & fs::perms::group_exec)!=fs::perms::none)
+    ||((permissions & fs::perms::others_exec)!=fs::perms::none);
```
gagbo, line 47:
```suggestion
    return (permissions & (fs::perms::owner_exec | fs::perms::group_exec | fs::perms::others_exec))!=fs::perms::none;
```

Should work as well and makes it clearer that we test for any executable flag

core/src/utils/FilesystemUtils.cpp
hzheng-ledger: after some discussion offline. we keep this for this moment

core/src/utils/FilesystemUtils.cpp
hzheng-ledger: tested and some small things are fixed in Cmake file

core/src/utils/FilesystemUtils.cpp
```diff
@@ -3,7 +3,7 @@
  * FilesystemUtils.cpp
```
hlafet-ledger, line 3: why moving files? why not keeping the old qt function and creating a new .cpp and .hpp file for the new one

core/src/utils/FilesystemUtils.hpp
```diff
@@ -28,20 +28,18 @@
...
-
-#include <string>
+#include <iostream>
 
 namespace ledger {
-    namespace qt {
+    namespace core {
         class FilesystemUtils {
         public:
+        	static bool isExecutable(const std::string& path);
```
hlafet-ledger, line 41: indentation

hzheng-ledger: fixed

CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(BOOST_INCLUDEDIR "${CMAKE_CURRENT_SOURCE_DIR}/core/lib/boost/")
 # Notes:
 # > This is a "blind" fix, no available 10.10.x macOS machine,
 # > Issue is specific to 10.10.x, 10.9.5 and > 10.10.x are fine
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.9" CACHE STRING "Minimum OS X version to target for deployment: 10.9" FORCE)
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum OS X version to target for deployment: 10.15" FORCE)
```
hlafet-ledger, line 32: Are supported OS X limited now to 10.15 or earlier ? Would this exclude some users from 10.9 to 10.14 ?

hlafet-ledger (approved):

PR #668, opened 2020-11-10
[COIN-840] cosmos: add support for Stargate and stargate-testnet

gagbo:
EDIT: Summary
Note that a lot of tests have been disabled as they depend
on a testnet chain state that is not reproducible.

Currency support
- add network parameters for stargate and testnet
- explorer
  + add "STARGATE_NODE" explorer engine
  + Add default Stargate explorer endpoint
  + Add tests and Stargate gRPC parsers for the explorer
  + This explorer still uses the legacy endpoints for
    * gas simulation
    * broadcast (assumes SIGN_TEXTUAL_MODE basically)
- observer
  + add "STARGATE_NODE" observer engine

Performance
- Remove dist. info. parsing for validators

Bug fixes
- Add support for valconsAddr from B64
  The new explorer API does not send an easy to use consensus public key
  so we have to add code that can take a base64 string and make a public
  key from it to use in our codebase.
- Allow Deter.Pubkeys from all cosmosbech32
- Transcode ATOM-Stargate validator cons. pubkeys
- Fix Bech32 encoding for non-address values
  The version needed to be also converted to base32 and the resulting
  value should also be padded before being sent to CosmosBech32::encode.
  We didn't have issues before because this code was only called for
  Addresses, and Addresses :
  + have no prepended version
  + don't need padding to the base-32 value to yield a valid Bech32
    encoding
- Stop erroring in StargateExplorer on 4xx or 5xx errors
  (as Stargate nodes return HTTP error code if account doesn't exist yet)
- Make serialization code work pre/post Stargate
- Remove hardcoded mentions of "atom" and "uatom" to use the account's
  wallet's currency's smallest unit instead.
  The context of parsing is given by the explorer, which is given its
  context by the WalletFactory. We pass the currency through everything
  instead of hardcoding values in the code.
- Avoid segfault when a tx has no logs


hlafet-ledger (approved):

gagbo:
Moving back to Draft status, because I'll make this a big monolithic PR with the new Explorer and the new unit/integration tests.

Just the network parameters won't be enough, and that will need to change because of the testnet chain changes that happen upstream as well

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+
+template <typename T>
+void parseFee(const T &node, cosmos::Fee &out)
+{
+    assert((node.HasMember(kGas)));
+    assert((node.HasMember(kAmount)));
+    out.gas = BigInt::fromString(node[kGas].GetString());
+    if (node[kAmount].IsArray()) {
+        const auto &amountArray = node[kAmount].GetArray();
+        out.amount.assign((std::size_t)amountArray.Capacity(), cosmos::Coin());
```
haammar-ledger, line 231: A reason for using `Capacity()` instead of `Size()` here?

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+    transaction.timestamp = DateUtils::fromJSON(node[kTimestamp].GetString());
+
+    if (vNode.HasMember(kMemo)) {
+        transaction.memo = vNode[kMemo].GetString();
+    }
+
+    assert((vNode.HasMember(kMessage)));
+    if (vNode[kMessage].IsArray()) {
+        const auto &msgArray = vNode[kMessage].GetArray();
+        transaction.messages.assign((std::size_t)msgArray.Capacity(), cosmos::Message());
```
haammar-ledger, line 593: same question here

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+    account.sequence = node[kSequence].GetString();
+    account.address = node[kAddress].GetString();
+
+    assert((node.HasMember(kPubKey)));
+    // Pubkey is null for accounts that only received funds
+    // "pub_key": {
+    //   "@type": "/cosmos.crypto.secp256k1.PubKey",
+    //   "key": "AnDGSF7VFQs9VKEceS6cWhfajxh/mmv+Mbu9bCM9BDzK"
+    // },
+    if (node[kPubKey].IsObject()) {
```
haammar-ledger, line 261: If pubkey can be null maybe the `assert()` is too much here, and the `HasMember` should be put into the `if()`?

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+template <typename T>
+void parseMsgBeginRedelegate(const T &n, cosmos::MessageContent &out)
+{
+    cosmos::MsgBeginRedelegate msg;
+
+    msg.delegatorAddress = n[kDelegatorAddress].GetString();
+    msg.validatorSourceAddress = n[kValidatorSrcAddress].GetString();
+    msg.validatorDestinationAddress = n[kValidatorDstAddress].GetString();
+    parseCoin(n[kAmount].GetObject(), msg.amount);
+    out = msg;
```
haammar-ledger, line 315: Missing `assert`s here and in following methods.

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+
+    // inputs: list<CosmosLikeMultiSendInput>;
+    if(n[kInputs].IsArray()){
+        const auto &inputsArray = n[kInputs].GetArray();
+        parseMultiSendInputs(inputsArray, msg.inputs);
+    }
+
+    // outputs: list<CosmosLikeMultiSendOutput>;
+    if(n[kOutputs].IsArray()) {
+        auto outputsArray = n[kInputs].GetArray();
```
haammar-ledger, line 406: Typo here, should be `kOutputs` right?

haammar-ledger (commented):
First batch, reviewed almost everything except the explorers, will try to finish tomorrow.

core/test/database/query_filters_tests.cpp
```diff
@@ -42,13 +42,13 @@ TEST(QueryFilters, SimpleFilter) {
 
 TEST(QueryFilters, DoubleConditionFilter) {
     auto filter = api::QueryFilter::accountEq("my_account")->op_and(api::QueryFilter::blockHeightGt(12000));
-    EXPECT_EQ(std::dynamic_pointer_cast<QueryFilter>(filter)->getHead()->toString(), "o.account_uid = :account_uid AND o.block_height > :block_height");
+    EXPECT_EQ(std::dynamic_pointer_cast<QueryFilter>(filter)->getHead()->toString(), "o.account_uid = :account_uid AND b.block_height > :block_height");
```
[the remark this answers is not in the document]
gagbo: I'll change the base branch soon, so maybe this won't be relevant anymore, we'll see

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+
+template <typename T>
+void parseMultiSendInputs(const T &array, std::vector<cosmos::MultiSendInput> &out)
+{
+    auto index = 0;
+    out.assign((std::size_t)array.Size(), cosmos::MultiSendInput());
+    for (const auto &n : array) {
+        auto input = n.GetObject();
+        out[index].fromAddress = input[kAddress].GetString();
+        if (input[kCoins].IsArray()) {
```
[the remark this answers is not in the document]
gagbo: that's a dirty copy paste from the other RPC parser, I think this code should be a filter to be cleaner. But the idea here is to just skip the node "n" if n[kCoins] is not an array, at least now

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
```diff
@@ -0,0 +1,907 @@
...
+        }
+    }
+    return "";
+}
+
+template <class T>
+void parseCoin(const T &node, cosmos::Coin &out)
+{
+    assert((node.HasMember(kAmount)));
+    assert((node.HasMember(kDenom)));
```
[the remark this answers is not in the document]
gagbo: I think I'll use them next time :), another difference to take into account is that this algorand helper also fails in non debug builds, while the raw "assert" aren't compiled in for non-debug builds

core/src/wallet/cosmos/explorers/StargateRpcsParsers.hpp
gagbo: I meant that if the account only received funds, the json is 
```json
"pub_key": null
```
so node[kPubKey] exists, but is not an object.

core/src/wallet/cosmos/factories/CosmosLikeWalletFactory.cpp
```diff
@@ -126,14 +128,15 @@ std::shared_ptr<AbstractWallet> CosmosLikeWalletFactory::build(const WalletDatab
         "44'/<coin_type>'/<account>'/<node>/<address>"));
     // Build wallet
 
+    // FIXME: Hardcoded MUON. getCurrency() returns ATOM when called from live
```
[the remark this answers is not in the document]
gagbo: It was just during the phase where I injected Stargate behaviour in current LL cosmos support, good catch thanks :)

core/test/database/query_filters_tests.cpp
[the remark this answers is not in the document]
gagbo: I have no idea how this change appeared, it looks like a few additional commits (between `live-prerelease` and `develop` at the time I forked) got in. I'll look at how tests are changing, and update that accordingly

core/src/wallet/cosmos/explorers/StargateGaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,1310 @@
...
+                if (!json.HasMember(kDelegationResponses)) {
+                    throw make_exception(
+                        api::ErrorCode::API_ERROR,
+                        fmt::format("The API response from explorer is missing the \"{}\" key", kDelegationResponses));
+                }
+
+                // Handle null
+                if (!json.GetObject()[kDelegationResponses].IsArray()) {
+                    return std::make_shared<BigInt>(BigInt::ZERO);
+                }
```
haammar-ledger, line 599: Consistency: do we want different results when `!HasMember()` and `!IsArray()`?

(applies to a few other places)

core/src/wallet/cosmos/explorers/StargateGaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,1310 @@
...
+
+                return std::make_shared<BigInt>(total_amt);
+            });
+}
+/// Get total available (spendable) balance
+FuturePtr<BigInt> StargateGaiaCosmosLikeBlockchainExplorer::getSpendableBalance(
+    const std::string &account) const
+{
+    // hardcoded muon
```
haammar-ledger, line 698: Not anymore? :)

core/src/wallet/cosmos/explorers/StargateGaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -0,0 +1,1310 @@
...
+        .mapPtr<std::vector<cosmos::Delegation>>(
+            getContext(), [this](const HttpRequest::JsonResult &response) {
+                // Catching http errors because Cosmos code returns 500 when there is no delegation.
+                if (std::get<0>(response)->getStatusCode() >= 300) {
+                     return std::make_shared<std::vector<cosmos::Delegation>>();
+                }
+                const auto &document = std::get<1>(response)->GetObject();
+                auto delegations = std::make_shared<std::vector<cosmos::Delegation>>();
+                // Handle null
+                if (document.HasMember("message")
```
haammar-ledger, line 823: Is it intended to check for "message" here? If so then the constant should be used.

core/src/wallet/cosmos/explorers/GaiaCosmosLikeBlockchainExplorer.cpp
```diff
@@ -213,23 +211,25 @@ void addMsgFeesTo(cosmos::Transaction &transaction, const T &node)
...
     addMsgFeesTo(transaction, node);
 }
 
-}  // namespace
 
 FuturePtr<cosmos::TransactionsBulk> GaiaCosmosLikeBlockchainExplorer::getTransactions(
     const CosmosLikeBlockchainExplorer::TransactionFilter &filter, int page, int limit) const
 {
+    // NOTE : the amount of memory involved here asks for a second pass to make
+    // sure that the least amount of copies is done.
```
haammar-ledger, line 226: Made my head explode, but well done.
Maybe just add a couple comments along the way to help follow the big steps.

haammar-ledger (commented):
2nd and last pass, not many new comments!

Gaia & Stargate explorers have obviously a lot of similar code that should be put in common, but I assume the Gaia one will be deprecated soon after Stargate is live, so no need to bother if that's the case.

core/src/wallet/cosmos/explorers/StargateGaiaCosmosLikeBlockchainExplorer.cpp
gagbo: yeah, because:
-  if there is no `delegation_responses` key at all, it's an issue with the explorer and we can't recover from it.
- if everything is correct, and the user has no current delegation at all, then the explorer response will be `"delegation_responses": null`. In this case that means the account has 0 delegated balance.

That's what the (arguably too small) "Handle null" comment means

core/src/wallet/cosmos/explorers/StargateGaiaCosmosLikeBlockchainExplorer.cpp
gagbo: > 2nd and last pass, not many new comments!
> 
> Gaia & Stargate explorers have obviously a lot of similar code that should be put in common, but I assume the Gaia one will be deprecated soon after Stargate is live, so no need to bother if that's the case.

Indeed they should not live together. As we plan to never make this live together I'll just delete the non-Stargate Gaia Explorer and raise an error in the wallet factory if the explorer_engine is still `COSMOS_NODE`

core/src/wallet/cosmos/explorers/StargateGaiaCosmosLikeBlockchainExplorer.hpp
```diff
@@ -0,0 +1,198 @@
...
+   private:
+    std::shared_ptr<HttpClient> _http;
+    api::Currency _currency;
+    /// GRPC namespace of the endpoints to use
+    std::string _ns{"cosmos"};
+    /// GRPC version of the endpoints to use
+    std::string _version{"v1beta1"};
+    /// Add a fees message to a transaction for internal purpose
+    template <typename T>
+    void addMsgFeesTo(cosmos::Transaction &transaction, const T &node) const;
```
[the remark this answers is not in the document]
gagbo: I put it in the private section because no one other than the explorer implementation should run this code, to avoid having multiple fees messages in transactions

haammar-ledger (commented):
Code LGTM :heavy_check_mark: 

Questions about the tests:
- Should also remove all tests `ledger-core-integration-CosmosLikeWalletSynchronization.*` now that legacy Cosmos code is removed, right?
- Are there plans to make the disabled tests work before Stargate release? Adapting them will probably be a lot of tedious work... not mandatory for this PR anyway.

gagbo:
Most of these tests are disabled because they rely on finding external transactions/accounts and the chains reset/change too often for that. I don't know when it'll be possible to reactivate them. I'd like to get what I have right now to be green first :p

PR #670, opened 2020-11-16
Don't force value/wipe when on delegation

twilgenbus-ledger:


twilgenbus-ledger:
Merging to allow pushing to S3 (in order to test locally on mac)

PR #671, opened 2020-11-23
[XTZ] Add confirmations parsing to TezosTransactionParser

twilgenbus-ledger:


gagbo (approved):
I guess we'll have to add a proper "Parse transaction from explorer" test later. It's not really reassuring that such a change doesn't change the status of the test suite

twilgenbus-ledger:
@gagbo thanks for your review!

I definitely agree with you. Instead of later, I'd like to add this test right now if possible (to prevent forgetting about it). Since I haven't done C++ in a while and since am not too familiar with the libCore architecture yet, is it something we could do in pair?
If time's a little short on you, any pointers/advices on how to best add those tests?

gagbo:
Not a lot of time now, but basically

- you need to find the entrypoint of the transaction parsing flow (this file seems to be full of helpers right now)
- You can fetch any transaction to parse directly from the explorer and use the json as the input
- Then write a test around it, most probably in `core/test/tezos/tx_parser_test.cpp`
- Add the new source file in `core/test/tezos/CmakeLists.txt` (there are only main.cpp and address_test.cpp being compiled for now so a new file would be ignored)
- and you can just add your tests there. Currently the stargate tests I'm writing look like this :
```cpp
TEST(CosmosGaiaParser, StargateAccount) {
  // Input data
  // Output (somewhat modified) of
  // /cosmos/auth/v1beta1/accounts/cosmos1x5wgh6vwye60wv3dtshs9dmqggwfx2ldnqvev0
  const std::string explorerResponse =
      // clang-format off
    R"json(
{
  "account": {
    "@type": "/cosmos.auth.v1beta1.BaseAccount",
    "address": "cosmos1x5wgh6vwye60wv3dtshs9dmqggwfx2ldnqvev0",
    "pub_key": {
      "@type": "/cosmos.crypto.secp256k1.PubKey",
      "key": "AnDGSF7VFQs9VKEceS6cWhfajxh/mmv+Mbu9bCM9BDzK"
    },
    "account_number": "11",
    "sequence": "1"
  }
}
    )json";
  // clang-format on

  // Execution of tested function
  rapidjson::Document accountJSON;
  accountJSON.Parse(explorerResponse.c_str());
  cosmos::Account parsedResult;
  stargate_rpcs_parsers::parseAccount(accountJSON.GetObject(), parsedResult);

  // Assertions
  EXPECT_EQ(parsedResult.accountNumber, "11");
  EXPECT_EQ(parsedResult.address,
            "cosmos1x5wgh6vwye60wv3dtshs9dmqggwfx2ldnqvev0");
  EXPECT_EQ(parsedResult.pubkey, "cosmospub1addwnpepqfcvvjz7652sk0255yw8jt5utgta4rcc07dxhl33hw7kcgeaqs7v5p3n38j");
  EXPECT_EQ(parsedResult.sequence, "1");
  EXPECT_EQ(parsedResult.balances.size(), 0);
}
```

hlafet-ledger (approved):

PR #672, opened 2020-11-27
Bitcoin sync fix

hzheng-ledger:
extend the keychain before sync so that it covers all the addresses


core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -197,6 +197,25 @@ namespace ledger {
...
+                    int currentBatchIndex = 0;
+                    int txSize;
+                    do {
+                        auto batch = vector::map<std::string, std::shared_ptr<AddressType>>(
+                            buddy->keychain->getAllObservableAddresses(currentBatchIndex * buddy->halfBatchSize, (currentBatchIndex + 1) * buddy->halfBatchSize - 1),
+                            [](const std::shared_ptr<AddressType>& addr) -> std::string {
+                                return addr->toString();
+                            }
+                        );
+                        auto txs = ledger::core::async::wait(_explorer->getTransactions(batch, optional<std::string>(), optional<void*>()));
```
pollastri-pierre, line 212: We must not wait, asynchronous task, waiting on a thread mean using 2 thread just for one task. This part needs to be rewritten, to avoid locking. You can use a combination of Future and Future::flatMap to avoid locking threads.

pollastri-pierre (changes requested):

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
hzheng-ledger: @pollastri-pierre Fixed. Could you please review my change again? Thanks

core/src/wallet/common/synchronizers/AbstractBlockchainExplorerAccountSynchronizer.h
```diff
@@ -246,6 +246,8 @@ namespace ledger {
                         buddy->token = Option<void *>(t);
                     }
                     return unit;
+                }).template flatMap<Unit>(account->getContext(), [buddy, self](const Unit&) {
+                    return self->extendKeychain(0, buddy);
```
pollastri-pierre, line 250: We can do a small optim by starting to extend the keychain starting by the last known batch.

If we already synchronize the wallet once and know that there is 2 batches. Next time we only need to try to fetch batch 3. If batch 3 has some. transaction we can continue. to fetch more batches, if it doesn't we can stop "extendKeychain" after one http call instead of 2 with the current code.

pollastri-pierre (changes requested):

pollastri-pierre (approved):

PR #678, opened 2020-12-15
Add a getTokenBalance method to tezos account and explorers

barjonr:
Add a method to get FA1.2 token balance for a given tz address using better-call-dev api.

### DIFF
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -31,14 +31,11 @@
 #include "ExternalTezosLikeBlockchainExplorer.h"
 #include <api/Configuration.hpp>
 #include <api/ErrorCode.hpp>
+#include <api/TezosConfiguration.hpp>
 #include <api/TezosConfigurationDefaults.hpp>
 
 namespace ledger {
     namespace core {
-        namespace {
-            const std::string BCD_API_ENDPOINT = "https://api.better-call.dev/v1";
-        }
-
         ExternalTezosLikeBlockchainExplorer::ExternalTezosLikeBlockchainExplorer(
                 const std::shared_ptr<api::ExecutionContext> &context,
                 const std::shared_ptr<HttpClient> &http,
@@ -48,6 +45,8 @@ namespace ledger {
                 TezosLikeBlockchainExplorer(configuration, {api::Configuration::BLOCKCHAIN_EXPLORER_API_ENDPOINT}) {
             _http = http;
             _parameters = parameters;
+            _bcd = configuration->getString(api::TezosConfiguration::BCD_API)
+                .value_or(api::TezosConfigurationDefaults::BCD_API_ENDPOINT);
         }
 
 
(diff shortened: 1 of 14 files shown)


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -371,19 +375,45 @@ namespace ledger {
...
+                                                             const std::string& tokenAddress) const {
+            const auto parseNumbersAsString = true;
+            return _http->GET(fmt::format("/account/mainnet/{}", accountAddress), {}, BCD_API_ENDPOINT)
+                .json(parseNumbersAsString)
+                .mapPtr<BigInt>(getContext(),
+                    [=](const HttpRequest::JsonResult &result) {
+                       const auto &json = *std::get<1>(result);
+                       if ((!json.IsObject() ||
+                           !json.HasMember("tokens") ||
+                           !json["tokens"].IsArray()) && !json.IsString()) {
```
hlafet-ledger, line 389: why this condition: !json.IsString()

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -371,19 +375,45 @@ namespace ledger {
...
+                           throw make_exception(api::ErrorCode::HTTP_ERROR,
+                                                fmt::format("Failed to get tokens for {}", accountAddress));
+                       }
+
+                       const auto tokens = json["tokens"].GetArray();
+                       for (const auto& token : tokens) {
+                           if (token["contract"].GetString() == tokenAddress) {
+                               return std::make_shared<BigInt>(BigInt::fromString(token["balance"].GetString()));
```
hlafet-ledger, line 397: Are we sure to have always "contract" and "balance" in each token?
the lib will crash if one of these 2 fields is missed

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -34,6 +34,10 @@
 
 namespace ledger {
     namespace core {
+        namespace {
+            const std::string BCD_API_ENDPOINT = "https://api.better-call.dev/v1";
```
hlafet-ledger, line 38: hardcoded endpoint ?

hlafet-ledger (changes requested):

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
barjonr: Right now yes, but I will add HasMember(...) to prevent crashes in case there is a change in the future

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
barjonr: I think I copied this from another place in the code, but you are right, it does not make much sense, I'll remove it

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
barjonr: I am not sure how to do it differently (except defining it in a djinni file). If I understand correctly, the endpoint in the configuration will point to TZSTATS, so I cannot overwrite it otherwise the other functions will not work anymore

hlafet-ledger: We should not change the endpoint for other functions.
But, we may define a specific endpoint in configuration for getTokenBalance.
By the way, Is this functionality not provided by the current tzstats api ?

barjonr: I did not find what I needed with tzstats, and I've been told that the tezos foundation recommanded us to use BCD to deal with tokens.

hlafet-ledger (approved):

PR #679, opened 2020-12-17
Jira/coin 1151: rework getting balance for originated accounts

hlafet-ledger:


core/src/wallet/tezos/TezosLikeWallet.cpp
```diff
@@ -159,5 +159,17 @@ namespace ledger {
...
+        Option<Amount> TezosLikeWallet::getBalanceFromCache(size_t accountIndex, const std::string& originatedAccount) {
+            return getBalanceCache().get(fmt::format("{}-{}-{}", getCurrency().name, accountIndex, originatedAccount));
+        }
+
+        void TezosLikeWallet::updateBalanceCache(size_t accountIndex, const std::string& originatedAccount, Amount balance) {
+            getBalanceCache().put(fmt::format("{}-{}-{}", getCurrency().name, accountIndex, originatedAccount), balance);
+        }
+
+        void TezosLikeWallet::invalidateBalanceCache(size_t accountIndex, const std::string& originatedAccount) {
+            getBalanceCache().erase(fmt::format("{}-{}-{}", getCurrency().name, accountIndex, originatedAccount));
```
twilgenbus-ledger, line 171: Maybe extract the format in a constant to make sure it's always formatted the same?

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
```diff
@@ -62,9 +63,11 @@ namespace ledger {
             }
 
             static inline void update_balance(std::shared_ptr<api::Operation> const& op, BigInt& sum) {
-                auto value = BigInt(op->getAmount()->toString());
                 auto tzOp = op->asTezosLikeOperation();
                 auto tzTx = tzOp->getTransaction();
+                auto value = (tzTx->getType() == api::TezosOperationTag::OPERATION_TAG_TRANSACTION)
```
twilgenbus-ledger, line 68: Maybe we could just not call `update_balance` when that was not a `OPERATION_TAG_TRANSACTION`?
Since we know this will not change the sum anyway

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
```diff
@@ -142,19 +145,24 @@ namespace ledger {
...
-                    for (auto const &op : ops) {
-                        OperationStrategy::update_balance(op, result);
-                    }
-
-                    return std::make_shared<Amount>(currencies::TEZOS, 0, result);
+            auto localAccount = _originatorAccount.lock();
+            if (!localAccount) {
+                throw make_exception(api::ErrorCode::NULL_POINTER, "Account was released.");
+            }
+            auto wallet = std::dynamic_pointer_cast<TezosLikeWallet>(localAccount->getWallet());
```
gagbo, line 152: we should check for null pointer here to avoid segfaults later

gagbo (commented):
It looks fine, I just don't understand:
- when the balanceCache is supposed to be invalidated ?
- TTLCache means Time To Live Cache ? Where is the duration for this cache defined ?

core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.h
```diff
@@ -115,7 +115,7 @@ namespace ledger {
                                         const std::vector<std::string> &matchableKeys);
 
             virtual Future<std::shared_ptr<BigInt>>
-            getBalance(const std::vector<TezosLikeKeychain::Address> &addresses) = 0;
```
twilgenbus-ledger, line 118: Maybe we could keep `TezosLikeKeychain::Address` instead of `std::string` to keep some more type-safety?

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
hlafet-ledger: update_balance is called by a generic code (shared between several coins), we cannot use tezos specific functions (as getType)

hlafet-ledger:
> It looks fine, I just don't understand:
> 
> * when the balanceCache is supposed to be invalidated ?
> * TTLCache means Time To Live Cache ? Where is the duration for this cache defined ?



> It looks fine, I just don't understand:
> 
> * when the balanceCache is supposed to be invalidated ?
> * TTLCache means Time To Live Cache ? Where is the duration for this cache defined ?

Yes this is a time to live cache. It is defined in the AbstractWallet class. Duration can be configured through api::Configuration::TTL_CACHE, and default value is 30 seconds

core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.h
hlafet-ledger: 1- As you can see for all other functions in this class, address was always a string.
2- the old format was using a std::vector which must contais only one element (else an error)
So to keep the code simple and consistent, I changed the signature of this function

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
twilgenbus-ledger: I see. Then maybe quit the function asap?
```
if (tzTx->getType() != api::TezosOperationTag::OPERATION_TAG_TRANSACTION)
    return;
```

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
hlafet-ledger: The only case that I can see to get a sigfault here is to have a TezosLikeAccount inside a wallet different from a TezosLikeWallet, which is not possible.
I know that we should check for null after a dynamic cast when we are not sure about type matching. But here we are sure about the pointer type.

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
hlafet-ledger: sum can be impacted by fees (even if value is equal to 0). We cannot quit the function

core/src/wallet/tezos/TezosLikeWallet.cpp
hlafet-ledger: I agree. I will create a single function to format the cache key

core/src/wallet/tezos/TezosLikeWallet.cpp
```diff
@@ -159,5 +159,17 @@ namespace ledger {
...
 
+        Option<Amount> TezosLikeWallet::getBalanceFromCache(size_t accountIndex, const std::string& originatedAccount) {
+            return getBalanceCache().get(fmt::format("{}-{}-{}", getCurrency().name, accountIndex, originatedAccount));
+        }
+
+        void TezosLikeWallet::updateBalanceCache(size_t accountIndex, const std::string& originatedAccount, Amount balance) {
+            getBalanceCache().put(fmt::format("{}-{}-{}", getCurrency().name, accountIndex, originatedAccount), balance);
+        }
+
+        void TezosLikeWallet::invalidateBalanceCache(size_t accountIndex, const std::string& originatedAccount) {
```
barjonr, line 174: Is this method useful? I don't see any references to it in the code

core/src/wallet/tezos/delegation/TezosLikeOriginatedAccount.cpp
```diff
@@ -62,9 +63,11 @@ namespace ledger {
             }
 
             static inline void update_balance(std::shared_ptr<api::Operation> const& op, BigInt& sum) {
```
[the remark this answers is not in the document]
hlafet-ledger: This function is called through a template call (see getBalanceHistoryFor in core/src/wallet/common/BalanceHistory.h)
If I change the name here, i have to modify it everywhere including in ethereum code.

core/src/wallet/tezos/TezosLikeWallet.cpp
hlafet-ledger: Yes it is not called. I added it as a helper function for, may be, a future use.

barjonr (approved):

PR #697, opened 2021-01-11
migrate tezos to new synchronization

hlafet-ledger:


hzheng-ledger (approved):
looks good to me after some offline discussion

PR #698, opened 2021-01-14
migrate algorand to new synchronization

hlafet-ledger:


core/src/wallet/algorand/AlgorandAccount.cpp
```diff
@@ -94,26 +96,36 @@ namespace algorand {
...
         const auto wallet = getWallet();
         if (wallet == nullptr) {
             throw Exception(api::ErrorCode::RUNTIME_ERROR, "Wallet reference is dead.");
         }
-        const auto txuid = TransactionDatabaseHelper::putTransaction(sql, getAccountUid(), transaction);
+
         const auto operation = Operation(shared_from_this(), transaction);
+        out.push_back(operation);
 
-        if (OperationDatabaseHelper::putAlgorandOperation(sql, txuid, operation)) {
```
haammar-ledger, line 113: Since `algorand::OperationDatabaseHelper::putAlgorandOperation` isn't used anymore you can probably remove `AlgorandOperationDatabaseHelper.hpp` and `AlgorandOperationDatabaseHelper.cpp` entirely?

core/src/wallet/algorand/AlgorandAccount.cpp
```diff
@@ -94,26 +96,36 @@ namespace algorand {
 
     bool Account::putBlock(soci::session& sql, const api::Block& block)
     {
-        if (BlockDatabaseHelper::putBlock(sql, block)) {
```
haammar-ledger, line 97: Where is `BlockDatabaseHelper::putBlock` called with this new implementation?

core/src/wallet/algorand/AlgorandAccount.cpp
```diff
@@ -94,26 +96,36 @@ namespace algorand {
...
-        }
-        return false;
+        emitNewBlockEvent(block);
+        return true;
     }
 
-    int Account::putTransaction(soci::session &sql, const model::Transaction &transaction)
+    void Account::interpretTransaction(const model::Transaction &transaction, std::vector<Operation> &out)
     {
+        soci::session sql;
```
hzheng-ledger, line 105: Do we need this line of code?

core/src/wallet/algorand/AlgorandAccount.cpp
```diff
@@ -94,26 +96,36 @@ namespace algorand {
...
-        const auto txuid = TransactionDatabaseHelper::putTransaction(sql, getAccountUid(), transaction);
+
         const auto operation = Operation(shared_from_this(), transaction);
+        out.push_back(operation);
 
-        if (OperationDatabaseHelper::putAlgorandOperation(sql, txuid, operation)) {
-            emitNewOperationEvent(operation.getBackend());
-        }
-        return static_cast<int>(operation.getBackend().type);
+        return;
```
hzheng-ledger, line 114: "return" not necessary?

core/src/wallet/algorand/database/AlgorandTransactionDatabaseHelper.hpp
```diff
@@ -32,13 +32,122 @@
...
+    static constexpr auto COL_TX_AXFER_SENDER_ADDRESS = 45;
+
+    static constexpr auto COL_TX_AFRZ_ASSET_ID = 46;
+    static constexpr auto COL_TX_AFRZ_FROZEN = 47;
+    static constexpr auto COL_TX_AFRZ_FROZEN_ADDRESS = 48;
+
+
+    // Helpers to deal with Option<> types,
+    // because SOCI only understands boost::optional<>...
+
```
hzheng-ledger, line 100: a question: why do you move a lot of methods from cpp file to hpp file?
because I think this hpp is included by other files and it will increase the compilation time in theory.

core/src/wallet/algorand/AlgorandAccount.cpp
```diff
@@ -94,26 +96,36 @@ namespace algorand {
 
     bool Account::putBlock(soci::session& sql, const api::Block& block)
     {
-        if (BlockDatabaseHelper::putBlock(sql, block)) {
-            emitNewBlockEvent(block);
-            return true;
-        }
-        return false;
+        emitNewBlockEvent(block);
```
hzheng-ledger, line 99: I assume that we don't have to return a bool for this method? void is enough?
because we don't use the return value.
Also, in the old code we emit a Block Event when it is a new block (the block doesn't exist in the DB)
in the new code, we always emit this event. 
Does it change the logic?

hzheng-ledger:
I leave several questions as comments.
Also maybe you can merge firstly the current nrt_jdk8 branch from the repo into your fork repo so that you can remove the CI build error. Thanks

core/src/wallet/algorand/database/AlgorandOperationsDatabaseHelper.hpp
```diff
@@ -0,0 +1,50 @@
+/*
+ *
+ * BitcoinLikeOperationDatabaseHelper.hpp
```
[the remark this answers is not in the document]
hlafet-ledger: fixed

core/src/wallet/algorand/AlgorandAccount.cpp
hlafet-ledger: done

core/src/wallet/algorand/AlgorandAccount.cpp
hlafet-ledger: 1- in algorand::OperationsDatabaseHelper::bulkInsert for operations block
2- I did a rollback because of the last block query (when Account::putBlock is called by the synchronizer)

core/src/wallet/algorand/database/AlgorandTransactionDatabaseHelper.hpp
hlafet-ledger: because those functions are called in AlgorandOperationsDatabaseHelper.cpp

hzheng-ledger (approved):
looks good to me

PR #700, opened 2021-01-19
migrate cosmos to new synchronization

hlafet-ledger:


core/src/wallet/cosmos/CosmosLikeAccount.cpp
```diff
@@ -395,28 +393,31 @@ int CosmosLikeAccount::putTransaction(soci::session &sql, const cosmos::Transact
...
 }
 
 bool CosmosLikeAccount::putBlock(soci::session &sql, const api::Block &block)
 {
-    if (BlockDatabaseHelper::putBlock(sql, block)) {
-        emitNewBlockEvent(block);
-        return true;
-    }
-    return false;
+    emitNewBlockEvent(block);
```
hzheng-ledger, line 419: I've discussed with pierre. 
we should emit NewBlockEvent only if the block doesn't exist in DB. So I think we should keep the old logic.

core/src/wallet/cosmos/database/CosmosLikeOperationsDatabaseHelper.cpp
```diff
@@ -0,0 +1,973 @@
+/*
+ *
+ * TezosLikeOperationsDatabaseHelper.cpp
```
hzheng-ledger, line 3: change "Tezos" to "Cosmos"

hzheng-ledger (changes requested):
see my comment

hzheng-ledger (approved):
OK for me after some offline discussion

PR #701, opened 2021-01-20
Remove abstract synchronizer from ripple

barjonr:


core/src/wallet/ripple/synchronizers/RippleLikeAccountSynchronizer.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include <events/ProgressNotifier.h>
+
+#include <memory>
+#include <mutex>
+
+namespace ledger {
+namespace core {
+
+class RippleLikeAccount;
+
```
hzheng-ledger, line 48: line 49- line 82
I find the code was copied from AbstractBlockchainExplorerAccountSynchronizer.h
because they are same structs. I think you can use the ones in AbstractBlockchainExplorerAccountSynchronizer.h
instead of copy-paste directly here. 
We can discuss offline if you want

core/src/wallet/ripple/synchronizers/RippleLikeAccountSynchronizer.cpp
```diff
@@ -0,0 +1,568 @@
+/*
+ *
```
hzheng-ledger, line 2: I think as we don't need a lot of methods in AbstractBlockchainExplorerAccountSynchronizer.h any more.
I don't know whether we can do a cleanup for AbstractBlockchainExplorerAccountSynchronizer.h

hzheng-ledger (changes requested):
See my comments please

core/src/wallet/ripple/synchronizers/RippleLikeAccountSynchronizer.cpp
barjonr: If the idea is to create independant synchronizer for each coin, we can just delete the file once it is done

hzheng-ledger (approved):
good to me

PR #742, opened 2021-02-16
Feature/tezos synchronizer

barjonr:


hlafet-ledger:
If the purpose of the last commit in this PR (Fix incremental synchronization) is to fix pagination issue, please revert it. This will be fixed with the PR https://github.com/LedgerHQ/lib-ledger-core/pull/744

core/src/wallet/tezos/synchronizers/TezosLikeAccountSynchronizer.hpp
```diff
@@ -77,17 +77,23 @@ class TezosLikeAccountSynchronizer : public DedicatedContext
...
+            , offset(offset)
         {}
 
         SavedState() = default;
 
         template<typename Archive>
         void serialize(Archive& archive)
         {
-            archive(blockHeight, blockHash);
+            archive(blockHeight, blockHash, offset);
```
hlafet-ledger, line 96: I'm not sure about backward compatibility here. Is this working when upgrading or rollbacking libcore versions ?

hlafet-ledger (changes requested):

core/src/wallet/tezos/synchronizers/TezosLikeAccountSynchronizer.hpp
barjonr: I think the  first synchronization after the upgrade would be a full one (i.e. from block height 0). Anyway, the `SavedState` object is not the same class anymore, so it would probably not be found in the preferences. But after the first synchronization it should be incremental again

PR #766, opened 2021-05-10
Trying to understand Read 0

gagbo:
Using comparison view to put comments


core/core.djinni
```diff
@@ -25,3 +25,19 @@ LedgerCore = interface +c {
...
+    # Total number of milliseconds
+    total_ms: i64;
+    # Number of records.
+    count: i32;
+}
+
+DurationMetrics = interface +c {
+    # Get all duration metrics
+    static getAllDurationMetrics(): map<string, DurationMetric>;
+}
```
gagbo, line 43: Should the metrics here be ported as well to hotfixs_3.5 branch ?

twilgenbus-ledger: I believe this is linked to the `_metrics` endpoint in the WD: https://github.com/LedgerHQ/ledger-wallet-daemon/blob/develop/src/main/scala/co/ledger/wallet/daemon/controllers/StatusController.scala#L53

This was very useful when we were working on NRT. But I don't think it has been integrated in DD nor Prometeus.
Any current plans on this topic @pollastri-pierre ?

PR #767, opened 2021-05-10
Try 2

gagbo:


core/src/config/Networks.cpp
```diff
@@ -43,6 +43,7 @@ namespace ledger {
                     {0x04, 0x88, 0xB2, 0x1E},
                     BitcoinLikeFeePolicy::PER_BYTE,
                     5420,
+                    BitcoinLikeDustPolicy::FIXED,
```
gagbo, line 46: TODO: check where this change in struct is propagated through JNI

gagbo: It's safe

gagbo: https://github.com/ledgerhq/lib-ledger-core/blob/e4d8111f264d84970d5d30e9471b1643e0d64c07/core/src/jni/jni/BitcoinLikeNetworkParameters.cpp#L16-L27

gagbo:
Fixed since

PR #780, opened 2021-06-01
LL-5485 LL-5782 [Algorand] Disable tests based on TestNet

haammar-ledger:
**Context**: we changed our Algorand node provider from PureStake to host it ourselves with Leger infra (https://ledgerhq.atlassian.net/browse/LL-5485).
However our infra doesn't provide a TestNet anymore, so we have to disable the libcore tests that were based on the TestNet.


hzheng-ledger (approved):
Looks good to me

PR #799, opened 2021-07-16
[VG 740] Add MEMPOOL_GRACE_PERIOD_SECS config var

gagbo:
Prevents BTC-like wallets from deleting "mempool/pending" transactions
if the underlying operation is too recent.
- Using 0 as a value is the same behaviour as hotfixs_3.5 currently
- Using std::numeric_limits<int32_t>::max() as a value is (essentially)
the same behaviour as release/3.5.6

Technically, the prevention happens in the BTC-like synchronizer, by
_refusing_ to add mempool transactions in the
`SynchronizationBuddy::transactionsToDrop` map, _if_ the operation
associated with that transaction is too recent (a grace period)

Removes RBF support (a task needs to be added to put it back)

Only BTC like wallets actually care about the GRACE_PERIOD value, all
other wallets act as if the value is 0 (because they don't override
`AbstractWallet::getMempoolGracePeriod()`). In any case, the
non-BTC-like synchronizers never use this code either.


twilgenbus-ledger (approved):
LGTM

PR #807, opened 2021-08-04
fix(btc): correct prev output in optimistic update

gagbo:
As we merge foreign outputs in BTC-like currencies, we cannot directly
access the "previous transaction" output with the index. It is necessary
to scan every output and find it with a matching predicate

Fixes VSD-678

### DIFF
--- a/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
+++ b/core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
@@ -35,6 +35,8 @@
 
 #include <collections/functional.hpp>
 
+#include <algorithm>
+
 #include <memory>
 #include <utils/DateUtils.hpp>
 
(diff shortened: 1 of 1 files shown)


pollastri-pierre (approved):

twilgenbus-ledger (approved):
LGTM

Only thing is maybe we could make it easier to read that critical code by abstracting the whole "findPrevTxOutputValue" in its own tiny function. Wdyt?

gagbo:
> LGTM
> 
> Only thing is maybe we could make it easier to read that critical code by abstracting the whole "findPrevTxOutputValue" in its own tiny function. Wdyt?

yeah why not, it's going to be a bigger PR though

twilgenbus-ledger (approved):
That's a lot better than what I had proposed, but it feels a lot better now! Ready to merge imo.

gagbo:
Merging even if CI is red as only Cosmos related tests fail (probably because of Stargate-related changes) and this version of Core is targeted for Vault

PR #808, opened 2021-08-06
Enable nix-based builds and CI

gagbo:
The goal of this PR is to provide a nix derivation for developers to work in (through `nix-shell`) and for CI to build on any machine without needing extra CCI configuration/orbs (so that later the builds/deployments can be offloaded to the simplest self-hosted runners, with only `nix` as dependency)

TODO
- [x] check macos GHA
- [x] make a ~~postInstall script~~ a smart github workflow step that uploads to S3
- [x] ~~setup a cachix GHA~~ Not using cachix for now, instead we will use github workflows artifacts
- [x] better GHA event filters
- [x] make 2 nix-builds : 
  + 1 release without `doInstallCheck` and the matching `-DENABLE_TESTS=0`, and the postInstall script
  + 1 debug with tests but without the postInstall
- [x] pin `nixpkgs`

### DIFF
--- a/.github/workflows/nix_release_builds.yml
+++ b/.github/workflows/nix_release_builds.yml
@@ -1,0 +1,99 @@
+name: "Nix Release builds"
+on:
+  pull_request:
+    branches:
+      - release/4.0.0
+      - release/4.1.0
+      - release/4.2.0
+      - develop
+  push:
+    branches:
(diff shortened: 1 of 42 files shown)


gagbo:
Still merging because the errors come from flakiness of tzstats http calls. To be handled as VG-877

PR #811, opened 2021-08-19
Refactor nix functions and use gcc10 everywhere

gagbo:
Ref: NixOS/nixpkgs#127249

Also

- [x] Change gha-libcore-jar.nix so that sbt actually builds the JAR with Java 8


gagbo:
It's easier to just use gcc10 than reverting to a working pin commit because of the other dependencies

PR #813, opened 2021-08-30
[VG-944] 4.1 – Port optimistic update fix

gagbo:
See #807


gagbo:
MacOS fails on CCI because of brew update
Windows fails because OpenSSL download is unavailable

PR #814, opened 2021-08-31
Add more context when an optimistic transaction fails

twilgenbus-ledger:


core/src/wallet/bitcoin/BitcoinLikeAccount.cpp
```diff
@@ -737,8 +737,8 @@ namespace ledger {
                 // and the update will occur at next synchro ...
                 // But still let's log that !
                 if (optimisticUpdate.isFailure()) {
-                    self->logger()->warn("{} Optimistic update failed for broadcasted transaction : {}", 
-                        CORRELATIONID_PREFIX(correlationId), txHash);
+                    self->logger()->warn("{} Optimistic update failed for broadcasted transaction [hash: {}]: {}",
+                        CORRELATIONID_PREFIX(correlationId), txHash, optimisticUpdate.getFailure().getMessage());
```
gagbo, line 741: Is it possible to add the error code as well, if available ?

gagbo (approved):

PR #812, opened 2021-08-24
Hotfix handling of unknown stellar transactions operation types

twilgenbus-ledger:
Prevents from writting them in DB and outputs a warning

PR #818, opened 2021-09-03
Forward port stellar fix #812

twilgenbus-ledger:


CHANGELOG.md
```diff
@@ -12,7 +14,6 @@
   Currently defaults to 900 seconds, 1.5 mean BTC block
 - Remove RBF support, that was incompatible with the new Grace Period
   for now
-- Fix optimistic update operation for BTC-like broadcasts when the
```
gagbo, line 15: This line should stay there

core/src/wallet/stellar/StellarLikeAccount.cpp
```diff
@@ -253,11 +253,20 @@ namespace ledger {
...
         void StellarLikeAccount::interpretTransaction(const stellar::Transaction &tx, std::vector<Operation> &out) {
+            auto log = logger();
             if (tx.envelope.type != stellar::xdr::EnvelopeType::ENVELOPE_TYPE_TX_V0 &&
                 tx.envelope.type != stellar::xdr::EnvelopeType::ENVELOPE_TYPE_TX) {
                 // Ignore transaction with unhandled envelope types.
                 return;
             }
+            // Ignore transaction if at least one operation has an unsupported type
+            const auto& operations = stellar::xdr::getOperations(tx.envelope);
+            for (const auto& op : operations) {
```
gagbo, line 264: For later, as it's not the goal of this PR

We could avoid skipping the entire transaction by just skipping the problematic operation with an early `continue` l.315

gagbo (changes requested):

core/src/wallet/stellar/StellarLikeAccount.cpp
twilgenbus-ledger: Not really I believe, because if we crossed an unknown operation during decoding it means we probably didn't how much binary data followed the operation and our binary read cursor is now lost and all following operations are lost.

It's true we could do a best effort in keeping all operations _until_ we meet one we don't know, but then we'd end up with half parsed transactions which is arguably worse than just a all-or-nothing approach.

gagbo (approved):

PR #822, opened 2021-09-15
feat(nix): use ethash as nix dependency to libcore

gagbo:
This commit shows the "blessed" way of externalizing one vendored lib of
libcore into a proper derivation


gagbo:
Obviously this didn't go well because the ethash used in libcore is both outdated (vs the `chfast/ethash` version) and not really compatible with nix since it didn't provide an install target.

So the PR uses a fork of the fork of the old dependency to add the install target to the `src` of `ethash` : ledgerhq/ethash#1

gagbo:
Dropped until LedgerHQ/ethash#1 is dealt with

PR #823, opened 2021-09-15
refactor(nix): Reuse default.nix for tests

gagbo:


gagbo:
Merging even with all failures :

- The JAR step fails because Github Maven doesn't allow to overwrite snapshots
- The tests steps have varied failures, mostly linked to networking issues

PR #834, opened 2021-11-30
Fix CI tests

viktorb-ledger:
This PR is intended to make CI green. It fixes some problems and disables tests that require a longer investigation.

- Cross-platform CI:
-- Fixes Linux and MacOS workflows
-- Disables tests in Windows
-- Disables publishing jar from git non-release branches
- leveldb is removed from dependencies: MemPreferencesBackend is used in tests instead of leveldb.
- Fixes minor problems:
-- VG-2191: DISTINCT ON doesn't work with SQLite
-- VG-2121: reuse of the same keychain name in many tests
-- uint isn't a standard C++ type (breaks Windows build)


core/src/CMakeLists.txt
```diff
@@ -224,7 +224,9 @@ if(APPLE)
         target_link_libraries(ledger-core-interface INTERFACE -static-libstdc++)
     endif()
     target_link_libraries(ledger-core-interface INTERFACE Threads::Threads)
-    target_link_libraries(ledger-core-obj PUBLIC Threads::Threads)
+    if (TARGET ledger-core-obj)
+        target_link_libraries(ledger-core-obj PUBLIC Threads::Threads)
```
jcoatelen-ledger, line 228: You can move this line up, in the correct block. Like at line 101.

viktorb-ledger: Threads is required only in APPLE block. So there is no way to make this shorter.

dkremer-ledger (approved):

.circleci/init_submodules.sh
```diff
@@ -12,22 +12,18 @@ git submodule sync
 #Otherwise can't init submodules
 echo "========> Update all submodules "
 #No need for those
-git submodule update -- cmake-scripts || echo "===========Cmake-scripts submodule already updated"
+git submodule update -- cmake/cmake-scripts || echo "===========Cmake-scripts submodule already updated"
```
jcoatelen-ledger, line 15: oops, my fault :)

.circleci/publish_jar.sh
```diff
@@ -6,6 +6,8 @@ JAVA_API_DIR=api/core/java
 SCALA_API_DIR=api/core/scala
 RESOURCE_DIR=$JAR_BUILD_DIR/src/main/resources/resources/djinni_native_libs
 
+bash tools/generate_interfaces.sh
```
jcoatelen-ledger, line 9: I think it would be preferable to keep scripts simple (KISS). What do you think of removing the `generate_interfaces.sh` for here and add the corresponding line in `config.yml` as another step ?

core/src/CMakeLists.txt
jcoatelen-ledger: Alright !

core/test/database/BaseFixture.cpp
```diff
@@ -74,8 +75,8 @@ std::shared_ptr<WalletPool> BaseFixture::newDefaultPool(std::string poolName) {
             nullptr,
             backend,
             api::DynamicObject::newInstance(),
-            nullptr,
-            nullptr
+            std::make_shared<ledger::core::test::MemPreferencesBackend>(),
```
jcoatelen-ledger, line 78: so great to see dependency injection :clap:

core/src/CMakeLists.txt
jcoatelen-ledger: Alright !

.circleci/publish_jar.sh
viktorb-ledger: done

PR #844, opened 2022-01-07
CI Tests: Web Service Mocking

viktorb-ledger:
This PR changes the way CI tests interact with the external web-services: instead of doing real HTTP requests, the tests use pre-saved responses from `core/tests/<test_group>/http_cache` folders. The use of pre-saved responses can be turned off with CMake parameters.

This PR adds two additional CMake parameters:
- `TESTS_UPDATE_HTTP_CACHE` - to recreate files in `http_cache` folders
- `TESTS_USE_EXTERNAL_WEB_SERVICES` - to run tests against real web-services without pre-saved responses.


CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(ledger-core)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 option(TARGET_JNI "Indicates wheter or not the toolchain must build for JNI or not" OFF)
 option(BUILD_TESTS "Indicates wheter or not the toolchain must build the test or not" ON)
+option(TESTS_UPDATE_HTTP_CACHE "Update HTTP cache for tests" OFF)
+option(TESTS_RUN_WITHOUT_HTTP_CACHE "Don't use cached HTTP responses" OFF)
```
jcoatelen-ledger, line 8: (optional) It looks strange to me that each option is `ON` for adding a behavior and this one is for `ON` removing a behavior. I would propose to rename the cmake variable to  `TESTS_USE_HTTP_CACHE`.

core/test/lib/libledger-test/ProxyHttpClient.cpp
```diff
@@ -1,28 +1,254 @@
...
+#ifdef LOAD_HTTP_CACHE
+                if (!_logger) {
+                    loadCache(impl::TrafficLogger::getCacheFileName());
+                }
+#endif
+            }
+
+            void ProxyHttpClient::execute(const shared_ptr<api::HttpRequest>& request) {
+                auto vector_uint8_to_string = [](const vector<uint8_t>& v) {
+                    return string((const char*)v.data(), v.size());
```
jcoatelen-ledger, line 178: I prefer the following syntax (more aligned with STL and C++17/20 paradigms): `std::string(v.begin(), v.end())`

core/test/lib/libledger-test/ProxyHttpClient.cpp
```diff
@@ -1,28 +1,254 @@
...
+#pragma message "ProxyHttpClient is configured to update HTTP cache"
+#elif !UPDATE_HTTP_CACHE && ALLOW_HTTP_ACCESS && !LOAD_HTTP_CACHE
+#pragma message "ProxyHttpClient is configured to run tests agains real web services"
+#elif !UPDATE_HTTP_CACHE && !ALLOW_HTTP_ACCESS && LOAD_HTTP_CACHE
+#pragma message "ProxyHttpClient is configured to use HTTP cache instead of real web services"
+#else
+#error "Incorrect configuration of HTTP caching"
+#endif
+
+using namespace std;
```
jcoatelen-ledger, line 26: Let's avoid `using namespace std;`, what do you think ?

core/test/lib/libledger-test/CMakeLists.txt
```diff
@@ -5,22 +5,44 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
...
+            MongooseHttpClient.cpp MongooseHttpClient.hpp
+            MongooseSimpleRestServer.cpp MongooseSimpleRestServer.hpp
+            NativePathResolver.cpp NativePathResolver.hpp
+            NativeThreadDispatcher.cpp NativeThreadDispatcher.hpp
+            OpenSSLRandomNumberGenerator.cpp OpenSSLRandomNumberGenerator.hpp
+            UvThreadDispatcher.hpp UvThreadDispatcher.cpp
+            callbacks.cpp callbacks.hpp
+            route.cc route.h)
+
+add_library(ledger-test-proxy-http-client STATIC
```
jcoatelen-ledger, line 27: I would move all `ledger-test-proxy-http-client`-related files in a dedicated subdirectory with its own `CMakeLists.txt` cmake conf file. This way It will be more clear what is the scope of either `ledger-test-proxy-http-client`  and `ledger-test`

core/test/lib/libledger-test/ProxyHttpClient.cpp
viktorb-ledger: done

core/test/lib/libledger-test/CMakeLists.txt
viktorb-ledger: done

CMakeLists.txt
viktorb-ledger: Done

jcoatelen-ledger: ok !

jcoatelen-ledger (approved):
ok for me

PR #857, opened 2022-02-23
Fix CI tests

viktorb-ledger:
This PR fixes several blinking tests.

### DIFF
--- a/core/test/lib/libledger-test/MemPreferencesBackend.cpp
+++ b/core/test/lib/libledger-test/MemPreferencesBackend.cpp
@@ -9,13 +9,15 @@ namespace ledger {
 
             std::experimental::optional<std::vector<uint8_t>> MemPreferencesBackend::get(const std::vector<uint8_t> & key)
             {
-               if (_data.find(key) != _data.end())
+                std::unique_lock<std::mutex> lock(_mtx);
+                if (_data.find(key) != _data.end())
                     return _data.at(key);
                 return optional<std::vector<uint8_t>>();
             }
 
             bool MemPreferencesBackend::commit(const std::vector<api::PreferencesChange> & changes)
             {
+                std::unique_lock<std::mutex> lock(_mtx);
                 for(const auto& change: changes)
                 {
                     switch (change.type) {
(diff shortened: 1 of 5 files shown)


core/test/lib/libledger-test/MemPreferencesBackend.cpp
```diff
@@ -9,13 +9,15 @@ namespace ledger {
 
             std::experimental::optional<std::vector<uint8_t>> MemPreferencesBackend::get(const std::vector<uint8_t> & key)
             {
-               if (_data.find(key) != _data.end())
+                std::unique_lock<std::mutex> lock(_mtx);
```
jcoatelen-ledger, line 12: why not using `std::lock_guard` ?

core/test/lib/libledger-test/MemPreferencesBackend.cpp
```diff
@@ -9,13 +9,15 @@ namespace ledger {
...
-               if (_data.find(key) != _data.end())
+                std::unique_lock<std::mutex> lock(_mtx);
+                if (_data.find(key) != _data.end())
                     return _data.at(key);
                 return optional<std::vector<uint8_t>>();
             }
 
             bool MemPreferencesBackend::commit(const std::vector<api::PreferencesChange> & changes)
             {
+                std::unique_lock<std::mutex> lock(_mtx);
```
jcoatelen-ledger, line 20: I would suggest to move this `lock` into the loop because `changes` does not need to be locked.

core/test/net/http_client_tests.cpp
```diff
@@ -71,6 +73,7 @@ TEST(HttpClient, GET) {
         });
 
         server->start(8000);
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
```
jcoatelen-ledger, line 76: This solution does no seem trustworthy. Not possible to "wait until ready' ?

core/test/lib/libledger-test/MemPreferencesBackend.cpp
viktorb-ledger: done

core/test/lib/libledger-test/MemPreferencesBackend.cpp
viktorb-ledger: done

jcoatelen-ledger (approved):
ok for me

PR #862, opened 2022-03-31
VG-4357: Fix Tezos tx parser to handle new attributes from explorer

jcoatelen-ledger:
## Changes
* Fix the Tezos transaction parser to adapt to the new version of the tzstat API considering `delegation` operations. In particular:
  * now a delegation operation has a `baker` field instead of a `delegate` field
  * now a delegation withdraw operation has a `previous_baker`  field instead of a `receiver` field
* bump libcore version to `4.1.8`

## References
* [VG-4357- All delegations and undelegations are shown with the "UNDELEGATE" icon on tx list (seems to be a gate issue)](https://ledgerhq.atlassian.net/browse/VG-4357)


ktollec-ledger (approved):
Nice :)

PR #863, opened 2022-04-01
Fix/tzindex gas price (outdated)

ktollec-ledger:
Since tzindex 12 (see [CHANGELOG](https://github.com/blockwatch-cc/tzindex/blob/master/CHANGELOG.md#v1200-api-v012-2022-03-25light)), `gas_price` field has been removed.

This PR aims to add the field back by computing it using the following formula : `fee / gas_limit`.
Note that in Tezos, `gas_price` does not really makes sense, the fee is paid in full, whatever how much gas was used.
The `gas_limit` can however be used by the baker to check whether the call is profitable or not.

### DIFF
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -34,8 +34,6 @@
 #include <api/TezosConfiguration.hpp>
 #include <api/TezosConfigurationDefaults.hpp>
 
-#include <iostream>
-
 namespace ledger {
     namespace core {
         ExternalTezosLikeBlockchainExplorer::ExternalTezosLikeBlockchainExplorer(
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -132,8 +132,8 @@ namespace ledger {
         ExternalTezosLikeBlockchainExplorer::getGasPrice() {
             const bool parseNumbersAsString = true;
             // Since tzindex 12.01, we don't have gas_price field anymore
-            // We have to calculate it instead from gas_used and fee
-            const auto gasUsedField = "gas_used";
+            // We have to calculate it instead from gas_limit and fee
+            const auto gasLimitField = "gas_limit";
             const auto feeField = "fee";
 
             return _http->GET("block/head")
@@ -144,11 +144,11 @@ namespace ledger {
                             throw make_exception(
                                 api::ErrorCode::HTTP_ERROR, "Failed to compute gas_price from network, block/head is not a JSON object");
                         }
-                        if (!json.HasMember(gasUsedField) || !json[gasUsedField].IsString()) {
+                        if (!json.HasMember(gasLimitField) || !json[gasLimitField].IsString()) {
                             throw make_exception(
                                 api::ErrorCode::HTTP_ERROR, fmt::format(
                                     "Failed to compute gas_price from network, no (or malformed) field \"{}\" in response",
-                                    gasUsedField));
+                                    gasLimitField));
                         }
                         if (!json.HasMember(feeField) || !json[feeField].IsString()) {
                             throw make_exception(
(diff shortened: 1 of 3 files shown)


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -131,19 +131,36 @@ namespace ledger {
...
+                        }
+                        if (!json.HasMember(feeField) || !json[feeField].IsString()) {
+                            throw make_exception(
+                                api::ErrorCode::HTTP_ERROR, fmt::format(
+                                    "Failed to compute gas_price from network, no (or malformed) field \"{}\" in response",
+                                    feeField));
                         }
-                        const std::string apiGasPrice = json[gasPriceField].GetString();
-                        const std::string picoTezGasPrice = api::BigInt::fromDecimalString(apiGasPrice, 6, ".")->toString(10);
+                        const uint64_t apiGasLimit = std::stoull(json[gasLimitField].GetString());
```
jcoatelen-ledger, line 159: `apiGasLimit` can be defined as a `double` directly so you don't have to cast it after

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -131,19 +131,36 @@ namespace ledger {
...
+                        const double apiFee = std::stod(json[feeField].GetString());
+                        const double gasPrice = apiFee / static_cast<double>(apiGasLimit);
```
jcoatelen-ledger, line 161: Please consider the case where `apiGasLimit` is null

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -131,19 +131,36 @@ namespace ledger {
...
+                        const std::string gasPriceAsString = std::to_string(gasPrice);
+                        const std::string picoTezGasPrice = api::BigInt::fromDecimalString(gasPriceAsString, 6, ".")->toString(10);
```
jcoatelen-ledger, line 163: weird, you gasPrice is a string, then converted to BigInt, then to string, then back again to BigInt. Something is wrong here or I don't understand the logic, can you please explicit?

jcoatelen-ledger (changes requested):

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
ktollec-ledger: Implicit casts are evil :smiling_imp:

jcoatelen-ledger: I don't really see where is the gain in this because you still have to make assumption when casting from `uint64_t` to `double` anyway. BUT I agree, your code is more expressive, which is great. So let's leave it as is :)

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
jcoatelen-ledger: You can initialize your first BigInt on the double

jcoatelen-ledger: Finally OK

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -131,19 +131,36 @@ namespace ledger {
...
 
-                        if (!json.IsObject() || !json.HasMember(gasPriceField) ||
-                            !json[gasPriceField].IsString()) {
-                            throw make_exception(api::ErrorCode::HTTP_ERROR,
-                                                 fmt::format("Failed to get gas_price from network, no (or malformed) field \"{}\" in response", gasPriceField));
+                        if (!json.IsObject()) {
+                            throw make_exception(
+                                api::ErrorCode::HTTP_ERROR, "Failed to compute gas_price from network, block/head is not a JSON object");
+                        }
+                        if (!json.HasMember(gasLimitField) || !json[gasLimitField].IsString()) {
```
jcoatelen-ledger, line 147: please, for the sake of retrocompatibility,  consider to check if `gas_price` is in the payload as a fallback solution

jcoatelen-ledger (approved):
LGTM

jcoatelen-ledger:
Handled in #864

PR #865, opened 2022-04-12
VG-4576 Fix tezos reveal operation

jcoatelen-ledger:
## Changes
* Fix incorrect conversions in gas price computation. This was due to incorrect locale set in the libcore, the fix consists to set the locale in function scope to make sure to use the correct character as decimal separator.
* Fix formula used to compute the gas price. The correct formula, [according to tzindex](https://github.com/blockwatch-cc/tzindex/blob/df40c4f3fc2887d1fff673a3369f069278e3447d/etl/op.go#L641), is to divide fee with gas_used and not gas_limit.
* Fix reveal fee formula used in tezos transaction crafter
* Bump patch version to `4.1.10`

## Todos
- [x] Check tests pass (waiting for the result)

### DIFF
--- a/core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
+++ b/core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
@@ -461,7 +461,7 @@ namespace ledger {
                             pub_key.c_str(), static_cast<SizeType>(pub_key.length()), allocator);
                         revealOp.AddMember("public_key", vString, allocator);
 
-                        static const auto fee = "257000";
+                        static const auto fee = "10000";
                         vString.SetString(fee, static_cast<SizeType>(std::strlen(fee)), allocator);
                         revealOp.AddMember("fee", vString, allocator);
 
--- a/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
@@ -135,8 +135,8 @@ namespace ledger {
         ExternalTezosLikeBlockchainExplorer::getGasPrice() {
             const bool parseNumbersAsString = true;
             // Since tzindex 12.01, we don't have gas_price field anymore
-            // We have to calculate it instead from gas_limit and fee
-            const auto gasLimitField = "gas_limit";
+            // We have to calculate it instead from gas_used and fee
+            const auto gasUsedField = "gas_used";
             const auto feeField = "fee";
             // We still use the legacy field in case we have a rollback
             const auto gasPriceField = "gas_price";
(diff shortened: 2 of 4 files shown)


core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
```diff
@@ -154,27 +172,31 @@ namespace ledger {
...
                                 throw make_exception(
-                                    api::ErrorCode::HTTP_ERROR, "Failed to compute gas_price from network, gas_limit of HEAD block is 0"
+                                    api::ErrorCode::HTTP_ERROR, "Failed to compute gas_price from network, gas_used of HEAD block is 0"
                                 );
                             }
-                            const double apiFee = std::stod(json[feeField].GetString());
-                            const double numericGasPrice = apiFee / static_cast<double>(apiGasLimit);
-                            gasPrice = std::to_string(numericGasPrice);
+
+                            double apiFee = std::stod(json[feeField].GetString());
```
glethuillier, line 194: `std::stod` could potentially throw an exception if the value of `feeField` cannot be converted. Perhaps should we `try/catch` it?

glethuillier: (Same thing for `std::stoull` above).

core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
```diff
@@ -461,7 +461,7 @@ namespace ledger {
                             pub_key.c_str(), static_cast<SizeType>(pub_key.length()), allocator);
                         revealOp.AddMember("public_key", vString, allocator);
 
-                        static const auto fee = "257000";
+                        static const auto fee = "10000";
```
YBadiss, line 464: Is there some doc link we could add for this magic value?

ktollec-ledger: https://tezos.stackexchange.com/a/787

YBadiss: Not really a doc though ^^ And the source ends on a 404

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
twilgenbus-ledger: fyi+: LL is not longer following our changes. So don't worry @ktollec-ledger, that's not to be supported.

core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
jcoatelen-ledger: This code aimed to produce a dry-run-ready payload for [estimating gas limit](https://github.com/LedgerHQ/lib-ledger-core/blob/dca4496b1de21fecd303ce8dafc05d94088f9fd1/core/src/wallet/tezos/explorers/TezosLikeBlockchainExplorer.cpp#L223) during a transaction crafting (unsigned tx). The value should be low enough to avoid triggering a balance error from the node but, after some empirical tests, this specific value does not change the result of the estimated gas limit.

core/src/wallet/tezos/explorers/ExternalTezosLikeBlockchainExplorer.cpp
jcoatelen-ledger: Fixed in 0d37eacba35870aadd1d883088d1c7b2a3e1b8c6

core/src/wallet/tezos/api_impl/TezosLikeTransactionApi.cpp
jcoatelen-ledger: @YBadiss considering [this source of info](https://medium.com/chain-accelerator/how-to-use-tezos-rpcs-16c362f45d64) it appears that `fee` can be anything, I even have seen `0` as a value on some forums. So considering this, do you accept the changes?

PR #872, opened 2022-05-19
Add some index for transaction listing

BertrandD:
Improve performances when getting operations (`GET /pools/pool_name/wallets/wallet_name/accounts/acc_idx/operations?batch=300&full_op=1`) for big accounts


jcoatelen-ledger (approved):
Looks nice ! Good catch !

PR #873, opened 2022-05-24
Add operation count query

BertrandD:
-> To get the number of operations, WD AccountController currently fetch all operations (sql) and then count the operations (scala). This PR adds a count() method, thus the counting is done by the database wich is much faster.

It aims to fix a huge memory usage when fetching account information on huge accounts with thousands of transactions


core/idl/wallet/common/wallet.djinni
```diff
@@ -265,6 +272,9 @@ OperationQuery = interface +c {
     # Execute query to retrieve operations.
     # @param callback, if execute method succeed, ListCallback object returning a List of Operation objects
     execute(callback: ListCallback<Operation>);
+    # Count operations by type
+    # @param callback, if execute method succeed, ListCallback object returning a List of OperationCount objects
+    count(callback: ListCallback<OperationCount>);
```
jcoatelen-ledger, line 277: Could be `const`

core/src/wallet/common/OperationQuery.cpp
```diff
@@ -110,6 +111,41 @@ namespace ledger {
...
+          return _builder.select("o.type, count(*)")
+              .from("operations").to("o")
+              .outerJoin("blocks AS b", "o.block_uid = b.uid")
+              .groupBy("o.type")
+              .execute(sql);
+        }
+
+        void OperationQuery::performCount(std::vector<api::OperationCount> &operations) {
+          soci::session sql(_pool->getPool());
+          soci::rowset<soci::row> rows = performCount(sql);
```
jcoatelen-ledger, line 139: I know that the repo is not representative of this good practice but let's start and keep defining variables `const` by default ;)  (here `rows`, `row`, `type` and `count` are good candidates)

core/src/database/query/QueryBuilder.cpp
```diff
@@ -133,5 +137,10 @@ namespace ledger {
             return *this;
         }
 
+        QueryBuilder &QueryBuilder::groupBy(std::string &&group) {
```
jcoatelen-ledger, line 140: Passing an argument as rvalue reference is rarely necessary and often considered as over-engineering. Here `QueryBuilder` is telling "Hello, I need to own the full string resource to perform the groupBy" where it should tell "Hello, I'm going to use my own resource based on what you are giving to me". In that sense, it should expect a value (`QueryBuilder &QueryBuilder::groupBy(std::string group)`). This still allows to pass an rvalue reference (by moving the given string, or passing an rvalue) and allow type erasure. You can then move your local copy to your resource as you did.

jcoatelen-ledger (changes requested):
Minor changes but let's consider them :)

core/idl/wallet/common/wallet.djinni
BertrandD: I can't make  `performCount` const as it uses `_builder` that do not offer const methods by design.

core/src/wallet/common/OperationQuery.cpp
BertrandD: done

core/src/database/query/QueryBuilder.cpp
BertrandD: Right, I just copied&renamed an other method without more thinking...

core/idl/wallet/common/wallet.djinni
jcoatelen-ledger: Ok then :/

jcoatelen-ledger (approved):

PR #878, opened 2022-06-15
VG-5378 SQLite (sqlcipher) 3.26.0 -> 3.37.2

viktorb-ledger:
Update SQLite to the recent version.
Motivation: use the support of ALTER TABLE DROP COLUMN to simplify DB schema migration.

Here is the procedure I used to obtain these new files:

```
git clone https://github.com/sqlcipher/sqlcipher.git && cd sqlcipher
git checkout v4.5.1
mkdir bld && cd bld
../configure
make sqlite3.c
```

### DIFF
--- a/core/lib/sqlcipher/sqlite3.h
+++ b/core/lib/sqlcipher/sqlite3.h
@@ -43,7 +43,30 @@ extern "C" {
 
 
 /*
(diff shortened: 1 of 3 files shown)


BertrandD (approved):
LGTM

PR #879, opened 2022-06-15
VG-2553 Fix input data dropped when > 255 chars

BertrandD:
A condition was existing to avoid data longer than a VARCHAR, but now the input data is store in a TEXT field (Cf `core/src/database/migrations.cpp:483`)


jcoatelen-ledger (approved):
Ok assuming addresses used in tests are not sensible data

PR #882, opened 2022-06-22
VG-2096 support batch size argument in when getting transactions list from explorers

BertrandD:
* When getting exactly 1k transaction, checks if first and last block hash are identical.
* If so, we might be in the edge case where there is more than 1k transaction in this block for this address.


twilgenbus-ledger (approved):
LGTM! 👍

jcoatelen-ledger (approved):

PR #883, opened 2022-06-27
VG-737 Fix crash on base58 decode with empty string

BertrandD:
Fix crash of WD when creating account with empty extended key


jcoatelen-ledger (approved):
nice catch !

PR #885, opened 2022-07-08
Upgrade to c++17

BertrandD:
I propose to upgade to c++17 in order t be able to use `if constexpr ()` (If validated, it will be used to simplify some changes in https://github.com/LedgerHQ/lib-ledger-core/pull/884)


jcoatelen-ledger (approved):
:older_man:  -> :adult:

PR #888, opened 2022-07-20
Fix explorer query to get only confirmed txs

BertrandD:
To avoid parsing errors on transactions with `block: null`


BertrandD:
Closed because we want to support pending transactions: https://ledgerhq.atlassian.net/browse/VG-8128

PR #895, opened 2022-08-10
Fix circle ci macos

BertrandD:


.circleci/config.yml
```diff
@@ -104,7 +104,7 @@ jobs:
                 ~/cmake-3.16.5-Linux-x86_64/bin/ctest --output-on-failure -timeout 120
   build_macos_release:
     macos:
-      xcode: "12.0.0"
+      xcode: "12.5.1"
```
jcoatelen-ledger, line 107: If i'm not mistaken, there will be a new version to come, see with Victor about that, for reference: https://github.com/LedgerHQ/lib-ledger-core-build-env/pull/9

jcoatelen-ledger (approved):

.circleci/config.yml
viktorb-ledger: @jcoatelen-ledger
How that build env update can fix xcode version?

jcoatelen-ledger: it's not related, I made a confusion sorry

jcoatelen-ledger (approved):
LGTM

PR #896, opened 2022-08-22
VG-8830 Fix Big int parsing overflow

BertrandD:
And:
* Bump libcore version
* Update tzkt explorer default configurations


twilgenbus-ledger (approved):
LGTM!
Maybe a tiny test to seal it up?

jcoatelen-ledger (approved):
erf yet again a cast issue, good catch !

PR #898, opened 2022-08-31
Add Release workflow

BertrandD:
This is a manual workflow to be run from github action page (manual run will be available once PR is merged to `main`)

* Set given version in CMakeLists.txt
* Commit CMakeLists.txt
* Add tag to commit
* Create release note (can be manully disabled)

Here is example of Release generated by this workflow: https://github.com/LedgerHQ/lib-ledger-core/releases/tag/4.3.4-testReleaseWorkflow I will delete it as soon as we validate this PR

### DIFF
--- a/.github/workflows/nix_release_builds.yml
+++ b/.github/workflows/nix_release_builds.yml
@@ -16,10 +16,15 @@ on:
       - develop
     tags:
       - '*'
+  workflow_run:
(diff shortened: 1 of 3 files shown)


jcoatelen-ledger (approved):
I like regexp and automation, go ! 
So great to this coming !

PR #901, opened 2022-09-06
Major CI Improvements

BertrandD:
Add ccache to accelerate CI tests
For this, I had to remove nix as ccache is not working with nix on macos see branch `nix_with_ccache`

Create libcore JAR & publish it to github registry to be used in WD: https://github.com/LedgerHQ/ledger-wallet-daemon/pull/502


tools/publish-jar-local.sh
```diff
@@ -0,0 +1,52 @@
+set -euo pipefail
+
+if [[ -f "$HOME/.jabba/jabba.sh" ]]; then
```
jcoatelen-ledger, line 3: For making sure that the correct version of java is used, I personnally use a different approach: instead of forcing the use of `jabba` (which is not necessarily required for that, the user may want to use something else), I just ensure the current version in the env is the correct one and otherwise I suggest to use jabba. Example from [this script](https://github.com/LedgerHQ/ledger-wallet-daemon/blob/db58572c98ccfe8869eec46510abb3984c48f69a/tooling/clitools/wd/src/core_build_command.sh#L34):

```bash
IFS=. read major minor patch <<<$(java -version 2>&1 | awk -F '"' '/version/ {print $2}')
if [ $major -ne 1 ] | [ $minor -ne 8 ]; then
    printf "Openjdk should match version 1.8\n\n"
    printf "Tip ! you can use:\n"
    printf "$ jabba use adopt@1.8.0-292\n"
    exit 1
fi
```

jcoatelen-ledger: I'm not saying my approach is better, just a suggestion :)

tools/publish-jar-local.sh
```diff
@@ -0,0 +1,52 @@
...
+  echo "Found jabba"
+  source "$HOME/.jabba/jabba.sh"
+  jabba use adopt@1.8.0-292
+fi
+
+JAVA_VERSION=`java -version 2>&1 | awk -F '"' '{print $2}'`
+if [[ $JAVA_VERSION != "1.8.0_292" ]]; then
```
jcoatelen-ledger, line 10: I would better check only major and minor, the patch is not necessary. It makes it very restrictive.

tools/publish-jar-local.sh
```diff
@@ -0,0 +1,52 @@
...
+# This script works with a directory name as first parameter
+# You need to have the libcore already built in this directory with JNI enabled
+# e.g.:
+# cd lib_build_dir_jni
+# cmake .. -DSYS_OPENSSL=ON -DOPENSSL_USE_STATIC_LIBS=TRUE -DTARGET_JNI=ON -DPG_SUPPORT=ON
+# cmake --build . --parallel
+#
+# And then you call the script:
+# ./publish-jar-local.sh ./lib_build_dir_jni
+LIB_BUILD_DIR=$1
```
jcoatelen-ledger, line 24: you should make sure `$1` is not empty

core/test/common/test_config.h
```diff
@@ -26,6 +26,31 @@
...
 
-#define POSTGRES_ON_LOCALHOST "postgres://localhost:5432"
-#define POSTGRES_TEST_DB "test_db"
-#define POSTGRES_TEST_DB_ON_LOCALHOST POSTGRES_ON_LOCALHOST "/" POSTGRES_TEST_DB
+#include <cstdlib>
+#include <string>
+
+inline std::string getPostgresUrl() {
+    const char *host = getenv("POSTGRES_HOST");
+    std::string host_var(host ? host : "localhost");
```
jcoatelen-ledger, line 34: `const` are missing (except for `user_var`)

.github/workflows/release.yml
```diff
@@ -37,14 +49,40 @@ jobs:
...
+    needs: [ Release ]
+    with:
+      release: true
+      version: ${{ github.event.inputs.version }}
+  Publish:
+    name: Publish to WD
+    needs: [Build]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
jcoatelen-ledger, line 76: line duplication?

jcoatelen-ledger (changes requested):
the most comment is on java version to check which seems too restrictive, otherwise it's optional changes

tools/publish-jar-local.sh
BertrandD: `set -u` at the beginning of the scripts ensures this

.github/workflows/release.yml
BertrandD: Nope, first one clones libcore, the second clones wallet-daemon, I need both here

BertrandD: I'll add names to make it more clear

tools/publish-jar-local.sh
BertrandD: Thanks !

core/test/common/test_config.h
BertrandD: Right !

tools/publish-jar-local.sh
jcoatelen-ledger: TIL

.github/workflows/release.yml
jcoatelen-ledger: thks !

jcoatelen-ledger (approved):

PR #904, opened 2022-09-15
VG-9372 Add tracing to sql requests

BertrandD:


core/lib/soci/core/session.cpp
```diff
@@ -16,6 +16,8 @@
 #pragma warning(disable:4355)
 #endif
 
+#include <iostream>
```
viktorb-ledger, line 19: Why do we need `<iostream>` here?

BertrandD: I used it for debug, I forgot to remove, nice catch !

viktorb-ledger (approved):
LGTM

PR #910, opened 2022-09-22
VG-9372 Fix sql session tracer

BertrandD:
Due to concurrency on SessionTracer we had a core dump crash on runtime

This PR adds the ability to have multiple spans at a time in the same SessionTracer, and adds thread-safety on it

### DIFF
--- a/core/src/database/DatabaseSessionPool.cpp
+++ b/core/src/database/DatabaseSessionPool.cpp
@@ -52,22 +52,33 @@ namespace ledger {
                 explicit SessionTracer(std::shared_ptr<api::CoreTracer> tracer) : _tracer(std::move(tracer)) {
                 }
 
-                void startSpan(std::string &string) override {
-                    _span = _tracer->startSpan(string);
-                    _span->setTagStr("db.statement", string);
-                    _span->setTagStr("sql.query", string);
-                    _span->setTagStr("db.system", "postgresql");
-                    _span->setTagStr("span.type", "sql");
-                    _span->setTagStr("service", "postgresql");
+                int startSpan(std::string &string) override {
+                    const std::lock_guard<std::mutex> lock(mut);
+                    auto span = _tracer->startSpan(string);
+                    span->setTagStr("db.statement", string);
+                    span->setTagStr("sql.query", string);
+                    span->setTagStr("db.system", "postgresql");
+                    span->setTagStr("span.type", "sql");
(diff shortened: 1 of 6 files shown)


core/src/database/DatabaseSessionPool.cpp
```diff
@@ -52,22 +52,33 @@ namespace ledger {
...
+                    _traces.at(traceId)->close();
+                    _traces.erase(traceId);
                 }
 
               private:
-                std::shared_ptr<api::Span> _span;
+                std::atomic<int> idx{};
+                std::map<int, std::shared_ptr<api::Span>> _traces;
                 std::shared_ptr<api::CoreTracer> _tracer;
+                std::mutex mut{};
```
jcoatelen-ledger, line 81: please use the same naming convention, `_mutex` for example

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -52,22 +52,33 @@ namespace ledger {
...
+                    int id = idx++;
+                    _traces.insert(std::make_pair(id, span));
+                    return id;
                 }
 
-                void finishSpan() override {
-                    _span->close();
+                void finishSpan(int traceId) override {
+                    const std::lock_guard<std::mutex> lock(mut);
+                    if (_traces.count(traceId) == 0) {
```
jcoatelen-ledger, line 70: prefer `if(_traces.find(traceId) == _traces.end())` for efficiency and expressiveness

core/src/database/DatabaseSessionPool.cpp
```diff
@@ -52,22 +52,33 @@ namespace ledger {
...
-                    _span->setTagStr("service", "postgresql");
+                int startSpan(std::string &string) override {
+                    const std::lock_guard<std::mutex> lock(mut);
+                    auto span = _tracer->startSpan(string);
+                    span->setTagStr("db.statement", string);
+                    span->setTagStr("sql.query", string);
+                    span->setTagStr("db.system", "postgresql");
+                    span->setTagStr("span.type", "sql");
+                    span->setTagStr("service", "postgresql");
+                    int id = idx++;
```
jcoatelen-ledger, line 63: `const std::lock_guard<std::mutex> lock(mut);` can be moved right before this line

jcoatelen-ledger (changes requested):

jcoatelen-ledger (approved):
LGTM !

PR #912, opened 2022-09-30
VG-10108 Fix get last block

BertrandD:
Because LAMA sync do not store all blocks in database, we need to ask explorer for last block


core/src/wallet/bitcoin/BitcoinLikeWallet.hpp
```diff
@@ -88,6 +90,7 @@ namespace ledger {
             std::shared_ptr<BitcoinLikeKeychainFactory> _keychainFactory;
             BitcoinLikeAccountSynchronizerFactory _synchronizerFactory;
             std::chrono::seconds _mempoolGracePeriod;
+            TTLCache<std::string, api::Block> _blockCache;
```
jcoatelen-ledger, line 93: cool idea !

jcoatelen-ledger (approved):
Well done !

PR #914, opened 2022-10-05
Improve Get Balance History

BertrandD:
Do not fetch keychain twice for the same address

Might fix VG-1215

### DIFF
--- a/core/src/utils/Cached.h
+++ b/core/src/utils/Cached.h
@@ -28,27 +28,88 @@
 #pragma once
 
 #include <functional>
+#include <tuple>
 #include <unordered_map>
+namespace hash_tuple {
 
+    template <typename TT>
+    struct hash {
+        size_t
+        operator()(TT const &tt) const {
+            return std::hash<TT>()(tt);
+        }
+    };
+
+    namespace {
+        template <class T>
+        inline void hash_combine(std::size_t &seed, T const &v) {
+            seed ^= hash_tuple::hash<T>()(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+        }
+
+        // Recursive template code derived from Matthieu M.
+        template <class Tuple, size_t Index = std::tuple_size<Tuple>::value - 1>
+        struct HashValueImpl {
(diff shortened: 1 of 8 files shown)


core/src/utils/Cached.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+#pragma once
+
+#include <functional>
+#include <unordered_map>
+
+namespace ledger {
+    namespace core {
+        template <class R, class P>
+        class Cached {
```
jcoatelen-ledger, line 36: This class by design can produce strange behavior: what if I create one instance and call `build` multiple times, how the cache is updated on the different calls to built functions? I would propose two different approaches to handle it (sorry I did not even tried to compile): 
* First possibility is to make one instance of the class per function and to wrap the call. It would look like something like: 
```cpp
template <class R, class P>
class Cached {

public:
    using FunctionType = std::function<R(const P &)>;

    Cached(FunctionType f): function{std::move(f)} {}

    R operator()(const P &) {
        const auto pos = cache.find(param);
        if (pos == cache.cend()) {
            const R contained = f(param);
            cache[param]      = contained;
            return contained;
        }
        return pos->second;
    }

private:
    FunctionType function{};
    std::unordered_map<P, R> cache{};
};
```
* Second possibility is to have a more functional approach with this kind of generic implementation:
```cpp
template <typename R, typename C, typename ...Args>
std::function<R(Args...)> cached(C & cache, std::function<R(Args...)> f)
{
    return [f,&cache](Args... args){
        auto key = make_tuple(args...);
        if (cache.count(key) > 0)
            return cache[key];
        else
        {
            R result = f(std::forward<Args>(args)...);
            cache.insert(std::pair<decltype(key), R>(key, result));
            return result;
        }
    };
}
```

jcoatelen-ledger (changes requested):

core/src/utils/Cached.h
jcoatelen-ledger: generic implementation with variadic template can be used in the first option as well

jcoatelen-ledger (approved):
live reviewed, LGTM !!

PR #915, opened 2022-10-19
Take dust into account when computing the balance

BertrandD:


jcoatelen-ledger (approved):
approved, as having unspendable funds shown in balance seems ok for clients :)

PR #916, opened 2022-10-25
VG-10629 Filter worthless UTXOs when crafting & listing

BertrandD:
resolve VG-10629

# Changes

Adds notion of **worthless utxos**:  at a certain time, utxo that adds more fees than the utxo value

- When crafting:
   - Filter out worthless utxos
- When listing utxo:
   - Filter out worthless utxos
-  When getting balance:
   - Use *all* utxos

Balance is not impacted by this fix

Cons:
	- List of UTXO will update day-to-day because of network fees
	-  **Locked amount on balance** increasing & fluctuating day-to-day because of network fees

# Test plan

## Automatic

**Proposal**: Add a new e2e test with praline
- Send dust, worthless, and big utxos to account
- Get balance should return total balance
- Craft should ignore dust & worthless
- Listing of utxos should ignore dust & worthless

## Manual

No manual testing needed


jcoatelen-ledger (approved):
live reviewed, great work !

PR #917, opened 2022-11-08
Improve get balance time to compute

BertrandD:
Use postgresql sum instead of useless fetching-all + ordering before summing in memory


twilgenbus-ledger (approved):
LGTM!!

jcoatelen-ledger (approved):
Nice !

PR #920, opened 2022-11-10
Filter wallets with unuspported currencies

BertrandD:


cdiaz-ledger (approved):
LGTM

twilgenbus-ledger (approved):
LGTM!

core/src/wallet/pool/database/PoolDatabaseHelper.cpp
```diff
@@ -54,12 +54,18 @@ namespace ledger {
...
+                   "'bitcoin_gold', 'zcash', 'zencash', 'litecoin', 'peercoin', 'digibyte', 'hcash', 'qtum', 'stealthcoin', 'vertcoin', 'viacoin', 'dash', 'dogecoin', "
+                   "'stratis','komodo', 'poswallet', 'pivx', 'clubcoin', 'decred', 'stakenet', 'cosmos', 'ethereum', 'ethereum_ropsten','ethereum_classic', 'ripple', "
+                   "'tezos', 'stellar', 'algorand')", use(pool.getName()), into(count);
             return count;
         }
 
         int64_t PoolDatabaseHelper::getWallets(soci::session &sql, const WalletPool &pool, int64_t offset, std::vector<WalletDatabaseEntry> &wallets) {
-            rowset<row> rows = (sql.prepare << "SELECT uid, name, currency_name, configuration FROM wallets WHERE pool_name = :pool "
+            rowset<row> rows = (sql.prepare << "SELECT uid, name, currency_name, configuration FROM wallets WHERE pool_name = :pool AND currency_name in ('bitcoin', "
+                                               "'bitcoin_testnet', 'bitcoin_regtest', 'bitcoin_cash', 'bitcoin_gold', 'zcash', 'zencash', 'litecoin', 'peercoin', 'digibyte', "
```
jcoatelen-ledger, line 66: I would have 2 concerns: 
- code duplication (must have): I would propose to compute the `supportedCurrencyList` in a central function
- hardcoded values (nice to have): no way to fetch the supported currency list from the db? For example, we have the [currencies table](https://github.com/LedgerHQ/lib-ledger-core/blob/77c853f74079e1f53cecfe25094e52bf97699a6c/core/src/database/migrations.cpp#L87), no way to use that instead?

jcoatelen-ledger (changes requested):

core/src/wallet/pool/database/PoolDatabaseHelper.cpp
BertrandD: * I first used a `constexpr std::array<std::string, 32>` for listing the valid currencies, but soci do not support std::array in the use() function. Thus I changed to `std::vector`, but, when you use an sql prepare statement, if ise `in (:currencies)`, the binder waits a single variable, but when giving a vector, you actualy give 32 ones... then the prepare statement complains. That's why I hard-coded it in the query. But I still can concatenate 2 strings instead of duplicating it, but it looked to me like an anti-pattern when you are actually trying to do a prepare statement instead of concat strings to build a sql query...

* The whole issue is that we cannot trust the database because other components are writing in it ! So If I list the currencies available in tbe DB, I will get polygon for example, which is exactly what I don't want !

jcoatelen-ledger: Ok for the second point, even if I remain worried about the maintenance of this in the future :) It's okay if we don"t add more coins of course hehe but for a "normal" code base I would not recommend this to have a list to update in SQL definition every time we add a coin. 

Concerning the main point, I would consider code duplication (and worst hardcoded code duplication) as a worst approach. It would possibly lead to possible inconsistency between getWallets and getWalletCount in time. Let's please do some old school string concatenation and have a single list to maintain :)

BertrandD: I totally agree that the design of this solution is :hankey: but we discussed about it and other ways of doing it, but we couldn't easily do something smart... Like often in WD/libcore :/

I did an update using fmt, should be better like that

jcoatelen-ledger: thanks !

PR #922, opened 2022-11-28
VG-11533 Remove keychain filtering when fetching operation

BertrandD:
VG-11533


core/src/wallet/tezos/TezosLikeAccount.cpp
```diff
@@ -285,13 +284,8 @@ namespace ledger {
                 std::vector<Operation> operations;
 
                 auto keychain = self->getKeychain();
```
jcoatelen-ledger, line 286: `keychain` not used

jcoatelen-ledger (changes requested):
minor changes only, well done!

jcoatelen-ledger (approved):

PR #925, opened 2022-12-01
Fix macos build

BertrandD:


ngattolin-ledger (approved):
LGTM

.github/workflows/tests.yml
```diff
@@ -101,6 +103,7 @@ jobs:
         run: .github/scripts/build.sh
         env:
           OPENSSL_ROOT_DIR: /usr/local/opt/openssl/
+          ARGS:
```
twilgenbus-ledger, line 106: Y no ninja in macos? 😭

BertrandD: it's failing with ninja on macos, we did not understand why

PR #928, opened 2022-12-07
Test that verifies that we don't create transactions with unsupported witness version

viktorb-ledger:
An additional test that verifies that we don't craft transaction with unsupported witness version.


BertrandD (approved):
LGTM !

PR #936, opened 2023-05-30
VG-750 fix(explorer): Return explorer error in broadcast error message

BertrandD:


jcoatelen-ledger (approved):
very cool, nice to see it coming !
we can add it for other coins later though 
LGTM

PR #938, opened 2023-06-27
Fix ripple explorer (de)serialization issues

jcoatelen-ledger:
Fixes [VSD-2389](https://ledgerhq.atlassian.net/browse/VSD-2389) and [VSD-2405](https://ledgerhq.atlassian.net/browse/VSD-2405)

Changes: 
- `ledger_index_min` is now returned as a integer object and not a string-integer ; the synchronization fails at fetching the account transactions. Fix: change `ledger_index_min` type from string to unsigned int64 
- `Amount` attribute is missing for `AccountDelete` tx type. Fix: parse `DeliveredAmount`, which is there for this tx type instead. 

Note: I would not suggest to use `delivered_amount` which seems always returned because I don't know how the parsing would go for other tx types, e.g. if `delivered_amount` is always matching `Amount` (even if I suppose they will be always equal). 

Results:
sync successful, no error at sync, correct amount saved in db
image

[VSD-2389]: https://ledgerhq.atlassian.net/browse/VSD-2389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VSD-2405]: https://ledgerhq.atlassian.net/browse/VSD-2405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
+++ b/core/src/wallet/ripple/explorers/NodeRippleLikeBlockchainExplorer.cpp
@@ -215,7 +215,8 @@ namespace ledger {
             bodyRequest.pushParameter("account", addresses[0]);
             bodyRequest.pushParameterBool("forward", true);
             if (fromBlockHash.hasValue() && _paginationMarker.empty()) {
-                bodyRequest.pushParameter("ledger_index_min", fromBlockHash.getValue());
+                BigInt blockHash{fromBlockHash.getValue()};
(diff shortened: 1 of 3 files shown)


BertrandD (approved):
LGTM, assuming it has been tested manually

